PR #30, opened 2017-05-03
Demo views

danrabbit:
This moves all of the views in the demo to their own files. It makes it much easier to find the "code samples"

Each view should be its own commit (with the exception of toasts which snuck in with modebutton, sorry) for easier reviewing


danrabbit:
Yeah I agree it's awkward to look at, but it follows the syntax of WidgetNameView

PR #40, opened 2017-05-06
Make ModeButton use toggled instead of button_press_event

### LINKED ISSUE #39: ModeButton activates on press instead of press + release
Buttons in GTK typically respond on "clicked" which is a press + release; for example, this lets you drag a window from a button without triggering the button. Granite ModeButtons, however, appear to activate on press only, meaning if you drag a window from a ModeButton, you also press the button.

This affects at least Music and Files with regards to dragging the window, but it's reproducible in Granite Demo's ModeButtons; compare the behavior of the "Clear Selected" button (pressing, then dragging off the button does not activate it) to the ModeButtons (pressing, then dragging off activates anyway).

cassidyjames:
Fixes #39

### DIFF
--- a/lib/Widgets/ModeButton.vala
+++ b/lib/Widgets/ModeButton.vala
@@ -120,9 +120,10 @@ namespace Granite.Widgets {
             item.scroll_event.connect (on_scroll_event);
             item.add (w);
 
-            item.button_press_event.connect (() => {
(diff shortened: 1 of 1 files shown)


danrabbit:
I can confirm that this solves the issue and it doesn't touch API so I think it's :ok_hand:

PR #41, opened 2017-05-06
Switch WelcomeButton from button_release_event to clicked

cassidyjames:
This makes WelcomeButton behave more similarly to normal buttons; instead of looking for the release event, we look for the click event. 

Gtk handles this slightly differently by requiring the press and release to be performed inside of the widget, which has a side effect of allowing a sort of cancel when you press, then drag outside the widget, then release.


danrabbit:
Super simple fix with no API changes so I'm happy :)

PR #42, opened 2017-05-07
Demo: update URLs

codygarver:


danrabbit:
I can verify all of these URLs are valid and correct

PR #44, opened 2017-05-08
Demo: Replace Welcome with a StackSidebar

danrabbit:
The number of items has really outgrown the welcome widget. This replaces using the welcome for navigation with a Gtk.StackSidebar and creates a new Welcome view with only two items.

I also added a frame and some margin to SourceList because it looked freakin weird
screenshot from 2017-05-07 17 45 46

### DIFF
--- a/demo/Views/WelcomeView.vala
+++ b/demo/Views/WelcomeView.vala
@@ -30,7 +30,7 @@ public class WelcomeView : Gtk.Grid {
             switch (index) {
                 case 0:
                     try {
-                        AppInfo.launch_default_for_uri ("https://valadoc.org", null);
+                        AppInfo.launch_default_for_uri ("https://valadoc.org/granite/Granite.html", null);
                     } catch (Error e) {
                         warning ("%s\n", e.message);
                     }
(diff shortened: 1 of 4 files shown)


cassidyjames:
UX looks good here, but need someone else to poke at the code.

demo/Views/WelcomeView.vala
```diff
@@ -0,0 +1,48 @@
...
+
+        add (welcome);
+
+        welcome.activated.connect ((index) => {
+            switch (index) {
+                case 0:
+                    try {
+                        AppInfo.launch_default_for_uri ("https://valadoc.org/granite/Granite.html", null);
+                    } catch (Error e) {
+                        warning ("%s\n", e.message);
```
donadigo: You don't need this "%s\n" string formatting here, `warning ()` takes care of that, just do:
`warning (e.message)`.

demo/Views/WelcomeView.vala
```diff
@@ -0,0 +1,48 @@
...
+                    }
+                    break;
+                case 1:
+                    try {
+                        AppInfo.launch_default_for_uri ("https://github.com/elementary/granite", null);
+                    } catch (Error e) {
+                        warning ("%s\n", e.message);
```
donadigo: Same here.

demo/Views/WelcomeView.vala
```diff
@@ -0,0 +1,48 @@
...
+                    }
```
donadigo: Newline?

PR #45, opened 2017-05-08
Demo: Add CSSView

danrabbit:
Adds a view showing off our rad CSS classes

screenshot from 2017-05-08 10 53 55

### DIFF
--- a/demo/Views/CSSView.vala
+++ b/demo/Views/CSSView.vala
@@ -1,0 +1,83 @@
+/*-
+ * Copyright (c) 2017 elementary LLC. (https://elementary.io)
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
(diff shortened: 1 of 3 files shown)


demo/Views/CSSView.vala
```diff
@@ -0,0 +1,83 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
+public class CSSView : Gtk.Grid {
+    public Gtk.Window window { get; set; }
```
donadigo: Since you only set window at construction, you could use `{ get; construct; }` here. If you want to avoid passing `Gtk.Window` to this class remember that you can have signals like `signal void color_changed (Gdk.RGBA color)`, and have the root set the color for the window. That's only just a tip though.

danrabbit:
You're right, changed to construct :+1: I think I'm okay with passing the window in here to keep things contained. One of my goals is to make sure it's clear for people who want to come check out the source code to see how something is done.

PR #54, opened 2017-05-11
Application.vala: make public properties get; set;

danrabbit:
Allow setting Granite.Application properties in the Object


tintou:
That's an API break

purejava:
@tintou could you please clarify what you mean?

jeremypw:
The C code required to access public variables (which are public members of the Application struct in the resulting C code, which can be accessed directly) is incompatible with that required access public properties (which are private members of ApplicationPrivate struct accessed via get and set functions).

Code accessing this library would therefore need to be recompiled or rewritten.

PR #56, opened 2017-05-11
Application.vala: deprecate create_appmenu

danrabbit:
Since AppMenu itself is now also deprecated


donadigo:
Conflicts with master.

PR #60, opened 2017-05-23
deprecate text style methods which are one-liners in gtk already

danrabbit:
This may have made sense at the time, but setting a style class is a one-liner in Gtk already and these text style classes have constants defined elsewhere in Granite


danrabbit:
@purejava I agree, but I think that's outside the scope of this branch

Issue #20, opened 2017-04-21
Seekbar widget [$15]


elementaryBot:
There might be a better word for this, but I can't think of it.

We have a little playback statusbar in Audience and in Noise that includes a gtk.scale and two labels showing current time and time remaining. In the case of Audience it also includes a preview popover. Now there's also a third party app, Vocal, using an imitation of this.

I think we should consider making this a widget.

At the very least, we should adopt a class name for this widget because I'd like the gtk.scale in this widget not to have a visible handle.

Launchpad Details: [#LP1317211](https://bugs.launchpad.net/bugs/1317211) Daniel Fore - 2014-05-07 17:54:00 +0000

elementaryBot:
+1. It should probably be like an hbox with the requisite elements inside it.

Launchpad Details: [#LPC](https://bugs.launchpad.net/granite/+bug/1317211/comments/1) Avi Romanoff - 2014-05-07 18:47:17 +0000

elementaryBot:
and of course helper methods/signals.

Launchpad Details: [#LPC](https://bugs.launchpad.net/granite/+bug/1317211/comments/2) Avi Romanoff - 2014-05-07 18:47:34 +0000

elementaryBot:
https://www.bountysource.com/issues/2115104-seekbar-widget

Launchpad Details: [#LPC](https://bugs.launchpad.net/granite/+bug/1317211/comments/3) Daniel Fore - 2016-04-21 19:39:24 +0000

danrabbit:
For reference: https://github.com/elementary/videos/blob/master/src/Widgets/Player/TimeWidget.vala

rickybas:
Can this be closed now?

artemanufrij:
@danrabbit @codygarver can we shipped it asap? I moved to Granite.Seekbar in my music player... Otherwise I have to move back. :confused:

codygarver:
@artemanufrij the new granite release will happen when elementary/granite#106 gets merged.

PR #62, opened 2017-05-31
Add Seek Bar

rickybas:
In reference to issue #20 

Ability to set duration and progress of the SeekBar, included methods/signals needed for Audience and Noise

### Changes 
* playback_duration: a public set and get property
* playback_progress: a public set and get property that takes a double from 0.0 -> 1.0
* Debug messages added
* Seek Bar sliding demo
* Seek Bar docs image


rickybas:
Hi @danrabbit , thank you for your feedback, I have tried to implement all your suggestions. 

Also as I was building the new demo I was wondering if you would like the preview popover to be part of the Seek Bar since I have to expose the scale widget as a public property:
line 46 demo/Views/SeekBarView.vala

danrabbit:
I'm going to assign @nathandyer as a reviewer here because I'd like to get a review of the API from someone who would be using it in their app

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,232 @@
...
+        if (width > 0 && width >= minimum_width) {
+            natural_width = width;
+        }
+    }
+
+
+    /*
+     * Converts seconds into the ISO 8601 standard date format for minutes (e.g. 100s to 01:40)
+     */
+    public string seconds_to_time (int seconds) {
```
donadigo: This function doesn't use anything from the instance, hence should be `static`. I'm also not sure about the place that it's in. Perhaps that should be moved into `Granite.DateTime`?

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,232 @@
...
+        string sec = normalize_time (seconds % 60);
+
+        if (hours > 0) {
+            return ("%d:%s:%s".printf (hours, min, sec));
+        } else {
+            return ("%s:%s".printf (min, sec));
+        }
+    }
+
+    private string normalize_time (int time) {
```
donadigo: The same goes for this function.

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,232 @@
...
+        }
+
+        this.playback_duration = duration;
+        time_label.label = seconds_to_time ((int) duration);
+    }
+
+    public override void get_preferred_width (out int minimum_width, out int natural_width) {
+        base.get_preferred_width (out minimum_width, out natural_width);
+
+        if (parent.get_window () == null)
```
donadigo: Could you get the window once? Also, there is no guarantee that `parent` will be non-null. You should check for that also, like that:
```
if (parent == null) {
    return;
}

var window = parent.get_window ();
if (window == null) {
    return;
}

var width = window.get_width ();
...
```

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,232 @@
...
 
+    /*
+     * Sets the progress of the SeekBar, with a value between 0.0 and 1.0.
+     * */
+    public void set_progress (double progress) {
+        if (progress < 0.0) {
+            warning ("Progress value less than zero, progress set to 0.0");
+            progress = 0.0;
+        }
+        else if (progress > 1.0) {
```
donadigo: Move that `else if` up one line:
```
if (progress < 0.0) {

} else if (progress > 1.0) {

}
```

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,232 @@
...
+            scale_motion (event);
+            return false;
+        });
+
+        scale.change_value.connect ((scroll, new_value) => {
+            scroll_action (scroll, new_value);
+            return false;
+        });
+
+        scale.size_allocate.connect ((alloc_rect) => {
```
donadigo: This connection seems to do nothing in it's lambda. You could just remove it altogheter.

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,232 @@
...
+     */
+    public Gtk.Scale scale {  get; private set;  }
 
-    /**
-     * Creates a new SeekBar
+
+    /*
+     * Emitted when the scale's window is hovered over.
+     */
+    public signal void scale_hover (Gdk.EventCrossing event);
```
donadigo: I don't think there is a need for this `scale_` API. The application using SeekBar can just access `scale` widget and operate / connect to it's signals itself without any overhead.

demo/Views/SeekBarView.vala
```diff
@@ -19,11 +19,93 @@
...
+        preview_popover.sensitive = false;
+        preview_popover.modal = false;
+        preview_popover.valign = Gtk.Align.CENTER;
+
+        preview_label = new Gtk.Label ("");
+        preview_popover.add (preview_label);
+        preview_popover.show_all ();
+        preview_popover.set_visible (false);
+
+        var seek_bar = new Granite.Widgets.SeekBar(100);
```
donadigo: Can we get space between the `SeekBar` and `(100)`?

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,232 @@
...
+     * If the pointer is grabbing the scale button.
+     */
+    public bool is_grabbing { get; private set; }
+
+    /*
+     * If the pointer is hovering over the scale.
+     */
+    public bool is_hovering { get; private set; }
+
+
```
donadigo: Remove this newline.

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,232 @@
...
+    /*
+     * The left label that displays the time progressed.
+     */
+    public Gtk.Label progression_label { get; private set; }
```
donadigo: Can we get `construct` instead of `private set` accessor here?

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,232 @@
...
+
+    /*
+     * The right label that displays the total duration time.
+     */
+    public Gtk.Label time_label {  get; private set;  }
```
donadigo: The same here. Also remove spaces beween the braces and accessors.

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,232 @@
...
+
+    /*
+     * The time of the full duration of the playback.
+     */
+    public Gtk.Scale scale {  get; private set;  }
```
donadigo: The same here.

demo/Views/SeekBarView.vala
```diff
@@ -19,11 +19,93 @@
...
 
 public class SeekBarView : Gtk.Grid {
+
+    private Gtk.Popover preview_popover;
+    private Gtk.Label preview_label;
+
+    public SeekBarView () {
+        Object (halign: Gtk.Align.CENTER,
+                valign: Gtk.Align.CENTER,
+                margin: 24);
```
donadigo: I don't think setting these properties are needed here, since the application will probably set those manually. In particular the `margin` property.

demo/Views/SeekBarView.vala
```diff
@@ -19,11 +19,93 @@
...
+        });
+
+        seek_bar.scale_button_release.connect (() => {
+            preview_label.height_request = 0;
+        });
+
+        seek_bar.scroll_action.connect ((scroll, new_value) => {
+            if (new_value >= 0.0 && new_value <=1.0) {
+                var duration_mins = seek_bar.seconds_to_time ((int) (new_value*seek_bar.playback_duration));
+                preview_label.label = duration_mins.to_string();
```
donadigo: Space between the parenthesis and `to_string`.

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,232 @@
...
     construct {
-        var progression_label = new Gtk.Label ("0:00");
+        get_style_context ().add_class ("seek-bar");
 
-        var scale = new Gtk.Scale.with_range (Gtk.Orientation.HORIZONTAL, 0, 1, 0.1);
-        scale.draw_value = false;
+        /* GUI */
+        orientation = Gtk.Orientation.HORIZONTAL;
+        column_spacing = 6;
+        halign = Gtk.Align.CENTER;
```
donadigo: Again I don't recommend setting `halign` here.

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,232 @@
...
+        progression_label = new Gtk.Label ("");
+        progression_label.get_style_context ().add_class ("progression-label");
+        time_label = new Gtk.Label ("");
```
donadigo: Can this `time_label` go underneath the `progression_label` label initialization? It would make things clearer to see.

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,232 @@
...
+        scale.get_style_context ().add_class ("scale");
         scale.hexpand = true;
+        scale.draw_value = false;
+        scale.can_focus = false;
+        scale.events |= Gdk.EventMask.POINTER_MOTION_MASK;
+        scale.events |= Gdk.EventMask.LEAVE_NOTIFY_MASK;
+        scale.events |= Gdk.EventMask.ENTER_NOTIFY_MASK;
+
+        is_grabbing = false;
+        is_hovering = false;
```
donadigo: Instead of setting these to `false` you could use the `default` in properties like this:
`public bool is_grabbing { get; private set; default = false; }`

lib/Widgets/SeekBar.vala
rickybas: Agreed

rickybas:
The requested changes have been implemented

rickybas:
Just wondering if you would like me to do anything else for this to be merged? @danrabbit

demo/Views/SeekBarView.vala
```diff
@@ -19,11 +19,96 @@
...
+                preview_label.label = duration_mins.to_string ();
+            }
+            return false;
+        });
+
+        add (seek_bar);
+
+        int progress = 0;
+        Timeout.add (500, () => {
+            if (seek_bar.is_grabbing) return true;
```
matzipan: Should this be managed here or in `set_progress` in the SeekBar?

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,159 @@
...
-        column_spacing = 12;
         add (progression_label);
         add (scale);
         add (time_label);
     }
 
+    /*
+     * Sets the progress of the SeekBar with a value between 0.0 and 1.0.
+     * */
+    public void set_progress (double progress) {
```
matzipan: Is there a reason why this is not an attribute and this is a setter on it?

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,159 @@
...
+        if (progress < 0.0) {
+            warning ("Progress value less than 0.0, progress set to 0.0");
+            progress = 0.0;
+        } else if (progress > 1.0) {
+            warning ("Progress value greater than 1.0, progress set to 1.0");
+            progress = 1.0;
+        }
+
+        this.playback_progress = progress;
```
matzipan: No need for `this` here.

lib/Widgets/SeekBar.vala
```diff
@@ -16,34 +16,159 @@
...
+        scale.set_value (playback_progress);
+        progression_label.label = DateTime.seconds_to_time ((int) (playback_progress * playback_duration));
+    }
+
+    /*
+     * Sets the duration of the SeekBar with a value greater than 0.0.
+     * */
+    private void set_duration (double duration) {
```
matzipan: Again, this should be a setter on an attribute

demo/Views/SeekBarView.vala
```diff
@@ -19,11 +19,96 @@
  */
 
 public class SeekBarView : Gtk.Grid {
+    private Gtk.Popover preview_popover;
+    private Gtk.Label preview_label;
+
+    public SeekBarView () {
+        Object (valign: Gtk.Align.CENTER,
+                margin: 24);
```
matzipan: Is there a reason why these are not set in the `construct` method?

matzipan:
Functional testing: touch screen handled fine, I think I have managed to stop the progressbar from working after I messed around with it with my mouse a few times - not sure if bug since I can't reproduce. 

Furthermore, I think it would be nice to have a bigger cursor (or at least the option for one), such as the one present in the Videos app.

demo/Views/SeekBarView.vala
danrabbit: `Object ()` is here to set Object properties ;)

demo/Views/SeekBarView.vala
rickybas: Not sure what you mean be this. This bit of code is used to simulate behaviour of a progressing media player.

matzipan: Of course. I'm referring strictly to the `if (seek_bar.is_grabbing) return true;` part. Also please add braces `{}` around `return true` as per code style.

demo/Views/SeekBarView.vala
matzipan: Sure, but it's hardcoding those values. I thought we're keeping all UI related code in the `construct`?

rickybas:
I think the slider moves all the way to the ends, its just the popover that moves across the slider's width with the mouse. I can't find any option to change the slider size and I haven't been able to reproduce the mouse bug.

demo/Views/SeekBarView.vala
rickybas: I think it should be managed there because the SeekBar, internally, doesn't decide how it should progress (the progression is decided externally) therefore, what happens when the Scale is grabbed, is decided by the user of the SeekBar.

matzipan:
Well the slider size has been fixed in Audience (the videos app), so have a look at the source. 

Video shows how the slider is not reaching the end of the seekbar https://photos.app.goo.gl/yoNVGKuUpd7XvuxU2

matzipan:
Hey @rickybas, any progress on this?

rickybas:
I've changed to attribute setters and have fixed the scale size problem. See commits above.

rickybas:
Hi @matzipan , I've changed to attribute setters and have fixed the scale size problem. See commits above.

lib/Widgets/SeekBar.vala
```diff
@@ -25,15 +25,51 @@
...
      */
-    public double playback_duration { get; construct set; }
+    public double playback_duration {
+        get {
+            return _playback_duration;
+        }
+        construct set {
+            double duration = value;
+            if (duration < 0.0) {
+                warning ("Duration value less than zero, duration set to 0.0");
```
matzipan: I don't think this should be a warning, but debug information.

lib/Widgets/SeekBar.vala
```diff
@@ -25,15 +25,51 @@
...
      */
-    public double playback_progress { get; private set; }
+    public double playback_progress {
+        get {
+            return _playback_progress;
+        }
+        set {
+            double progress = value;
+            if (progress < 0.0) {
+                warning ("Progress value less than 0.0, progress set to 0.0");
```
matzipan: Same here.

lib/Widgets/SeekBar.vala
```diff
@@ -25,15 +25,51 @@
...
+                progress = 0.0;
+            } else if (progress > 1.0) {
+                warning ("Progress value greater than 1.0, progress set to 1.0");
```
matzipan: Same here.

rickybas:
@danrabbit @nathandyer @donadigo How is the code now?

danrabbit:
I think the only thing missing to merge this is to prepare a PR or two against relevant apps to use this widget. Then we can know for sure that the API is complete and immediately useful.

Marking this as "Blocked" until we have a PR against Music, Videos, Vocal, or another app that would use this widget

artemanufrij:
I tested current PR with PlayMyMusic (https://github.com/artemanufrij/playmymusic).

All functions work as expected. Current master branch is based on this seekbar. 

bildschirmfoto von 2017-10-17 20 07 14

It would be useful to ship it next time :smile:. otherwise I can't release new versions :smile:

rickybas:
Thanks!

PR #63, opened 2017-06-05
Avatar: Make it possible to update the file

danrabbit:
Adds a `file` property to easily set a new file for the avatar so apps don't have to muck about with pixbufs.

Rename `filepath` to `filename` to be consistent with Gtk


lib/Widgets/Avatar.vala
```diff
@@ -28,10 +28,34 @@ public class Granite.Widgets.Avatar : Gtk.EventBox {
...
 
     /**
+     * The "pixel-size" property can be used to specify a fixed size.
+     */
+    public int pixel_size { get; construct; }
+
+    /**
+     * The "file" property can be used to set a new avatar image from a file path.
+     */
+    public string? file {
```
donadigo: 1. In the context of GLib-based code for me `file` would mean that I need to pass a `GLib.File` object to this property, not a string. I think this should be quite the opposite. The `filename` should be the name of the property as it points that we require a path so a `string`.

2. The type of `file` is implying that the property is nullable (which is correct), that means that that the API user can pass `null` to `file` and expect it to work correctly. This is not the case here as you don't handle `null` in anyway. This will result in errors in Granite. Yes, you're catching the error from creation of the pixbuf but before that you should actually check if the value is `null`. If yes, then show the default avatar. If no, continue loading the Pixbuf. If creation of the Pixbuf fails, throw a warning and show the default avatar

danrabbit: I used `file` because that's how `Gtk.Image` does it: https://valadoc.org/gtk+-3.0/Gtk.Image.file.html . Trying to emulate Gtk+ API here as much as possible. It also matches the name of the creation method since it is `from_file` and not `from_filename`. Same as `Gtk.Image.from_file` which also takes a file: https://valadoc.org/gtk+-3.0/Gtk.Image.Image.from_file.html

Yeah you're right, I should check for null and not rely on the catch for that. will fix!

PR #70, opened 2017-06-23
Add SettingsSidebar

danrabbit:
This is a replacement for the various settings sidebar implementations in Switchboard plugs:

screenshot from 2017-07-12 19 41 42

Plugs with ports already:
- [A11y](https://github.com/elementary/switchboard-plug-a11y/pull/23)
- [Security & Privacy](https://github.com/elementary/switchboard-plug-security-privacy/pull/26)
- [Sharing](https://github.com/elementary/switchboard-plug-sharing/pull/18)


lib/Widgets/SettingsSidebar/SettingsSidebar.vala
```diff
@@ -0,0 +1,97 @@
...
+        listbox.selection_mode = Gtk.SelectionMode.SINGLE;
+
+        add (listbox);
+
+        foreach (unowned Gtk.Widget child in stack.get_children ()) {
+            string name;
+
+            stack.child_get (child, "name", out name, null);
+
+            var page = (SettingsPage) child;
```
donadigo: The `child` is not guaranteed to be `SettingsPage` hence this can throw errors or even crash. You should check if the `child` is actually a type of `SettingsPage`.

lib/Widgets/SettingsPage/AbstractSettingsPage.vala
```diff
@@ -21,15 +21,59 @@
  * AbstractSettingsPage is a #Gtk.ScrolledWindow subclass with properties used
  * by other Granite settings widgets.
  */
-
 public abstract class Granite.SettingsPage : Gtk.ScrolledWindow {
+    protected string _icon_name;
+    protected string _title;
+
+    public enum StatusType {
```
donadigo: I feel like this enum should have some kind of `to_icon_name` method so that it can be easily converted into a string that can be used in Gtk widgets.

lib/Widgets/SettingsPage/AbstractSettingsPage.vala
```diff
@@ -21,15 +21,59 @@
...
+
+    /**
+     * A header to be sorted under in a Granite.SettingsSidebar
+     */
+    public string? header { get; construct; }
+
+    /**
+     * An status string associated with #this
+     */
+    public string status { get; set; }
```
donadigo: This `status` property isn't exactly clear on what it does. Perhaps say it is the description of the status like "enabled" and "disabled"?

lib/Widgets/SettingsPage/AbstractSettingsPage.vala
```diff
@@ -21,15 +21,59 @@
...
+
     /**
      * An icon name associated with #this
      */
-    public string? icon_name { get; construct set; }
+    public string? icon_name {
```
donadigo: Hm, can you explain why this has changed to a structure like this? This is exactly the same as what the previous code for this property was.

lib/Widgets/SettingsPage/AbstractSettingsPage.vala
```diff
@@ -21,15 +21,59 @@
...
+        construct set {
+            _icon_name = value;
+        }
+    }
 
     /**
      * A title associated with #this
      */
-    public string title { get; construct set; }
+    public string title {
```
donadigo: The same here.

lib/Widgets/SettingsSidebar/SettingsSidebarRow.vala
```diff
@@ -0,0 +1,124 @@
...
+
+    public string? header { get; set; }
+
+    public string icon_name {
+        get {
+            return _icon_name;
+        }
+        set {
+            _icon_name = value;
+            ((Gtk.Image) display_widget).icon_name = value;
```
donadigo: You make the `display_widget` be an arbitrary type of `Gtk.Widget`, and here you cast it to `Gtk.Image` which cannot be done like that. You have to check if the widget is indeed a `Gtk.Image` here.

lib/Widgets/HeaderLabel.vala
```diff
@@ -0,0 +1,40 @@
...
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*/
+
+
+/**
+ * HeaderLabel is a start-aligned Gtk.Label with the Granite H4 style class
+ */
+public class Granite.HeaderLabel : Gtk.Label {
```
donadigo: This class should be in `Widgets` namespace.

lib/Widgets/SettingsPage/AbstractSettingsPage.vala
```diff
@@ -21,15 +21,59 @@
  * AbstractSettingsPage is a #Gtk.ScrolledWindow subclass with properties used
  * by other Granite settings widgets.
  */
-
 public abstract class Granite.SettingsPage : Gtk.ScrolledWindow {
```
donadigo: I know this branch didn't change it, but this class should be within `Widgets` namespace.

lib/Widgets/SettingsSidebar/SettingsSidebarRow.vala
```diff
@@ -0,0 +1,124 @@
...
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU Library General Public License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*/
+
+private class Granite.SettingsSidebarRow : Gtk.ListBoxRow {
```
donadigo: This class should be in `Widgets` namespace.

lib/Widgets/SettingsPage/AbstractSettingsPage.vala
danrabbit: Yes, it's because we need the backing variables to be `protected` instead of `private`. This way `SimpleSettingsPage` can access these variables and the cast in `SettingsSidebar` works as expected

demo/Views/SettingsView/SimpleSettingsPage.vala
```diff
@@ -57,6 +60,14 @@ public class SettingsView : Granite.SimpleSettingsPage {
 
         var button = new Gtk.Button.with_label ("Test Button");
 
+        if (status_switch.active) {
```
donadigo: Can we get this block into a separate method like `update_status`?

demo/Views/SettingsView/SimpleSettingsPage.vala
```diff
@@ -65,6 +76,16 @@ public class SettingsView : Granite.SimpleSettingsPage {
             icon_name = icon_entry.text;
         });
 
+        status_switch.notify["active"].connect (() => {
```
donadigo: And then just connecting into `update_status` here?

lib/Widgets/AbstractSettingsPage.vala
```diff
@@ -0,0 +1,82 @@
...
+* GNU Library General Public License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*/
+
+/**
+ * AbstractSettingsPage is a #Gtk.ScrolledWindow subclass with properties used
```
donadigo: Instead of the `#Gtk.ScrolledWindow` syntax which doesn't work on valadoc you should use `{@link Gtk.ScrolledWindow}`.

lib/Widgets/AbstractSimpleSettingsPage.vala
```diff
@@ -28,8 +28,6 @@ public abstract class Granite.SimpleSettingsPage : Granite.SettingsPage {
     private Gtk.Label description_label;
     private Gtk.Label title_label;
     private string _description;
-    private string _icon_name;
-    private string _title;
 
     /**
      * A #Gtk.ButtonBox used as the action area for #this
```
donadigo: The same `#Gtk.` syntax goes here.

lib/Widgets/AbstractSettingsPage.vala
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * A status string to be displayed underneath the title in a Granite.SettingsSidebar
+     */
+    public string status { get; set; }
+
+    /**
+     * An icon name to be displayed in a Granite.SettingsSidebar
+     */
+    public string? icon_name {
```
donadigo: I still cannot understand why there is `_icon_name` field. This `icon_name` property should be `virtual` and classes like `SimpleSettingsPage` should override it. Currently this code throws warnings like:
```
AbstractSimpleSettingsPage.vala:70.5-70.27: warning: Granite.SimpleSettingsPage.icon_name hides inherited property `Granite.SettingsPage.icon_name'. Use the `new' keyword if hiding was intentional
AbstractSimpleSettingsPage.vala:85.5-85.23: warning: Granite.SimpleSettingsPage.title hides inherited property `Granite.SettingsPage.title'. Use the `new' keyword if hiding was intentional
```

lib/Widgets/AbstractSettingsPage.vala
```diff
@@ -0,0 +1,82 @@
...
+        } 
+        construct set {
+            _icon_name = value;
+        }
+    }
+
+    /**
+     * A title to be displayed in a Granite.SettingsSidebar
+     */
+    public string title {
```
donadigo: The same hiding property is also here.

lib/Widgets/AbstractSettingsPage.vala
danrabbit: Remember we tried it as virtual and it breaks the sidebar. We need these backing variables to be protected instead of private because they need to be accessible to the subclass, but they also need to be accessible when we cast as the superclass.

lib/Widgets/SettingsSidebar.vala
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    construct {
+        var listbox = new Gtk.ListBox ();
+        listbox.activate_on_single_click = true;
+        listbox.selection_mode = Gtk.SelectionMode.SINGLE;
+
+        add (listbox);
+
+        foreach (unowned Gtk.Widget child in stack.get_children ()) {
```
Philip-Scott: You should connect to the Stack's .add event, and do everything here when items are added to the stack. It not, adding a new page would require a new sidebar @

lib/Widgets/AbstractSettingsPage.vala
matzipan: Then perkaps add `new` in the overriding declaration?

danrabbit: Again, it cannot be an override. overriding breaks the cast

matzipan: Fair enough - should this then be escalated upstream to Vala?

PR #72, opened 2017-06-24
Toast.vala: Add timeouts

### LINKED ISSUE #71: Add a duration property for Toasts
It looks like [the Android implementation](https://developer.android.com/reference/android/widget/Toast.html#setDuration(int)) offers a property `duration`.

Gtk.Stack and Gtk.Revealer both offer a property `uint transition_duration`, so I think `duration` is a sane name.

The Android implementation also offers two constants:
```
LENGTH_LONG = 3500; // 3.5 seconds
LENGTH_SHORT = 2000; // 2 seconds
```

I'm not a fan of the names. It seems counter-intuitive to me to have `duration` be the property and then all the sudden the constant is `LENGTH_FOO`. Especially because I think "length" could also refer to physical dimensions so it's kind of an ambiguous name. So I would be in favor of maybe `DURATION_LONG` and `DURATION_SHORT`. In practice this would look like
```
var toast = new Granite.Widgets.Toast (_("This is a notification"));
toast.duration = Granite.Widgets.Toast.DURATION_SHORT;
```

Couple of open questions:
* Should the default duration still be infinity or should non-actionable Toasts disappear by default at some duration?
* Should the `duration` property be a `uint` like `Gtk.Foo.transition_duration` or should it be an `enum` like on Android? (ie no custom durations, only a choice between predefined durations)

danrabbit:
Fixes #71 

Duration is currently private in this branch. I guess if we're deciding that we don't want everlasting toasts we might want to consider if duration should even be public at all

### DIFF
--- a/lib/Widgets/Toast.vala
+++ b/lib/Widgets/Toast.vala
@@ -65,11 +65,6 @@ namespace Granite.Widgets {
         }
 
         /**
-         * The duration until #this is automatically dismissed
-         */
-        public uint duration { get; set; }
-
-        /**
          * Creates a new Toast with #title as its title
          */
         public Toast (string title) {
@@ -134,6 +129,8 @@ namespace Granite.Widgets {
         public void send_notification () {
             if (!child_revealed) {
                 reveal_child = true;
+
+                uint duration;
 
                 if (default_action_button.visible) {
                     duration = DURATION_LONG;
--- a/lib/Widgets/Toast.vala
+++ b/lib/Widgets/Toast.vala
@@ -46,8 +46,6 @@ namespace Granite.Widgets {
         private Gtk.Button default_action_button;
         private string _title;
         private uint timeout_id;
-        private const uint DURATION_LONG = 3500;
-        private const uint DURATION_SHORT = 2000;
 
         /**
          * The notification text label to be displayed inside of #this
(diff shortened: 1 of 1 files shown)


lib/Widgets/Toast.vala
```diff
@@ -79,15 +79,19 @@ namespace Granite.Widgets {
             default_action_button.no_show_all = true;
             default_action_button.clicked.connect (() => {
                 reveal_child = false;
-                Source.remove (timeout_id);
+                if (timeout_id != 0) {
+                    Source.remove (timeout_id);
+                }
```
jeremypw: Insert timeout_id = 0 here.

lib/Widgets/Toast.vala
```diff
@@ -79,15 +79,19 @@ namespace Granite.Widgets {
...
             });
 
             var close_button = new Gtk.Button.from_icon_name ("close-symbolic", Gtk.IconSize.MENU);
             close_button.get_style_context ().add_class ("close-button");
             close_button.clicked.connect (() => {
                 reveal_child = false;
-                Source.remove (timeout_id);
+                if (timeout_id != 0) {
+                    Source.remove (timeout_id);
+                }
```
jeremypw: Insert timeout_id = 0 here

PR #73, opened 2017-06-26
Add SettingsPage and SimpleSettingsPage

danrabbit:
This is a smaller PR as a step towards #70. The SimpleSettingsPage itself is still valuable independent of SettingsSidebar in the Bluetooth plug for example.

SettingsPage is pretty useless in this PR, but I decided to include it to avoid changing API between this merge and a merge including SettingsSidebar

screenshot from 2017-06-26 11 46 41

### DIFF
--- a/demo/Views/SettingsView.vala
+++ b/demo/Views/SettingsView.vala
@@ -27,13 +27,48 @@ public class SettingsView : Granite.SimpleSettingsPage {
             title: "SimpleSettingsPage"
         );
 
-        var label = new Gtk.Label ("This is the content area where settings would go!");
-        label.hexpand = true;
-        label.vexpand = true;
+        var icon_label = new Gtk.Label ("Icon Name:");
+        icon_label.xalign = 1;
 
-        content_area.add (label);
+        var icon_entry = new Gtk.Entry ();
+        icon_entry.hexpand = true;
+        icon_entry.placeholder_text = "This page's icon name";
+        icon_entry.text = "preferences-system";
+
+        var title_label = new Gtk.Label ("Title:");
+        title_label.xalign = 1;
+
+        var title_entry = new Gtk.Entry ();
+        title_entry.hexpand = true;
+        title_entry.placeholder_text = "This page's title";
+
+        var description_label = new Gtk.Label ("Description:");
+        description_label.xalign = 1;
+
+        var description_entry = new Gtk.Entry ();
+        description_entry.hexpand = true;
+        description_entry.placeholder_text = "This page's description";
+
+        content_area.attach (icon_label, 0, 0, 1, 1);
+        content_area.attach (icon_entry, 1, 0, 1, 1);
+        content_area.attach (title_label, 0, 1, 1, 1);
+        content_area.attach (title_entry, 1, 1, 1, 1);
+        content_area.attach (description_label, 0, 2, 1, 1);
+        content_area.attach (description_entry, 1, 2, 1, 1);
 
         var button = new Gtk.Button.with_label ("Test Button");
+
+        description_entry.changed.connect (() => {
+            description = description_entry.text;
+        });
+
+        icon_entry.changed.connect (() => {
+            icon_name = icon_entry.text;
+        });
+
+        title_entry.changed.connect (() => {
+            title = title_entry.text;
+        });
 
         action_area.add (button);
     }
--- a/demo/Views/SettingsView.vala
+++ b/demo/Views/SettingsView.vala
@@ -28,6 +28,8 @@ public class SettingsView : Granite.SimpleSettingsPage {
         );
 
         var label = new Gtk.Label ("This is the content area where settings would go!");
+        label.hexpand = true;
+        label.vexpand = true;
 
         content_area.add (label);
 
--- a/demo/Views/SettingsView.vala
+++ b/demo/Views/SettingsView.vala
@@ -1,0 +1,38 @@
+// -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
+/*-
+ * Copyright (c) 2017 elementary LLC. (https://elementary.io)
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
+public class SettingsView : Granite.SimpleSettingsPage {
+    public SettingsView () {
+        Object (
+            activatable: true,
+            description: "This is a demo of Granite's SimpleSettingsPage",
+            icon_name: "preferences-system",
+            title: "SimpleSettingsPage"
+        );
+
+        var label = new Gtk.Label ("This is the content area where settings would go!");
+
+        content_area.add (label);
+
+        var button = new Gtk.Button.with_label ("Test Button");
+
+        action_area.add (button);
+    }
+}
--- a/demo/Views/SettingsView.vala
+++ b/demo/Views/SettingsView.vala
@@ -1,5 +1,4 @@
-// -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
-/*-
+/*
  * Copyright (c) 2017 elementary LLC. (https://elementary.io)
  *
  * This library is free software; you can redistribute it and/or
--- a/demo/Views/SettingsView.vala
+++ b/demo/Views/SettingsView.vala
@@ -33,7 +33,7 @@ public class SettingsView : Granite.SimpleSettingsPage {
         var icon_entry = new Gtk.Entry ();
         icon_entry.hexpand = true;
         icon_entry.placeholder_text = "This page's icon name";
-        icon_entry.text = "preferences-system";
+        icon_entry.text = icon_name;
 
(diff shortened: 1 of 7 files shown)


lib/Widgets/SettingsPage/AbstractSettingsPage.vala
```diff
@@ -0,0 +1,34 @@
...
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*/
+
+/**
+ * AbstractSettingsPage is a #Gtk.ScrolledWindow subclass with properties used
+ * by other Granite settings widgets.
+ */
+
+public abstract class Granite.SettingsPage : Gtk.ScrolledWindow {
```
donadigo: `icon_name` and `title` should be settable, so that these could be changed later on when something has changed and the app / plug wants to change the page information.

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * SimpleSettingsPage is a widget divided into three sections: a predefined header,
+ * a content area, and an action area.
+ */
+
+public abstract class Granite.SimpleSettingsPage : Granite.SettingsPage {
+    /**
+     * A #Gtk.ButtonBox used as the action area for #this
+     */
+    public Gtk.ButtonBox action_area;
```
donadigo: Can we get this be `{ get; construct; }`?

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * A #Gtk.Grid used as the content area for #this
+     */
+    public Gtk.Grid content_area;
```
donadigo: The same with this.

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * A #Gtk.Switch that appears in the header area when #this.activable is #true
+     */
+    public Gtk.Switch? status_switch;
```
donadigo: I'm not sure about having this switch nullable. I think it would be better to have `activatable` settable and then if `activatable` is `true`, then show the switch, otherwise no.

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * Creates a #Gtk.Switch #status_switch in the header of #this
+     */
+    public bool activatable { get; construct; }
+    public string description { get; construct; }
```
donadigo: The description should be changeable.

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * A #Gtk.Grid used as the content area for #this
+     */
+    public Gtk.Grid content_area { get; construct; }
+
+    /**
+     * A #Gtk.Switch that appears in the header area when #this.activable is #true
+     */
+    public Gtk.Switch status_switch;
```
donadigo: This also should be `{ get; construct; }`

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Creates a #Gtk.Switch #status_switch in the header of #this
+     */
+    public bool activatable { get; construct; }
```
donadigo: I think this `activatable` should be still controlable, so that the switch could be shown / hidden dynamically.

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
```diff
@@ -0,0 +1,159 @@
...
+        if (description != null) {
+            description_label = new Gtk.Label (description);
+            description_label.xalign = 0;
+            description_label.wrap = true;
+
+            header_area.attach (description_label, 1, 1, 1, 1);
+        }
+
+        if (activatable) {
+            status_switch = new Gtk.Switch ();
```
donadigo: I think the switch should be created here always and `activatable` should just control whetever it is visible or not.

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        content_area = new Gtk.Grid ();
+        content_area.column_spacing = 12;
+        content_area.row_spacing = 12;
+        content_area.vexpand = true;
+
+        action_area = new Gtk.ButtonBox (Gtk.Orientation.HORIZONTAL);
+        action_area.set_layout (Gtk.ButtonBoxStyle.END);
+        action_area.set_spacing (6);
```
donadigo: Could we use `=` as we use it everywhere in this class?
`action_area.spacing = 6`

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
```diff
@@ -0,0 +1,159 @@
...
+        title_label.get_style_context ().add_class ("h2");
+
+        var header_area = new Gtk.Grid ();
+        header_area.column_spacing = 12;
+        header_area.row_spacing = 3;
+        header_area.attach (header_icon, 0, 0, 1, 2);
+        header_area.attach (title_label, 1, 0, 1, 1);
+
+        if (description != null) {
+            description_label = new Gtk.Label (description);
```
donadigo: I think the same should be done here. The `description_label` should be created always, even without the description. Just remove the check for `description != null`.

lib/Widgets/SettingsPage/AbstractSettingsPage.vala
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * An icon name associated with #this
+     */
+    public string? icon_name { get; construct set; }
```
donadigo: Now, I'm not sure about having only `icon_name` as a property that you can use to set an image for the page. That means that you cannot use anything like `Gdk.Pixbuf` or `GLib.Icon`. I would say, that there should be another property like a Pixbuf or one that's a `GLib.Icon`. Then you could have helper constructors like `from_icon_name` and `from_file` easily.

donadigo:
Some diff comments again regarding the API.

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
danrabbit: I don't really agree with this. If there's a case where the page is not able to be activated at that time, you should set insensitive, not remove the switch.

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
danrabbit: I don't agree. This seems like something that you either do or your don't do in your page, not something that will be changing.

lib/Widgets/SettingsPage/AbstractSettingsPage.vala
danrabbit: We should never be manually setting a pixbuf, but remember this is SimpleSettingsPage. If you want to do some custom implementation you can do so with SettingsPage

donadigo: Well, I still don't agree. The point of SimpleSettingsPage is to have an already pre-made page that can be used in most of the views. I wouldn't like to implement the entire same page just to display a Pixbuf in my app (like loading an image from a file).

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
donadigo: Okay, I can take that :P

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
donadigo: I can take that too.

lib/Widgets/SettingsPage/AbstractSettingsPage.vala
danrabbit: Do you have a use case for loading a pixbuf?

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
danrabbit: I don't think we should create a widget that will never be shown. Seems like a waste of memory

lib/Widgets/SettingsPage/AbstractSettingsPage.vala
donadigo: I don't. But I suppose if someone really have to use a file for this image, they will either don't use this API at all or copy the entire SimpleSettingsPage.vala into their source and change it just so they can use it. Well, I won't be blocking this PR, because I'm not the one who proposes it. I would like to hear other's opinion about this.

donadigo: Oups, seems like I misundersood the structure of this API. The said class is abstract so others can extend it. I'm fine with that.

lib/Widgets/SettingsPage/AbstractSimpleSettingsPage.vala
donadigo: Okay, then. So you should make `status_switch` nullable, and point out in the documentation that it will be `null` when `activatable` is set to `false`

danrabbit: done :)

demo/Views/SettingsView.vala
```diff
@@ -0,0 +1,75 @@
+// -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
```
Philip-Scott: Drop this man, come on! It's not even a feature in Code!

:hankey:

PR #74, opened 2017-06-28
Deprecate About Dialog

danrabbit:
Since the plan is that going forward all apps should have this data in their appdata.xml (and thus AppCenter) instead


donadigo:
Perhaps it is worth adding a paragraph to the documentation that AboutDialog should not be used and what replaces it?

danrabbit:
Done :+1:

Philip-Scott:
Maybe also remove the About dialog from the demo? :)

PR #75, opened 2017-06-30
Reload and resize Avatar on DPI change

davidmhewitt:
I've been doing some investigation into some of the HiDPI issues with the greeter to try and get a better understand of the code and one of the things that stood out to me was the scale of the Avatar widget. While all the other widgets sort of resize themselves when they detect the scaling of the screen, the Avatar doesn't.

This is reproducible by running the granite-demo and adjusting the display's scale factor with the org.gnome.desktop.interface.scale-factor key. All the GTK widgets resize OK, but the Avatar doesn't. This change now reloads the pixbuf and resizes the widget when the DPI of the monitor changes and hence fixes the issue in the Greeter.

The only downside of this is that it won't have any effect on Avatars that have had their pixbuf directly set by the calling code rather than using the `from_file` or `with_default_icon` helpers.

### DIFF
--- a/lib/Widgets/Avatar.vala
+++ b/lib/Widgets/Avatar.vala
@@ -64,7 +64,7 @@ public class Granite.Widgets.Avatar : Gtk.EventBox {
 
     private void load_image (string filepath, int pixel_size) {
         try {
-            var size = pixel_size * get_style_context ().get_scale ();
+            var size = pixel_size * get_scale_factor ();
             pixbuf = new Gdk.Pixbuf.from_file_at_size (filepath, size, size);
         } catch (Error e) {
             show_default (pixel_size);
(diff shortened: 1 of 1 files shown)


davidmhewitt:
@danrabbit Hm, weird. On master, on my HiDPI display when I scale down from 2x to 1x using the gsettings key, I get this:
screenshot from 2017-07-01 00 37 57

Then after this branch, everything works as expected. I'm not sure I understand the issue you're experiencing. Do you experience the same issue on master, or something different?

danrabbit:
Yeah master is undoubtedly broken, but I can't confirm that this branch fixes the brokenness :)

I installed this branch and then restarted and when I hit the greeter my avatar is half-size. It looks like the greeter is setting avatar from file, so I guess that's not the issue...

davidmhewitt:
Yeah, that would be my fault. Should be resolved now. Good catch.

PR #77, opened 2017-07-01
Utils.vala: replace deprecated list_schemas

danrabbit:


Philip-Scott:
Interestingly enough, we don't use this on any of our apps and a GitHub search couldn't also find any app with this.  Maybe it could be a good candidate for deprecation?

PR #81, opened 2017-07-04
Remove deprecated and buggy app instance

### LINKED ISSUE #80: Granite.Application prevents destructors being called
When subclassing `Granite.Application`, Granite itself keeps a static reference to your app under `Granite.app`. This reference never gets released within Granite which unfortunately leads to destructors in the subclass not being called. The reference is set here:
https://github.com/elementary/granite/blob/master/lib/Application.vala#L158

That property has been deprecated since Granite 0.1 so I'd be interested to see if there was anything still using it.

I have had to work around this in Torrential with the following:
https://github.com/davidmhewitt/torrential/blob/master/src/Application.vala#L118

When you know what the problem is, it's not too bad to solve, but it took a lot of debugging to get to the point of blaming Granite rather than my own app :wink:

donadigo:
Fixes #80.

This branch entirely removes the `Granite.app` static field which prevented destructors from being ran. I would say this change is high priority since it can cause confusion for developers, as seen in #80.

### DIFF
--- a/lib/Application.vala
+++ b/lib/Application.vala
@@ -18,13 +18,6 @@
  */
 
 namespace Granite {
-
-    /**
-     * Global deprecated object..
-     */
-    [Version (deprecated = true, deprecated_since = "0.1", replacement = "")]
-    public static Granite.Application app;
-
     /**
      * This is the base class for all Granite-based apps. It has methods that help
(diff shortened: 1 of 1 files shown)


danrabbit:
Requested review from @davidmhewitt since he discovered this issue

PR #85, opened 2017-07-14
Solve warnings with datepicker

danrabbit:
* No `using`
* change `out` to `ref` to avoid warning about use of possibly unassigned parameter `rect`


lib/Widgets/DatePicker.vala
```diff
@@ -126,16 +123,16 @@ namespace Granite.Widgets {
             Object (format: format);
         }
 
-        private void on_icon_press (EntryIconPosition position) {
+        private void on_icon_press (Gtk.EntryIconPosition position) {
             Gdk.Rectangle rect = Gdk.Rectangle ();
```
donadigo: Do not initialize the Rectangle here, just do `Gdk.Rectangle rect`.

lib/Widgets/DatePicker.vala
```diff
@@ -126,16 +123,16 @@ namespace Granite.Widgets {
...
-            position_dropdown (out rect);
+            position_dropdown (ref rect);
             popover.pointing_to = rect;
             popover.position = Gtk.PositionType.BOTTOM;
             popover.show_all ();
             calendar.grab_focus ();
         }
 
-        protected virtual void position_dropdown (out Gdk.Rectangle rect) {
+        protected virtual void position_dropdown (ref Gdk.Rectangle rect) {
```
donadigo: Leave the `out` parameter here and initialize the Rectangle here: `rect = Gdk.Rectangle ()`

lib/Widgets/DatePicker.vala
```diff
@@ -139,6 +136,7 @@ namespace Granite.Widgets {
             Gtk.Allocation size;
             get_allocation (out size);
 
+            rect = new Gdk.Rectangle ();
```
donadigo: The `Gdk.Rectangle` is a struct, you should create it like that:
`rect = Gdk.Rectangle ();`

PR #89, opened 2017-07-14
Bump so version to 4.2

danrabbit:
Because we broke API


jeremypw:
Should it go from version 3.1 to 4.2 or to 4.0?   i.e. if the major version bumped should the minor version go back to 0?

danrabbit:
Closing as superseded by #117

PR #98, opened 2017-08-23
Add a Granite.MessageDialog

donadigo:
Adds a `Granite.MessageDialog`. The documentation grammar could not perfect, so feel free to point issues out about it or about the API itself.


lib/Widgets/MessageDialog.vala
```diff
@@ -0,0 +1,237 @@
...
+     * add those actions with {@link Gtk.Dialog.add_buttons} or {@link Gtk.Dialog.add_action_widget}.
+     */
+    public Gtk.ButtonsType buttons { 
+        construct {
+            switch (value) {
+                case Gtk.ButtonsType.NONE:
+                    break;
+                case Gtk.ButtonsType.OK:
+                    add_button (_("_OK"), Gtk.ResponseType.OK);
+                    break;
```
danrabbit: `OK` is not an acceptable button label

lib/Widgets/MessageDialog.vala
```diff
@@ -0,0 +1,237 @@
...
+                case Gtk.ButtonsType.CLOSE:
+                    add_button (_("_Close"), Gtk.ResponseType.CLOSE);
+                    break;
+                case Gtk.ButtonsType.CANCEL:
+                    add_button (_("_Cancel"), Gtk.ResponseType.CANCEL);
+                    break;
+                case Gtk.ButtonsType.YES_NO:
+                    add_button (_("_No"), Gtk.ResponseType.NO);
+                    add_button (_("_Yes"), Gtk.ResponseType.YES);
```
danrabbit: "Yes" and "No" are not acceptable button labels

lib/Widgets/MessageDialog.vala
```diff
@@ -0,0 +1,237 @@
...
+                    break;
+                case Gtk.ButtonsType.OK_CANCEL:
+                    add_button (_("_Cancel"), Gtk.ResponseType.CANCEL);
+                    add_button (_("_OK"), Gtk.ResponseType.OK);
```
danrabbit: "OK" is not an acceptable button label

lib/Widgets/MessageDialog.vala
```diff
@@ -0,0 +1,237 @@
...
+            image_icon: new ThemedIcon (icon_name),
+            buttons: buttons
+        );
+    }
+
+    construct {
+        image = new Gtk.Image ();
+        image.valign = Gtk.Align.START;
+
+        primary_label = new Gtk.Label (null);
```
danrabbit: This label should probably be selectable as well

donadigo:
@danrabbit issues addresed.

donadigo:
@danrabbit you mean, add a new `transient` parameter to the constructors?

danrabbit:
@donadigo not necessarily. You could just set it in the demo, like `dialog.transient_for = window`. I think that'd be more manageable

donadigo:
@danrabbit done.

lib/Widgets/MessageDialog.vala
```diff
@@ -0,0 +1,240 @@
...
+     *  * {@link Gtk.ButtonsType.OK_CANCEL}
+     *
+     * If you wish to provide more specific actions for your dialog
+     * pass a {@link Gtk.ButtonsType.NONE} to {@link Granite.MessageDialog} and manually
+     * add those actions with {@link Gtk.Dialog.add_buttons} or {@link Gtk.Dialog.add_action_widget}.
+     */
+    public Gtk.ButtonsType buttons { 
+        construct {
+            switch (value) {
+                case Gtk.ButtonsType.NONE:
```
danrabbit: Thinking about it, `NONE` probably isn't a sane option because there would be no way to close the window

lib/Widgets/MessageDialog.vala
```diff
@@ -0,0 +1,240 @@
...
+     * Constructs a new {@link Granite.MessageDialog}.
+     * See {@link Gtk.Dialog} for more details.
+     * 
+     * @param primary_text the title of the dialog
+     * @param secondary_text the body of the dialog
+     * @param image_icon the {@link GLib.Icon} that is displayed on the left side of the dialog
+     * @param buttons the {@link Gtk.ButtonsType} value that decides what buttons to use, defaults to {@link Gtk.ButtonsType.NONE},
+     *        see {@link Granite.MessageDialog.buttons} on details and what values are accepted
+     */
+    public MessageDialog (string primary_text, string secondary_text, GLib.Icon image_icon, Gtk.ButtonsType buttons = Gtk.ButtonsType.NONE) {
```
danrabbit: `CLOSE` might be saner default here

lib/Widgets/MessageDialog.vala
donadigo: `NONE` has to be supported because, suppose you want to add your own buttons instead of the built-in ones.

lib/Widgets/MessageDialog.vala
donadigo: Done.

lib/Widgets/MessageDialog.vala
danrabbit: Oops you're right

lib/Widgets/MessageDialog.vala
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * The message grid that contains all the message widgets: the {@link Granite.MessageDialog.primary_label},
+     * {@link Granite.MessageDialog.secondary_label} and the image widget.
+     * 
+     * This grid can be used to add any custom widget to the message area such as a {@link Gtk.ComboBox} or {@link Gtk.CheckButton}.
+     * Note that after adding such widget you will need to call {@link Gtk.Widget.show} or {@link Gtk.Widget.show_all} on the widget itself for it to appear in the dialog.
+     */
+    public Gtk.Grid message_grid { get; construct; }
```
danrabbit: I'm not a huge fan of exposing the grid that contains the actual message labels for a couple reasons:
* It seems pretty non-intuitive that I have to start packing my widgets at `[1, 2]`. This requires me to know about the internal structure of the dialog and if I don't, I can attach widgets overlapping with the pre-populated ones or at the incorrect alignment.

* additional widgets packed into the dialog should be separated from the secondary text label by 24px (effectively, we double space between sections). It would be nice if this was done automatically for me instead of requiring me to know this.

I think it would be more intuitive to use a grid `content_area` and override Gtk.Dialog's `get_content_area` if possible. But if not, I think the publicly exposed grid should not include the icon and message text.

donadigo: @danrabbit it is not possible to override the `get_content_area` of the `Gtk.Dialog` class. We have 2 main solutions to this:

* Add a new special container inside the `message_grid` and expose it through the API.
* Keep the `message_grid` property and describe in the documentation it's details and structure including all the values like `[1, 2]`.

The first approach is what `Gtk.MessageDialog` uses: https://valadoc.org/gtk+-3.0/Gtk.MessageDialog.get_message_area.html, but with a difference that it's returning a `Gtk.Box` instead of a grid.

danrabbit: I think adding another container would be much better. I don't think it feels good to expect developers to know about the internal structure of the widget

lib/Widgets/MessageDialog.vala
```diff
@@ -0,0 +1,242 @@
...
+     * 
+     * See {@link Gtk.Dialog} for more details.
+     * 
+     * @param primary_text the title of the dialog
+     * @param secondary_text the body of the dialog
+     * @param icon_name the icon name to create the dialog image with
+     * @param buttons the {@link Gtk.ButtonsType} value that decides what buttons to use, defaults to {@link Gtk.ButtonsType.CLOSE},
+     *        see {@link Granite.MessageDialog.buttons} on details and what values are accepted
+     */
+    public MessageDialog.from_icon_name (string primary_text, string secondary_text, string icon_name = "dialog-information", Gtk.ButtonsType buttons = Gtk.ButtonsType.CLOSE) {
```
matzipan: Unclear from the name of the constructor what the icon is used for.

donadigo: @matzipan what name do you propose?

matzipan: Perhaps `with_image_from_icon_name`?

donadigo:
@matzipan @danrabbit changes addressed.

danrabbit:
Icon should not be nullable

donadigo:
@danrabbit reverted.

lib/Widgets/MessageDialog.vala
```diff
@@ -0,0 +1,261 @@
...
+     * If you want to add multiple widgets to this area, create a new container such as {@link Gtk.Grid} or {@link Gtk.Box} and then add it to the custom bin.
+     * 
+     * When adding a custom widget to the custom bin, the {@link Granite.MessageDialog.secondary_label}'s bottom margin will be expanded automatically
+     * to compensate for the additional widget in the dialog.
+     * Removing the previously added widget will remove the bottom margin.
+     * 
+     * If you don't want to have any margin between your custom widget and the {@link Granite.MessageDialog.secondary_label}, simply add your custom widget
+     * and then set the {@link Gtk.Label.margin_bottom} of {@link Granite.MessageDialog.secondary_label} to 0.
+     */
+    public Gtk.Bin custom_bin { get; construct; }
```
danrabbit: Using `Gtk.Bin` here sounds like a pain in the ass. Why not just use Grid and save having to create a grid?

donadigo: @danrabbit main problem is, having a Grid here disables us from having an ability to listen when a widget is added. E.g: calling `attach` on a Grid won't give us any signal that the widget was actually added. If we want a grid, the API user will have to set the margin on the secondary label manually.

lib/Widgets/MessageDialog.vala
```diff
@@ -0,0 +1,261 @@
...
+
+        secondary_label = new Gtk.Label (null);
+        secondary_label.use_markup = true;
+        secondary_label.selectable = true;
+        secondary_label.max_width_chars = 50;
+        secondary_label.wrap = true;
+        secondary_label.xalign = 0;
+
+        custom_bin = new SingleWidgetBin ();
+        custom_bin.add.connect (() => secondary_label.margin_bottom = 24);
```
danrabbit: You have a `row_spacing` of 6, so you need to subtract that and have this margin be `18`

lib/Widgets/MessageDialog.vala
danrabbit: Ah okay that makes sense :+1:

donadigo:
@danrabbit done.

danrabbit:
One last nitpick. Can we make this match the example dialog from the HIG? :) https://elementary.io/docs/human-interface-guidelines#dialogs

donadigo:
@danrabbit done.

PR #100, opened 2017-09-10
Add primary style class

danrabbit:
make sure we're using style constants for our styles


lib/StyleClass.vala
```diff
@@ -63,6 +63,10 @@ namespace Granite {
     public const string STYLE_CLASS_H3_LABEL = "h3";
     public const string STYLE_CLASS_H4_LABEL = "h4";
     public const string STYLE_CLASS_OVERLAY_BAR = "overlay-bar";
+    /**
+     * Style class for primary label text in a dialog
```
donadigo: Could we perhaps specifically say in what dialogs it's used in?
Like: `Style class for primary label text in a {@link Granite.MessageDialog}.`

Not sure if it's used anywehre else.

PR #102, opened 2017-09-11
add relative date time utils

### LINKED ISSUE #91: Create a relative date time function
We use relative date time logic in a lot of places. We should consolidate the logic into a method of `Granite.DateTime`.

This would replace the logic for:
- [Mail](https://github.com/elementary/mail/blob/rewrite/src/ConversationList/ConversationListItem.vala#L118)
- [Notification indicator](https://github.com/elementary/wingpanel-indicator-notifications/blob/master/src/Widgets/NotificationEntry.vala#L114)
- Some other third party apps like vocal.

danrabbit:
Fixes #91 

Adds the relative date time functions from Mail

For some reason relative date time doesn't seem to work properly until you change the date once. weird.

### DIFF
--- a/lib/DateTime.vala
+++ b/lib/DateTime.vala
@@ -88,6 +88,7 @@ namespace Granite.DateTime {
                     return ngettext ("%dh from now", "%dh from now", (ulong) rounded).printf (rounded);
                 }
             }
+
             return date_time.format (get_default_time_format (is_clock_format_12h (), false));
         } else if (is_same_day (date_time.add_days (1), now)) {
             return _("Yesterday");
--- a/lib/DateTime.vala
+++ b/lib/DateTime.vala
@@ -82,10 +82,10 @@ namespace Granite.DateTime {
                 diff = -1 * diff;
                 if (diff < TimeSpan.HOUR) {
                     var minutes = diff / TimeSpan.MINUTE;
-                    return ngettext ("%dm from now", "%dm from now", (ulong) (minutes)).printf ((int) (minutes));
+                    return ngettext ("in %dm", "in %dm", (ulong) (minutes)).printf ((int) (minutes));
                 } else if (diff < 12 * TimeSpan.HOUR) {
                     int rounded = (int) Math.round ((double) diff / TimeSpan.HOUR);
-                    return ngettext ("%dh from now", "%dh from now", (ulong) rounded).printf (rounded);
+                    return ngettext ("in %dh", "in %dh", (ulong) rounded).printf (rounded);
                 }
             }
 
--- a/lib/DateTime.vala
+++ b/lib/DateTime.vala
@@ -111,10 +111,7 @@ namespace Granite.DateTime {
      * @return true if day1 and day2 occur on the same day of the same year. False otherwise
      */
     public static bool is_same_day (GLib.DateTime day1, GLib.DateTime day2) {
-        if (day1.get_day_of_year () == day2.get_day_of_year () && day1.get_year () == day2.get_year ()) {
-            return true;
-        }
-        return false;
+        return day1.get_day_of_year () == day2.get_day_of_year () && day1.get_year () == day2.get_year ();
     }
 
     /**
--- a/lib/DateTime.vala
+++ b/lib/DateTime.vala
@@ -68,20 +68,32 @@ namespace Granite.DateTime {
         var diff = now.difference (date_time);
 
         if (is_same_day (date_time, now)) {
-            if (diff < TimeSpan.MINUTE) {
-                return _("Now");
-            } else if (diff < TimeSpan.HOUR) {
-                var minutes = diff / TimeSpan.MINUTE;
-                return ngettext ("%dm ago", "%dm ago", (ulong) (minutes)).printf ((int) (minutes));
-            } else if (diff < 12 * TimeSpan.HOUR) {
-                int rounded = (int) Math.round ((double) diff / TimeSpan.HOUR);
-                return ngettext ("%dh ago", "%dh ago", (ulong) rounded).printf (rounded);
+            if (diff > 0) {
+                if (diff < TimeSpan.MINUTE) {
+                    return _("Now");
+                } else if (diff < TimeSpan.HOUR) {
+                    var minutes = diff / TimeSpan.MINUTE;
+                    return ngettext ("%dm ago", "%dm ago", (ulong) (minutes)).printf ((int) (minutes));
+                } else if (diff < 12 * TimeSpan.HOUR) {
+                    int rounded = (int) Math.round ((double) diff / TimeSpan.HOUR);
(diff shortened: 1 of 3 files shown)


lib/DateTime.vala
```diff
@@ -56,6 +56,37 @@ namespace Granite.DateTime {
     }
 
     /**
+     * Compares a {@link GLib.DateTime} to {@link GLib.DateTime.now_local} and @return a localized, relative date and time string
+     */
+    public string get_relative_datetime (GLib.DateTime date_time) {
```
donadigo: * The function should be `static`.
* It would be better if you could pass a different `now` so you could calculate between different spans like with default being to `now_local`:

```
public static string get_relative_datetime (GLib.DateTime date_time, GLib.DateTime now = new GLib.DateTime.now_local ()) {
...
}
```

* You should make the documentation more extensive on what the function actually does, not just what it returns like: that it can return "Yesterday", "x h ago" etc.
* You should declare all the parameters and returns like this:
```
/**
 * ...
 * @param date_time the {@link GLib.DateTime} that represents the past point in time
 * @param now the {@link GLib.DateTime} that represents the current time, by default {@link GLib.DateTime.now_local}.
 * @return a localized, relative date and time string [possibly more details here]
 * 
```

lib/DateTime.vala
```diff
@@ -68,6 +99,16 @@ namespace Granite.DateTime {
     }
 
     /**
+     * Compare two {@link GLib.DateTime} and @return true if they occur on the same day of the same year
+     */
+    public bool is_same_day (GLib.DateTime day1, GLib.DateTime day2) {
```
donadigo: All of the points I made above apply here including, `static`, documentation, using proper `@param`'s and `@return`'s.
You should write here that it comapares the year as well as the day but nothing more.

lib/DateTime.vala
```diff
@@ -68,6 +99,16 @@ namespace Granite.DateTime {
...
+        if (day1.get_day_of_year () == day2.get_day_of_year () && day1.get_year () == day2.get_year ()) {
```
donadigo: This function could as well be:
`return day1.get_day_of_year () == day2.get_day_of_year () && day1.get_year () == day2.get_year ()`.

lib/DateTime.vala
```diff
@@ -56,6 +56,37 @@ namespace Granite.DateTime {
...
+                return date_time.format (format);
+            }
+        } else if (is_same_day (date_time.add_days (1), now)) {
+            return _("Yesterday");
+        } else if (diff < 6 * TimeSpan.DAY) {
+            return date_time.format ("%a");
+        } else if (date_time.get_year () == now.get_year ()) {
+            return date_time.format (_("%b %-e"));
+        } else {
+            return date_time.format ("%x");
```
matzipan: This should use `Granite.DateTime.get_default_date_format`

lib/DateTime.vala
```diff
@@ -56,6 +56,37 @@ namespace Granite.DateTime {
...
+            } else {
+                var format = Granite.DateTime.get_default_time_format (is_clock_format_12h (), false);
+                return date_time.format (format);
+            }
+        } else if (is_same_day (date_time.add_days (1), now)) {
+            return _("Yesterday");
+        } else if (diff < 6 * TimeSpan.DAY) {
+            return date_time.format ("%a");
+        } else if (date_time.get_year () == now.get_year ()) {
+            return date_time.format (_("%b %-e"));
```
matzipan: This should use `Granite.DateTime.get_default_date_format`

lib/DateTime.vala
danrabbit: Do you have a use case for a different frame of reference? This would complicate things significantly, none of the existing strings would make since without the frame of reference being "now".

donadigo: It's not really about "use case" here. You can't just assume that the right "now" moment will always be `DateTime.now_local`. There can be possibility that the API user will want to pass e.g: `now_utc`, `now` or even a custom DateTime that was constructed manually with getting the date from external sources like the Internet.

lib/DateTime.vala
danrabbit: Use of `%x` is intentional here. It is defined as "the preferred date representation for the current locale without the time"

lib/DateTime.vala
danrabbit: We shouldn't make this overly complicated for a feature that will never be used. Why do you think we can't assume that now local is what is wanted? Afaik that is what all of the existing examples are requesting. I think we need a solid example that not using now local is a valid desire before increasing complexity.

donadigo: It's would not be overly complicated. You set the default and that's the default. The user can call the API just as he could do this now, but if he wants to, he can provide an additional argument to set that.

danrabbit:
@matzipan can you provide an example of not using now as the frame of reference?

matzipan:
A public transport app where it says you have to catch the next bus "5 minutes later". A calendar app which notifies you your event has been rescheduled "1 hour later".

danrabbit:
The first example is a little confusing to me. Isn’t that “5 minutes from now”? If not, then “5 minutes later” from when?

How would you present this information to a user? Just the phrase “one hour later” doesn’t really seem to mean much without the surrounding context. Wouldn’t it be much clearer to use absolute time in this case and say “your meeting has been rescheduled for 5pm”?

matzipan:
You are looking for a route home and it shows you that you have to hop on this bus and then 5 minutes later get off and hop on another bus.

matzipan:
Well you could say "your meeting has been moved 1 hour later, at 5 pm"

danrabbit:
In that case, at least in English, you don't say "5 minutes later", you just mark the length of that step as "5 minutes". I could see a case for another function that gives you absolute times in natural langue, but printing absolute times is not what this function is for.

matzipan:
Well this function is absolute of duration + ago/from now

matzipan:
> I think proper grammar for the meeting example would be "Your meeting has been moved 1 hour, to 5pm"

Well it's unclear if it was moved 1 hour later or earlier :P

cassidyjames:
@matzipan implementing moment-from-now is super common, whereas just a super generic relative time library is a lot more complex. I'm happy merging in moment-from-now support.

donadigo:
In my opinion API != real use case. We are here talking about final examples that show in the UI while this is code and it's just really common for such API's to use a "from" and "to" points, even when there are no apparent use cases for them. This is limiting the developer to a set of things he can do, if he wants to do a little bit over that, he'll probably just not use this API at all because it doesn't allow him to do so.

btkostner:
I'm going to drop the the link to the [moment js implimentation](https://momentjs.com/docs/#/displaying/fromnow/) as it's well thought out, internationalization, and I had no problems using it in projects before.

Just as a tidbit: the 'ago' part of the string is a flag so the developer can pick if it's needed.

PR #121, opened 2017-10-20
Added support for Meson Build System

esodan:
I had to change Widgets.DroppedDelegate, to:

namespace Granite {
	public delegate void WidgetsDroppedDelegate ();
}
This makes GIR compilation works.


esodan:
Granite is not using a versioned installation, so any API/ABI break will affect any application relaying in it.
I suggest to change that, by using granite-0 for current unstable API, or granite-0.x for different API versions until 1.0 release is declared.

codygarver:
Thank you very much for your work, this pull request has been superseded by #177

PR #123, opened 2017-10-27
Granite.Color: Parse and to_string

Philip-Scott:
- Adds a two new construction methods for passing a string, and passing a Gdk.RGBA on constructors
- adds a `to_string ()` method that returns the contents of the class to a CSS compatible string

Question: 

Should we deprecate the `Gdk.Color` constructor?

### DIFF
--- a/lib/Drawing/Color.vala
+++ b/lib/Drawing/Color.vala
@@ -27,7 +27,7 @@ namespace Granite.Drawing {
      * A class containing an RGBA color and methods for more powerful color manipulation.
      */
     public class Color : GLib.Object, SettingsSerializable {
-    
+
         /**
          * The value of the red channel, with 0 being the lowest value and 1.0 being the greatest value.
          */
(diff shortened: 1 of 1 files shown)


lib/Drawing/Color.vala
```diff
@@ -441,16 +461,38 @@ namespace Granite.Drawing {
...
+         *
+         * r, g, and b are represented as integers in the range 0 to 255, and a is represented as
+         * floating point value in the range 0 to 1.
+         *
+         * Note that this string representation may lose some precision, since r, g and b are represented
+         * as 8-bit integers. If this is a concern, you should use a different representation.
+         *
+         * @return the text string
+         */
+        public string to_string () {
```
donadigo: Please mention the `Gdk.RGBA` struct here, along with the link to the `Gdk.RGBA.to_string` method so that people know how this is converted to string.

lib/Drawing/Color.vala
```diff
@@ -54,311 +54,331 @@ namespace Granite.Drawing {
...
+        public Color.from_rgba (Gdk.RGBA color) {
+            set_from_rgba (color);
+        }
+
+        /**
+         * Constructs a new {@link Granite.Drawing.Color} from a string.
+         *
+         * @param color the string specifying the color
+         */
+        public Color.from_string (string color) {
```
donadigo: Mention the `Granite.Color.to_string` here since current description doesn't tell anything about the format.

`For more details on formatting and how this function works see {@link Granite.Color.to_string}`

PR #128, opened 2017-11-04
Add a visible child name property to SettingsSidebar

danrabbit:
This unbreaks search in Switchboard. Setting the visible child in the stack seems to get overridden by what is selected in the sidebar. This allows us to set the selected child in the sidebar and thus the visible child in the stack

### DIFF
--- a/lib/Widgets/SettingsSidebar.vala
+++ b/lib/Widgets/SettingsSidebar.vala
@@ -35,9 +35,15 @@ public class Granite.SettingsSidebar : Gtk.ScrolledWindow {
     /**
      * The name of the currently visible Granite.SettingsPage
      */
-    public string visible_child_name {
+    public string? visible_child_name {
         get {
-            return ((SettingsSidebarRow) listbox.get_selected_row ()).name;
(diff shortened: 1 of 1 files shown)


lib/Widgets/SettingsSidebar.vala
```diff
@@ -33,6 +33,22 @@ public class Granite.SettingsSidebar : Gtk.ScrolledWindow {
     public Gtk.Stack stack { get; construct; }
 
     /**
+     * The name of the currently visible Granite.SettingsPage
+     */
+    public string visible_child_name {
```
donadigo: This should be nullable.

lib/Widgets/SettingsSidebar.vala
```diff
@@ -33,6 +33,22 @@ public class Granite.SettingsSidebar : Gtk.ScrolledWindow {
...
+        get {
+            return ((SettingsSidebarRow) listbox.get_selected_row ()).name;
```
donadigo: Before returning that, check if the `get_selected_row` doesn't return `null`. If it did, return null also.

PR #137, opened 2017-12-14
[SettingsSidebar] Do not rely on child names, make the SimpleSettingsPage relyable

tintou:


danrabbit:
That's much smarter to connect to the property change signal than to override the properties. Do those properties need to be marked as deprecated though? Couldn't that just be removed and the API won't change because that property will be inherited from the parent class?

Also way smarter to construct the row from a settings page and do all the binding and everything inside that class.

tintou:
@danrabbit Removing it will create an API change at the C side (meaning binary breakage for vala apps). The `new` keyword put there makes that there are both:
`const gchar* granite_settings_page_get_icon_name (GraniteSettingsPage* self);`
and
`const gchar* granite_simple_settings_page_get_icon_name (GraniteSimpleSettingsPage* self);`

So people need to cast Granite.SimpleSettingsPage into Granite.SettingsPage to make it use the right function

lib/Widgets/AbstractSimpleSettingsPage.vala
```diff
@@ -96,12 +98,10 @@ public abstract class Granite.SimpleSettingsPage : Granite.SettingsPage {
...
      * Creates a new SimpleSettingsPage
+     * Deprecated: Subclass this instead.
      */
     public SimpleSettingsPage () {
-        Object (activatable: activatable,
-                icon_name: icon_name,
-                description: description,
-                title: title);
+        
     }
```
jeremypw: You cannot instantiate an abstract class anyway so this (empty) creator function should be removed. The preceding comments are redundant.  But why is this class abstract an abstract class?  It does not contain any abstract methods nor does its parents.

tintou: Yeah I fully agree with you, but it's too late (or too early) in the cycle to introduce an API break so i'll keep it like this right now

PR #138, opened 2017-12-20
Add methods from converting the color to integer

donadigo:
Adds a set of methods that can help with serializing and storing the Granite.Drawing.Color as an integer.
This will be useful for https://github.com/elementary/granite/pull/122.


lib/Drawing/Color.vala
```diff
@@ -103,6 +143,27 @@ namespace Granite.Drawing {
             set_from_rgba (rgba);
         }
 
+        /**
+         * Conststructs a new {@link Granite.Drawing.Color} from an integer.
```
jeremypw: Spelling - "Constructs"

lib/Drawing/Color.vala
```diff
@@ -103,6 +143,27 @@ namespace Granite.Drawing {
...
+         * //*_from_int// static method collection such as {@link Granite.Drawing.Color.alpha_from_int}.
+         * 
+         * @param color the integer specyfying the color
+         */
+        public Color.from_int (int color) {
+            R = (double)red_from_int (color) / uint8.MAX;
+            G = (double)green_from_int (color) / uint8.MAX;
+            B = (double)blue_from_int (color) / uint8.MAX;
+            A = (double)alpha_from_int (color) / uint8.MAX;
+        }
```
jeremypw: Will Vala always automatically do the correct conversions? Else should be:
 `(double)red)_from_int (color) / (double)uint8.MAX`
 etc.

donadigo:
@jeremypw thanks for the review! Changes applied.

PR #140, opened 2018-01-21
Handle negative seconds in seconds_to_time function

markodp:
+ remove normalize_time and use printf format option instead

### DIFF
--- a/lib/DateTime.vala
+++ b/lib/DateTime.vala
@@ -180,22 +180,20 @@ namespace Granite.DateTime {
      * @return returns an ISO 8601 formatted string
      */
     public static string seconds_to_time (int seconds) {
+        int sign = 1;
+        if (seconds < 0) {
+            seconds = - seconds;
+            sign = -1;
+        }
+
         int hours = seconds / 3600;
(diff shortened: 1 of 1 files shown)


davidmhewitt:
I approve of the removal of that `normalize_time` method. You definitely did it much better with printf formatting. However, why does this function need to handle negative seconds?

The ISO8601 standard has no allowances for negative time. Not that this function is fully compliant with ISO8601 anyway.

markodp:
Since the parameter is an int, which can be negative, why not also display the output properly :)
Before it was displayed like 0-10:0-9 now it shows -10:09 which looks way better. Maybe to display time remaining, or something, not

lib/DateTime.vala
```diff
@@ -180,22 +180,20 @@ namespace Granite.DateTime {
      * @return returns an ISO 8601 formatted string
      */
     public static string seconds_to_time (int seconds) {
+        int sign = 1;
+        if (seconds < 0) {
+            seconds = - seconds;
```
donadigo: Remove space after `-` here.

PR #141, opened 2018-01-23
Create AsyncNetworkImage Widget.

harisvsulaiman:
This code is something i hacked together with code from opensource apps from Nathan and Extending AsyncImage Widget for an app im working on.

The code needs to be cleaned up and some resources needs to handled properly.

This is a proposal to see if this kind of widget is something the elementary team is looking for.

Edit: This requires libsoup.


leonardo-lemos:
Changing this to "In Progress" since it needs more work to get it ready for merge.

harisvsulaiman:
@donadigo can you review this?

tintou:
I have NOT reviewed this code at all, but just note that it's not possible to include this into Granite because it's not under the LGPLv3+ license.

harisvsulaiman:
@tintou I don't know how i would license the code, the diskwatcher util is from @nathandyer's apps code. 
It was under Under that license, so i included it by default.

I believe this is a widget which is necessary for a lot of devs, and it's something i just hacked together,I hope that someone puts forward an alternative solution.

donadigo:
@harisvsulaiman hello. Aside from reviewing this, it looks like this branch is trying to accomplish what `set_from_file_async` in `AsyncImage` is already doing. You can already asynchronously load an image from a remote URL like this:
`async_image.set_from_file_async (File.new_for_uri ("http://...") ...)`

Aside from that, if you want more control over the cache, you should implement additional file caching in the `set_from_file_async` method itself.

The same goes for the spinner, if you want to show a spinner in the place of the loading image, you should also expose that API directly in the `AsyncImage` class.

The branch has a few showstoppers:
* it's based on libsoup, which in return would mean that every Granite app has a libsoup dependency
* lack of documentation
* reimplements already existing cache in `AsyncImage` class
* caches files to a user directory under some unknown name "splosh"
* omits scale factor changes: `get_style_context ().set_scale (1);`

harisvsulaiman:
@donadigo Thank you for the detailed review.
I didn't know some of these methods ( eg: File.new_for_uri() ) existed.We should have better documentation than valadoc, We can have a Elementary os guide like the meteor guide where the best practices are mentioned.


I'm going to close this PR.

PR #144, opened 2018-01-27
Update symbols

ricotz:


tintou:
Got the same results with `dpkg-gensymbols`

Issue #67, opened 2017-06-17
ModeButton can be untoggled


markodp:
If you click on mode that is selected then no mode is shown as active anymore. Sometimes this is wanted, but sometimes you need one of modes to be always on.

Some examples where this issue can be observed:
- view modes in Files
- Primary button section in Mouse & Touchpad settings

markodp:
My suggestion would be to add new bool property that would:
- allow case for all modes being deselected, or
- enforce one mode to be always selected

peteruithoven:
Example:
![snaptastic buttons](https://user-images.githubusercontent.com/523210/35806600-567aebe2-0a80-11e8-83af-f000f98aca5e.gif)

fabiozaramella:
Since it's called ModeButton, it expresses a mode of being so it shouldn't be possible for a normal user to disable it, the only way to do it should be programmatically.

PR #156, opened 2018-02-18
Prevent ModeButton deselection (for #67)

jlnr:
This is my attempt at fixing the cosmetic glitch in #67 without breaking backward compatibility. However:

* I'm not sure if `can_deselect` is the best name for this property within the larger Gtk universe.
* I'm also not sure if deselection is a supported feature of the ModeButton right now. The `selected` property is not updated to -1 when it happens and it can't be observed using the `mode_changed` signal, and "fixing" that could break apps which don't expect deselection to happen.
  If no app relies on the current deselection behavior, it might be better to *always* prevent deselection instead of making it configurable. That would also fix affected apps without changing their code.
* The code I've added relies very much on the order of execution to distinguish user events from programmatic deselection. It's probably a bit too "clever" :/

Feedback welcome, I'll be happy to iterate on this.


jlnr:
After using eOS for a bit, I haven't seen any use of ModeButton where deselection did anything, so I've removed the `can_deselect` property. This PR now changes the behavior of a ModeButtons.

jlnr:
I've fixed all single-line if statements and removed trailing whitespace in one line.

PR #160, opened 2018-02-28
Remove several deprecation warnings and useless functions

tintou:


CMakeLists.txt
```diff
@@ -37,7 +37,7 @@ set (PKG_DEPS
     gee-0.8)
 
 set (PKG_DEPS_CHECK
-    gtk+-3.0>=3.14
+    gtk+-3.0>=3.22
```
jeremypw: This breaks Travis CI atm, which installs v3.18

PR #162, opened 2018-03-09
Deprecate overlay property in OverlayBar

### LINKED ISSUE #110: Overlaybar should not own reference to overlay
Because Overlaybar holds an owned reference to the overlay it is in, and vice-versa, a circular reference is created preventing automatic destruction of these widgets and hence possible memory leaks.

jeremypw:
Fixes #110 

This is a compromise solution with less chance of side-effects by keeping constuctor signature the same.

Will be followed by a further PR to remove overlay as constructor parameter in the interests of maintainability.

### DIFF
--- a/lib/Widgets/OverlayBar.vala
+++ b/lib/Widgets/OverlayBar.vala
@@ -34,8 +34,9 @@
  * the widget.
  *
  * For this widget to function correctly, the event {@link Gdk.EventMask.ENTER_NOTIFY_MASK} must be set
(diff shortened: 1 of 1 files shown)


jeremypw:
Is this better?

tintou:
This looks way better to me as this means that there is not API change.
(the `replacement` property in the Version tag is optional, if it's empty you don't have to specify it)
We have to check that we are never using the `overlay` property in our code though

jeremypw:
OK.  It is hard to imagine why you would want to use this property as if you have access to the OverlayBar you already have access to its parent.

PR #163, opened 2018-03-09
No constructor parameter for OverlayBar

### LINKED ISSUE #110: Overlaybar should not own reference to overlay

jeremypw:
Fixes #110.

This is a more far-reaching solution based on PR #162.   This would require refactoring of all apps that use the Granite.OverlayBar widget so that they add the widget to the overlay themselves and enable enter notify events if required.

### DIFF
--- a/demo/Views/OverlayBarView.vala
+++ b/demo/Views/OverlayBarView.vala
@@ -28,9 +28,13 @@ public class OverlayBarView : Gtk.Overlay {
         grid.valign = Gtk.Align.CENTER;
         grid.add (button);
 
-        var overlaybar = new Granite.Widgets.OverlayBar (this);
+        /* Need to listen to ENTER events to enable overlaybar dodge function */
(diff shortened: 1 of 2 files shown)


jeremypw:
I did change the preamble to the class, but I am not sure what needs to be done to update valadocs.

tintou:
I'm worried that this is an API break.
Can you do:
`public Gtk.Overlay overlay { get; construct; }` and mark it as deprecated
and set it nullable, it shouldn't change anything from the C point of view.
On the same page, keep the constructor but set the overlay nullable:
`public OverlayBar (Gtk.Overlay? overlay = null)`
That way the transition should be smooth.

jeremypw:
I have incorporated these changes into #110, so closing this PR.

PR #164, opened 2018-03-20
DynamicNotebook: make the keybinding trigger the right signal

tintou:
Because of this, the restore features were not always working with `Ctrl` + `w` (especially with Code)


tintou:
See https://github.com/elementary/code/pull/368 for a complete test in Code

PR #167, opened 2018-04-11
AlertView: Hide icon if no icon name

### LINKED ISSUE #166: Make icon nullable in alertview
Sometimes we want to create an alertview without an icon because we've already got one on screen for this page (Like list boxes inside a SimpleSettingsPage). To cater to that we should make the icon nullable and default to null. While the icon is null, all text should be center aligned.

danrabbit:
Fixes #166 

Also get rid of a random eventbox

### DIFF
--- a/lib/Widgets/AlertView.vala
+++ b/lib/Widgets/AlertView.vala
@@ -78,7 +78,11 @@ public class Granite.Widgets.AlertView : Gtk.Grid {
             return image.icon_name;
         }
         set {
-            image.set_from_icon_name (value, Gtk.IconSize.DIALOG);
+            if (value != null && value != "") {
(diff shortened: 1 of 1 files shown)


lib/Widgets/AlertView.vala
```diff
@@ -73,12 +73,16 @@ public class Granite.Widgets.AlertView : Gtk.Grid {
     /**
      * The icon name
      */
-    public string icon_name {
+    public string? icon_name {
         owned get {
             return image.icon_name;
```
jeremypw: It it valid to reference a property of a (possibly) destroyed widget?  Do we need to be able to set the icon name after construction of the AlertView?

PR #175, opened 2018-05-09
Fixes to comply with valac 0.41

ricotz:


lib/Widgets/DynamicNotebook.vala
```diff
@@ -933,11 +941,6 @@ namespace Granite.Widgets {
         }
 
         ~DynamicNotebook () {
```
BaptisteGelez: Any reason to keep this block now it is empty?

ricotz: No particular reason, I guess it doesn't hurt to keep it for future use.

PR #180, opened 2018-05-22
Bump soname for 5.0 release

codygarver:


ricotz:
Might be a good time to bump the build requirement to valac >= 0.40.0 and use --abi-stability

Was there some source clean up like reordering fields and methods? E.g. like avoiding public fields at all.

ricotz:
--abi-stability is a buildflag available since 0.40 to keep symbol ordering of original vala source

PR #181, opened 2018-05-22
Add libgranite5

codygarver:


codygarver:
Closes #182

ricotz:
The -dbg package should be dropped in favour of the auto-generated -dbgsym package

PR #187, opened 2018-05-28
Add tooltip support for SourceList

### LINKED ISSUE #178: Tooltips on SourceList
Correct me if I'm wrong, but it looks like we don't have a way of setting/displaying tooltips on SourceList items, so for items that are truncated because they're in a sidebar, we can't show the full name on hover.

Raised on behalf of: https://github.com/elementary/code/issues/343

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58550582-tooltips-on-sourcelist?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #178 

Adds two properties to `SourceList.Item` that allow us to specify tooltips for both the tooltip item itself, and the activatable button that sits to the right of it (example use case would be the eject button next to an item in Files).

By default, if a tooltip isn't set for the item itself, it displays the item's name for the case where the name in the list may be truncated due to the SourceList being too narrow.

I've added an activatable icon to one of the items in the demo and given it a tooltip to demonstrate the feature.


davidmhewitt:
@jeremypw Thanks for the review! Good point with the keyboard tooltips, that CTRL+F1 shortcut wasn't something I was aware of. So I've learnt something new and that should be fixed now.

As for the positioning of them. That's fully under the control of GTK, we have no control of that. I assume that since a TreeView is a single Widget and the items within it aren't strictly GTK widgets, it can't accurately place the tooltips.

It looks as though if it can't fit the tooltip inside the bounds of the treeview close to the mouse cursor, it has some arbitrary algorithm to find the next best place relative to the treeview, which is obviously a larger widget than the one we're trying to target, so it's not quite accurate.

I can't really think what to suggest to improve that. It's either put up with the not so well positioned tooltips due to limitations in the toolkit or not have tooltips at all? Which I guess is a UX decision.

snowparrot:
ping

seanballais:
@davidmhewitt, I know it's been long but any progress on this? #178 blocks any attempt in fixing elementary/code#343.

davidmhewitt:
I've merged master so this branch is less stale, but I have no interest in working on this anymore. But will leave it here in case someone else wants to take it over.

I seem to remember that I spent quite a while on this and couldn't resolve the issues mentioned above. I was beginning to suspect a GTK bug I think, but can't remember now.

jeremypw:
It may be that the remaining issues are not sufficiently serious to out-weigh the benefits - I'll have another look at this soon.

jeremypw:
I have made it so that the tooltip appears on hovering the icon as well as the text. I think this can be merged now if there is no objection.

PR #193, opened 2018-06-26
Replace ParseArguments with CMakeParseArguments

yangfl:
I'm aware of the plan to replace CMake with Meson. But since it looks like we won't get rid of CMake completely for the short time, I think this pacth is still helpful.


codygarver:
What problem does this solve?

PR #195, opened 2018-07-17
Add Granite.ModeSwitch

### LINKED ISSUE #192: ModeSwitch
I count at least three apps using this widget now. None are core, but three are AppCenter apps: LookBook, Principles, and Aesop. In my book that's enough to pull this into Granite, but I'm open to discussion.

danrabbit:
Fixes #192 

Adds Granite.ModeSwitch and uses it in the demo

### DIFF
--- a/lib/Widgets/ModeSwitch.vala
+++ b/lib/Widgets/ModeSwitch.vala
@@ -1,0 +1,127 @@
+/*
+* Copyright (c) 2018 elementary, Inc. (https://elementary.io)
+*
+* This program is free software; you can redistribute it and/or
+* modify it under the terms of the GNU General Public
+* License as published by the Free Software Foundation; either
+* version 2 of the License, or (at your option) any later version.
+*
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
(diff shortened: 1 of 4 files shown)


lib/Widgets/ModeSwitch.vala
```diff
@@ -0,0 +1,127 @@
...
+        add (mode_switch);
+        add (secondary_icon_box);
+
+        bind_property ("primary-icon-name", primary_icon, "icon-name", GLib.BindingFlags.SYNC_CREATE);
+        bind_property ("primary-icon-tooltip-text", primary_icon, "tooltip-text");
+        bind_property ("secondary-icon-name", secondary_icon, "icon-name", GLib.BindingFlags.SYNC_CREATE);
+        bind_property ("secondary-icon-tooltip-text", secondary_icon, "tooltip-text");
+
+        this.notify["active"].connect (() => {
+            if (Gtk.StateFlags.DIR_RTL in get_state_flags ()) {
```
donadigo: Can we have these two lambdas (`this.notify["active"].connect (() => ...` and ` mode_switch.notify["active"].connect (() => ...` put in one method?

danrabbit:
Added a way to construct using GLib.Icons instead of icon names.

The two lambdas kind of do the opposite thing, translating from the "active" state of the ModeSwitch widget to the internal "active" state of the Gtk.Switch. The problem is making sure the switch points the right direction when using RTL

PR #205, opened 2018-08-12
Add show_error_expander method

### LINKED ISSUE #203: MessageDialog: easily add error details expander
At least two apps (AppCenter and Files) are using this pattern to add extended error information to dialogs. We should probably make it easy to just pass a string to Granite.MessageDialog and have it do the right thing.

Relevant code: https://github.com/elementary/appcenter/blob/master/src/Dialogs/UpdateFailDialog.vala#L34

32525261-72c9b0a6-c3d8-11e7-866e-ba4e6148f64e

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62048782-messagedialog-easily-add-error-details-expander?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github).
</bountysource-plugin>

donadigo:
Fixes #203.

Adds a `show_error_expander` method to MessageDialog to show the expander with the terminal view to show the detailed error message.


lib/Widgets/MessageDialog.vala
```diff
@@ -253,4 +263,45 @@ public class Granite.MessageDialog : Gtk.Dialog {
...
+     * to {@link Granite.MessageDialog.custom_bin}. Calling this method with a manually added widget
+     * to {@link Granite.MessageDialog.custom_bin} will result in a critical message and no
+     * addition.
+     * 
+     * Subsequent calls to this method will change the error message to a new one.
+     * 
+     * @param error_message the detailed error message to display
+     */
+    public void show_error_expander (string error_message) {
+        if (custom_bin.get_child () != expander) {
```
danrabbit: I don't think these should conflict. There might be a situation where you want to use both

lib/Widgets/MessageDialog.vala
```diff
@@ -253,4 +263,45 @@ public class Granite.MessageDialog : Gtk.Dialog {
...
+     * terminal-like text view. Note that this method adds the widget itself
+     * to {@link Granite.MessageDialog.custom_bin}. Calling this method with a manually added widget
+     * to {@link Granite.MessageDialog.custom_bin} will result in a critical message and no
+     * addition.
+     * 
+     * Subsequent calls to this method will change the error message to a new one.
+     * 
+     * @param error_message the detailed error message to display
+     */
+    public void show_error_expander (string error_message) {
```
danrabbit: I think this should be named something that doesn't imply the UI in case we want to change the way errors are presented in the future

lib/Widgets/MessageDialog.vala
```diff
@@ -253,4 +263,45 @@ public class Granite.MessageDialog : Gtk.Dialog {
...
+            critical ("Cannot add an error expander to non-empty custom bin.");
+            return;
+        }
+
+        if (details_view == null) {
+            details_view = new Gtk.TextView ();
+            details_view.editable = false;
+            details_view.pixels_below_lines = 3;
+            details_view.wrap_mode = Gtk.WrapMode.WORD;
+            details_view.get_style_context ().add_class ("terminal");
```
danrabbit: There's a `Granite.STYLE_CLASS_TERMINAL` for this. As a general rule, we should always use constants for style classes in granite to make sure that our style classes are stable

danrabbit:
Screenshot for reference:
screenshot from 2018-08-27 17 02 48 2x


- [ ] I think we need to make sure to double space between the details area and the custom bin as well.

- [ ] There should probably be some kind of padding inside the terminal area. This would be the `border_width` property in this case. Probably 6 or 12px

PR #211, opened 2018-09-03
Issue 208 add show seconds property

### LINKED ISSUE #208: Add a `show_seconds` property
I want to use TimePicker in Photos but I need to handle seconds. We should add a property to TimePicker that allows us to show seconds.

It would also be convenient to have a constructor like `new Granite.TimePicker.with_seconds ()`

FriendOfEntropy:
Fixes #208 

- Added new property **show_seconds**.
- Added spinbutton for seconds and refactored the parse text algorithm to support seconds
- Added new constructor "with_second".
- Added switch to Granite-Demo to allow changing the show_seconds property. Poor UX, I'm sure, but this is just a demo, and I didn't want to blow out the grid to multiple columns of labels or split into more vertical groups.  A tooltip describes that the switch is for seconds.
- Added a second TimePicker control to Granite-Demo that utilizes the with_second constructor.

spinbuttons

show_seconds_tooltip

show_seconds_switch_on

There are still a few issues with parse text, but I used several test cases from https://github.com/elementary/granite/pull/115 and confirmed that while I don't believe I introduced any new regressions or issues, the deficiencies I found in testing already existed before I started making this change.

The parse routine was quite complex before, and is now quite nested.  If it is critical to continue to support all these various loose delimitless formats (7AM, etc) and we want to try to resolve these pre-existing cases below, we might want to consider masking or otherwise hooking into the Gtk.Entry insert_text signal, or even changing to regex-based parsing to handle all these various lengths.

When this project is converted to meson, I would love to refactor the parse routine to either extract it to the datetime utility class, or at the least make it public within the TimePicker and with a return value of string so that we could add a TestRunner and introduce a lot of unit tests.  I realize that most components in this project are UI widgets, but this parse in particular could benefit greatly with automated testing.

```

AM/PM - Show Seconds false

    Input           Parsed Result
    12:35 PM        12:35 PM
     5:23 AM         5:23 AM
    03:45           03:45 AM
    0345            03:45 AM
    1230            12:30 PM
    1230 AM          6:00 AM   (pre-existing issue)
    1230AM           6:00 AM   (pre-existing issue)
    7AM              7:00 AM
    6 23AM           6:23 AM
    623AM           11:00 PM   (pre-existing issue)
    sdfgh           12:00 AM


AM/PM - Show Seconds true

    Input           Parsed Result
    12:35 PM        12:35:00 PM
     5:23 AM         5:23:00 AM
    03:45           03:45:00 AM
    0345            03:45:00 AM
    1230            12:30:00 PM
    1230 AM          6:00:00 AM   (pre-existing issue)
    1230AM           6:00:00 AM   (pre-existing issue)
    01:23:35 AM      1:23:35 AM
    7AM              7:00:00 AM
    6 23AM           6:23:00 AM
    623AM           11:00:00 PM   (pre-existing issue)
    sdfgh           12:00:00 AM


24 Hour - Show Seconds false

    Input           Parsed Result
    12:35 PM        12:35
     5:23 AM        05:23
     5:23 PM        17:23
    03:45           03:45
    0345            03:45
    1545            15:45
    1230            12:30
    1230 AM         06:00   (pre-existing issue)
    1230AM          06:00   (pre-existing issue)
    7AM             07:00
    6PM             18:00
    6 23AM          06:23
    623AM           23:00   (pre-existing issue)
    sdfgh           00:00



24 Hour - Show Seconds true

    Input           Parsed Result
    12:35 PM        12:35:00
     5:23 AM        05:23:00
     5:23 PM        17:23:00
    03:45:13        03:45:13
    034513          03:45:13
    0345            03:45:00
    03:45           03:45:00
    1545            15:45:00
    123056          12:30:56
    1230 AM         05:00:00   (pre-existing issue)
    1230AM          06:00:00   (pre-existing issue)
    7AM             07:00:00
    6PM             18:00:00
    6 27AM          06:27:00
    627AM           03:00:00
    sdfgh           00:00:00
```

### DIFF
--- a/demo/Views/DateTimePickerView.vala
+++ b/demo/Views/DateTimePickerView.vala
@@ -46,11 +46,9 @@ public class DateTimePickerView : Gtk.Grid {
 		
 		timepicker = new Granite.Widgets.TimePicker ();
 		
-		var show_seconds_label = new Gtk.Label ("Seconds:");        
-		show_seconds_label.halign = Gtk.Align.END;
-
 		show_seconds = new Gtk.Switch ();
 		show_seconds.halign = Gtk.Align.START;
+		show_seconds.set_tooltip_text ("Show Seconds");
 		show_seconds.notify["active"].connect (() => {
 			timepicker.show_seconds = show_seconds.active;
 		});
@@ -61,6 +59,7 @@ public class DateTimePickerView : Gtk.Grid {
 		time_label_with_seconds.halign = Gtk.Align.END;
 
 		timepicker_with_seconds = new Granite.Widgets.TimePicker.with_second (true);
+		timepicker_with_seconds.set_tooltip_text ("TimePicker Example using with_second create method");
 
 		var formatting_label = new Gtk.Label ("String Formatting");
 		formatting_label.margin_top = 6;
@@ -110,8 +109,7 @@ public class DateTimePickerView : Gtk.Grid {
 		attach (time_label, 0, 2, 1, 1);
 		attach (timepicker, 1, 2, 1, 1);
 		
-		attach (show_seconds_label, 0, 3, 1, 1);
-		attach (show_seconds, 1, 3, 1, 1);
+		attach (show_seconds, 2, 2, 1, 1);
 
 		attach (time_label_with_seconds, 0, 4, 1, 1);
 		attach (timepicker_with_seconds, 1, 4, 1, 1);
--- a/demo/Views/DateTimePickerView.vala
+++ b/demo/Views/DateTimePickerView.vala
@@ -25,117 +25,117 @@
  */
 
 public class DateTimePickerView : Gtk.Grid {
-	private Gtk.Label relative_datetime;
-	private Granite.Widgets.DatePicker datepicker;
-	private Granite.Widgets.TimePicker timepicker;
-	private Granite.Widgets.TimePicker timepicker_with_seconds;
-	private Gtk.Switch show_seconds;
+    private Gtk.Label relative_datetime;
+    private Granite.Widgets.DatePicker datepicker;
+    private Granite.Widgets.TimePicker timepicker;
+    private Granite.Widgets.TimePicker timepicker_with_seconds;
+    private Gtk.Switch show_seconds;
 
-	construct {
-		var pickers_label = new Gtk.Label ("Picker Widgets");
-		pickers_label.xalign = 0;
-		pickers_label.get_style_context ().add_class (Granite.STYLE_CLASS_H4_LABEL);
+    construct {
+        var pickers_label = new Gtk.Label ("Picker Widgets");
+        pickers_label.xalign = 0;
+        pickers_label.get_style_context ().add_class (Granite.STYLE_CLASS_H4_LABEL);
 
-		var date_label = new Gtk.Label ("DatePicker:");
-		date_label.halign = Gtk.Align.END;
+        var date_label = new Gtk.Label ("DatePicker:");
+        date_label.halign = Gtk.Align.END;
 
-		datepicker = new Granite.Widgets.DatePicker ();
+        datepicker = new Granite.Widgets.DatePicker ();
 
-		var time_label = new Gtk.Label ("TimePicker:");
-		time_label.halign = Gtk.Align.END;
-		
-		timepicker = new Granite.Widgets.TimePicker ();
-		
-		show_seconds = new Gtk.Switch ();
-		show_seconds.halign = Gtk.Align.START;
-		show_seconds.set_tooltip_text ("Show Seconds");
-		show_seconds.notify["active"].connect (() => {
-			timepicker.show_seconds = show_seconds.active;
-		});
-		show_seconds.set_active (false);        
+        var time_label = new Gtk.Label ("TimePicker:");
+        time_label.halign = Gtk.Align.END;
+        
+        timepicker = new Granite.Widgets.TimePicker ();
+        
+        show_seconds = new Gtk.Switch ();
+        show_seconds.halign = Gtk.Align.START;
+        show_seconds.set_tooltip_text ("Show Seconds");
+        show_seconds.notify["active"].connect (() => {
+            timepicker.show_seconds = show_seconds.active;
+        });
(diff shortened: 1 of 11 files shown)


FriendOfEntropy:
Yes, apologies for making a mess of my first try at a real PR.  I also failed to see your cleanup in https://github.com/elementary/granite/pull/210/commits so I have incorporated your changes there, and will try to figure out how to resubmit.

FriendOfEntropy:
Okay, hopefully I've cleaned up my mess and it's really ready for review...thanks for your patience.

FriendOfEntropy:
Fixed regression issue concerning the with_format.  If the with_format constructor is used, it is respected again.  If that format constructor is used, however, the show_seconds property is not respected in the view...the custom format that was set at instantiation overrides.  

Since the show_seconds won't apply there because of mutual exclusion, I wonder if we should just have the with_format and with_second constructors and not expose a show_seconds property?  That way, you must choose at creation time and your behavior is locked.  

Is there a use-case to flip the show_seconds property post-creation?  If so, then I guess I should just document that property heavily and let the consumer of the widget know that show_seconds is not relevant if you used the with_format constructor?

lib/Widgets/TimePicker.vala
```diff
@@ -193,35 +238,65 @@ namespace Granite.Widgets {
...
-            Object (format_12: format_12, format_24: format_24);
+            Object (custom_format_12: format_12, custom_format_24: format_24);
+        }
+
+        /**
+         * Creates a new TimePicker.
+         *
+         * @param with_seconds Should the format include seconds. For example: "%l:%M:%S %p" or "%H:%M:%S".
+         */
+        public TimePicker.with_second (bool with_second) {
```
danrabbit: I don't think this should take an argument. It probably doesn't make sense to say `TimePicker.with_seconds (false)`. That would just be the same as doing `TimePicker ()`

demo/Views/DateTimePickerView.vala
```diff
@@ -41,8 +44,29 @@ public class DateTimePickerView : Gtk.Grid {
 
         var time_label = new Gtk.Label ("TimePicker:");
         time_label.halign = Gtk.Align.END;
-
+
```
danrabbit: Looks like you accidentally added a bunch of spaces here

demo/Views/DateTimePickerView.vala
```diff
@@ -41,8 +44,29 @@ public class DateTimePickerView : Gtk.Grid {
...
         timepicker = new Granite.Widgets.TimePicker ();
+        
+        show_seconds = new Gtk.Switch ();
+        show_seconds.halign = Gtk.Align.START;
+        show_seconds.set_tooltip_text ("Show Seconds");
+        show_seconds.notify["active"].connect (() => {
+            timepicker.show_seconds = show_seconds.active;
+        });
+        show_seconds.set_active (false);        
+
```
danrabbit: Can you remove the double space here an just use one line of whitespace?

demo/Views/DateTimePickerView.vala
```diff
@@ -41,8 +44,29 @@ public class DateTimePickerView : Gtk.Grid {
...
         var time_label = new Gtk.Label ("TimePicker:");
         time_label.halign = Gtk.Align.END;
-
+        
         timepicker = new Granite.Widgets.TimePicker ();
+        
+        show_seconds = new Gtk.Switch ();
+        show_seconds.halign = Gtk.Align.START;
+        show_seconds.set_tooltip_text ("Show Seconds");
+        show_seconds.notify["active"].connect (() => {
```
danrabbit: This is probably better as a [Property Binding](https://valadoc.org/gobject-2.0/GLib.Object.bind_property.html)

demo/Views/DateTimePickerView.vala
```diff
@@ -41,8 +44,29 @@ public class DateTimePickerView : Gtk.Grid {
 
         var time_label = new Gtk.Label ("TimePicker:");
         time_label.halign = Gtk.Align.END;
-
+        
         timepicker = new Granite.Widgets.TimePicker ();
+        
+        show_seconds = new Gtk.Switch ();
+        show_seconds.halign = Gtk.Align.START;
+        show_seconds.set_tooltip_text ("Show Seconds");
```
danrabbit: `tooltip_text` is a valid property so you can do `tooltip_text = "Show Seconds"`

demo/Views/DateTimePickerView.vala
```diff
@@ -41,8 +44,29 @@ public class DateTimePickerView : Gtk.Grid {
...
+        show_seconds.notify["active"].connect (() => {
+            timepicker.show_seconds = show_seconds.active;
+        });
+        show_seconds.set_active (false);
```
danrabbit: `active` is a valid property so you can do `active = false`

demo/Views/DateTimePickerView.vala
```diff
@@ -78,23 +102,31 @@ public class DateTimePickerView : Gtk.Grid {
         set_selected_datetime ();
         datepicker.changed.connect (() => set_selected_datetime ());
         timepicker.changed.connect (() => set_selected_datetime ());
+        timepicker_with_seconds.changed.connect (() => set_selected_datetime_with_seconds ());
+
+        
 
         column_spacing = 12;
-        row_spacing = 6;
+        row_spacing = 9;
```
danrabbit: Can you revert this spacing change please?

FriendOfEntropy: Doh...at one point I thought this was the number of rows in the grid for some reason, and forgot that I had changed it under that wrong impression. Done.

lib/Widgets/TimePicker.vala
FriendOfEntropy: Absolutely. Done.

FriendOfEntropy:
So as you suggested, I'll set aside my current versions of these 2 files and then open a new PR with only refactorings before we revisit this show seconds festival.

FriendOfEntropy:
Closing this to open up a new one based on refactorings only.  After that one is merged, I will follow up with a new one related to show_seconds only which addresses all issues raised here.

PR #222, opened 2018-10-03
Change settings to look at a more relevant value for time format

dahenson:
This particular PR will fix elementary/calendar#268 and fix elementary/calendar#91.

This will only fix those issues for Calendar in elementary, but it will probably break things in other systems that don't have the wingpanel indicator, or a way to change the format of the wingpanel indicator.

Either way, it seems reasonable to expect Granite to respect the system setting for elementary over anything else.

### DIFF
--- a/lib/DateTime.vala
+++ b/lib/DateTime.vala
@@ -110,7 +110,7 @@ namespace Granite.DateTime {
      * @return true if the clock format is 12h based, false otherwise.
      */
     private static bool is_clock_format_12h () {
-        var h24_settings = new Settings ("org.gnome.desktop.interface");
+        var h24_settings = new Settings ("io.elementary.desktop.wingpanel.datetime");
         var format = h24_settings.get_string ("clock-format");
         return (format.contains ("12h"));
     }


hosiet:
I'm wondering if there could be any fallback option for systems on which `io.elementary.desktop.wingpanel.datetime` does not exist. Otherwise the library will crash in such circumstances.

Ref: https://bugs.debian.org/916173

tintou:
This shouldn't have been done, the `clock-format` from the GNOME scheme is the only one that we should use and is enough

dahenson:
This may have been done a little to hastily, but if the GNOME scheme is the only one that should be used, it should be used everywhere, which is not the case for Granite.

JoshStrobl:
This change was made with seemingly zero consideration for any other platforms or application developers which may leverage Granite but still wish to ship their apps on other non-elementary OS platforms, as a key example being bleakgrey/tootle#107. The `org.gnome.desktop.datetime` schema can reliably be found on other platforms where GTK applications and `gsettings-desktop-schemas` (which is what ships the schema) are, such as:

- Platforms / operating systems which ship Budgie Desktop, which uses `gsettings-desktop-schemas` as well as `gnome-session` which depends on it.
- Platforms / operating systems which ship GNOME Shell, which uses `gsettings-desktop-schemas` as well as `gnome-session` (which GNOME Shell and GNOME Session Shell depend on).
- Platforms / operating systems which ship MATE and leverage `mate-session-manager`
- Platforms / operating systems which ship KDE and leverage `kde-gtk-config` to apply GTK settings.

tl;dr - Literally everywhere else. I've personally reverted this change [downstream in Solus](https://dev.getsol.us/R1052:e050745fc223971102b5105d28196221d2d98337).

EbonJaeger:
"which is not the case for calendar I believe (among other things)."

So maybe fix those instead of maybe breaking stuff for everyone else?

JoshStrobl:
PR submitted in #256

dahenson:
> "which is not the case for calendar I believe (among other things)."
> 
> So maybe fix those instead of maybe breaking stuff for everyone else?

You are absolutely right. This was done in error, and should be reverted. It was not my intent to break things for everyone else.

JoshStrobl:
@dahenson I submitted a PR which maintains your change while making sure it works with the prior schema. If you'd like to test it (#256) it'd be appreciated :slightly_smiling_face:

ni-ka:
Still an issue which prevents to use calendar app on arch linyx

PR #225, opened 2018-10-21
Fix erroneous time format translations

aljelly:
Some of the .po files erroneously had 12 hour formats translated as 24 hour formats. This was causing problems like Wingpanel always showing 24 hour time for some locales regardless of the time format setting. I've fixed a few of the files but there may be more I've not found (Github's search function was not particularly cooperative).


tintou:
I've committed the changes using Weblate

tintou:
Thanks for looking at it carefully

PR #229, opened 2018-10-23
 Add utilities for marking up accel tooltips

### LINKED ISSUE #228: ShortcutTooltip
We've started implementing this pattern across apps (i.e. https://github.com/elementary/applications-menu/pull/150), so it could make sense to centralize the implementation here. Probably just need to pass in a description string and a shortcut string.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65148223-shortcuttooltip?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #228 

Adds two new utils:
* `accel_to_string`, which takes something like `"<Super>Right"` and returns `"⌘ + →"`
* `markup_accel_tooltip`, which takes a description and an accel string array and returns the markup pictured below for tooltips

screenshot from 2018-10-23 11 07 26 2x

### DIFF
--- a/lib/Widgets/Utils.vala
+++ b/lib/Widgets/Utils.vala
@@ -83,7 +83,7 @@ public enum Granite.CloseButtonPosition
  *
  * @return a human-readable string like "Ctrl + A" or "⌘ + →"
  */
-public string accel_to_string (string accel) {
+public static string accel_to_string (string accel) {
     uint accel_key;
     Gdk.ModifierType accel_mods;
     Gtk.accelerator_parse (accel, out accel_key, out accel_mods);
@@ -135,7 +135,7 @@ public string accel_to_string (string accel) {
  *
  * @return {@link Pango} markup with the description label on one line and a list of human-readable accels on a new line
  */
-public string markup_accel_tooltip (string description, string[] accels) {
+public static string markup_accel_tooltip (string description, string[] accels) {
     int i = 0;
     foreach (string accel in accels) {
         accels[i] = accel_to_string (accel);
--- a/lib/Widgets/Utils.vala
+++ b/lib/Widgets/Utils.vala
@@ -129,7 +129,12 @@ public static string accel_to_string (string accel) {
 }
 
 /**
- * Takes a description and an array of accels and returns {@link Pango} markup for use in a {@link Gtk.Tooltip}
+ * Takes a description and an array of accels and returns {@link Pango} markup for use in a {@link Gtk.Tooltip}. This method uses {@link Granite.accel_to_string}.
+ *
+ * Example:
+ *
+ * Description
+ * Shortcut 1, Shortcut 2
  *
  * @param a string array of accelerator labels like {"<Control>a", "<Super>Right"}
  *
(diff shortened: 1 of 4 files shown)


lib/Widgets/Utils.vala
```diff
@@ -76,6 +76,78 @@ public enum Granite.CloseButtonPosition
...
+            break;
+        case Gdk.Key.Right:
+            arr += "→";
+            break;
+        default:
+            arr += Gtk.accelerator_get_label (accel_key, 0);
+            break;
+    }
+
+    return string.joinv (" + ", arr);
```
cassidyjames: Should this also be translatable? I wonder if other symbols or ordering would make sense in other languages?

lib/Widgets/Utils.vala
```diff
@@ -76,6 +76,78 @@ public enum Granite.CloseButtonPosition
...
+    int i = 0;
+    foreach (string accel in accels) {
+        accels[i] = accel_to_string (accel);
+        i++;
+    }
+
+    ///TRANSLATORS: This is a delimiter that separates two keyboard shortcut labels like "⌘ + →, Control + A"
+    var accel_label = string.joinv (_(", "), accels);
+
+    return "%s\n<span weight=\"600\" size=\"smaller\" alpha=\"75%\">%s</span>".printf (description, accel_label);
```
cassidyjames: Could use `"""` as @Philip-Scott mentioned to avoid all the inline escaping.

lib/Widgets/Utils.vala
danrabbit: I'm not sure. I guess we'd have to look for RTL and then reverse the direction of the array? Otherwise the delimiter here is evenly spaced on both sides

lib/Widgets/Utils.vala
cassidyjames: Never mind, this gets annoying with the `\n` afaict. Disregard!

lib/Widgets/Utils.vala
cassidyjames: Ah yeah that could be sane to just reverse the array. Sounds good to be but I'd want to get sign-off from a RTL user.

peteruithoven:
(Fixes: https://github.com/elementary/granite/issues/228)

danrabbit:
Something I just thought of, I wonder if the description should be nullable so this can be used comfortably with text buttons as well

danrabbit:
Requesting a review from @donadigo since this is based on, I think, a lot of his code from the applications menu and system settings

lib/Widgets/Utils.vala
```diff
@@ -76,6 +76,88 @@ public enum Granite.CloseButtonPosition
...
+ *
+ * @param a string array of accelerator labels like {"<Control>a", "<Super>Right"}
+ *
+ * @param description a standard tooltip text string
+ *
+ * @return {@link Pango} markup with the description label on one line and a list of human-readable accels on a new line
+ */
+public static string markup_accel_tooltip (string[] accels, string? description = null) {
+    int i = 0;
+    foreach (string accel in accels) {
```
donadigo: I guess you could really just do a for loop here:
```
for (int i = 0; i < accels.length; i++) {
   accels[i] = accel_to_string (accels[i]); 
}
```

lib/Widgets/Utils.vala
```diff
@@ -76,6 +76,88 @@ public enum Granite.CloseButtonPosition
...
+        default:
+            arr += Gtk.accelerator_get_label (accel_key, 0);
+            break;
+    }
+
+    return string.joinv (" + ", arr);
+}
+
+/**
+ * Takes a description and an array of accels and returns {@link Pango} markup for use in a {@link Gtk.Tooltip}
```
donadigo: I think this should be more descriptive of what this function actually does. I would add here the format it returns like:
```
Description
Shortcut 1, Shortcut 2
```

I would also include that it uses `accel_to_string` to convert all the accels to readable strings.

Issue #227, opened 2018-10-23
Include keyboard shortcut in tooltips


peteruithoven:
Just like we've added the keyboard shortcut to the Applications menu button (in the Wingpanel) we could add keyboard shortcuts to widgets like the Tabs widget. For example in the tooltips of:
- [ ] new tab button - Ctrl + T
- [ ] close current tab - Ctrl + W  (can be shown with close Tab tooltip on focused tab)

peteruithoven:
Blocked by #228

PR #236, opened 2018-10-26
Tabs shortcuts in tooltips

peteruithoven:
Fixes: #227 

- New tab button - Ctrl + T
- Only on current tab: close current tab - Ctrl + W


cassidyjames:
Are the shortcuts actually handled by granite, or is that up to the app using the DynamicNotebook? If the keyboard shortcut isn't enforced in Granite I don't think I'd want to show the shortcut here.

peteruithoven:
Ha, you guys are quick, I was working on a seperate branch to fix that empty accel line:
https://github.com/elementary/granite/pull/237

peteruithoven:
The shortcuts I've added are implemented in this Granite widget. Not sure what not enforcing means? Is it easy to disable this?

Shouldn't these shortcuts work anyway, because it's a common convention?

peteruithoven:
I could also add a boolean property to DynamicNotebook (that's turned true by default)?

danrabbit:
@peteruithoven Hm maybe it would be best if we made it so you could pass in your own markup for the buttons, especially since I think these shortcuts are different in Terminal for example

peteruithoven:
But, with the current defaults? (I think showing the current defaults makes sense since we also handle those accels by default)

I've made it possible for the add button, I'll look into the close button.

peteruithoven:
So the close button tooltips where a bit tricky since they change depending on whether they are the current tab. 
I've now chosen for creating 2 properties on the notebook that are passed to tabs:
- `close_button_tooltip`
- `current_close_button_tooltip_accels`

This PR depends on: #237

lib/Widgets/DynamicNotebook.vala
```diff
@@ -668,6 +695,45 @@ namespace Granite.Widgets {
...
 
+        public string add_button_tooltip_markup {
+            owned get { return add_button.get_tooltip_markup (); }
+            set { add_button.tooltip_markup = value; }
+        }
+
+        /**
+         * The text shown in the close button tooltip
+         */
+        private string _close_button_tooltip = _("Close Tab");
```
danrabbit: Is this necessary for having the accel array? I feel like this might be out of scope and suited for another branch/discussion

lib/Widgets/DynamicNotebook.vala
```diff
@@ -668,6 +695,45 @@ namespace Granite.Widgets {
         // Use temporary field to avoid breaking API this can be dropped while preparing for 0.4
         string _add_button_tooltip;
 
+        public string add_button_tooltip_markup {
+            owned get { return add_button.get_tooltip_markup (); }
+            set { add_button.tooltip_markup = value; }
+        }
+
+        /**
```
danrabbit: I think these comments need to be right above the public properties to show in the docs

lib/Widgets/DynamicNotebook.vala
```diff
@@ -668,6 +695,45 @@ namespace Granite.Widgets {
         // Use temporary field to avoid breaking API this can be dropped while preparing for 0.4
         string _add_button_tooltip;
 
+        public string add_button_tooltip_markup {
```
danrabbit: I don't think we want to make the entire markup open-ended do we? I think we just want to set accelerators

lib/Widgets/DynamicNotebook.vala
```diff
@@ -668,6 +695,45 @@ namespace Granite.Widgets {
...
+                    tabs.foreach ((t) => {
+                        t.close_button_tooltip = value;
+                    });
+                }
+
+                _close_button_tooltip = value;
+            }
+        }
+
+        /**
```
danrabbit: I think these comments need to be right above the public properties to show in the docs

lib/Widgets/DynamicNotebook.vala
peteruithoven: Could you clarify the controversy? Do you think the default text should be empty for now?
The text pre-dates this PR:
https://github.com/elementary/granite/pull/236/files?utf8=%E2%9C%93&diff=unified#diff-11a3fe7bae677d4582a9515547517dc9L238

lib/Widgets/DynamicNotebook.vala
peteruithoven: I've now created separated tooltip text and accels properties. Making it consistent with the current close button tooltip. Do you think that's better?

lib/Widgets/DynamicNotebook.vala
danrabbit: Yeah I'm just wondering what's the use case behind allowing developers to change this tooltip text? Shoudn't it always be "Close Tab"?

peteruithoven: Ah fair point. I was thinking it would allow people to use something more specific than "tab", like "Close file", "Close "image" etc. but I'll admit that's a bit of a stretch... 
We could easily remove the text properties (also for the add button).

danrabbit: Yeah, I think we should keep the diff here small and focused on just adding the shortcut accels

peteruithoven:
I removed the properties that allowed customizing the text part of the add tab and close tab tooltips. 
I've also remove the old `add_button_tooltip` property:
https://github.com/elementary/granite/pull/236/files#diff-11a3fe7bae677d4582a9515547517dc9L664
It would override the add button tooltip completely, which was problematic. It also had this note:
```
// Use temporary field to avoid breaking API this can be dropped while preparing for 0.4
```
And we're way past 0.4 by now.

peteruithoven:
Searching on github through the elementary organization (I don't fully trust that search) and code I had locally I found one user of `add_button_tooltip`, the Elementary-IDE:
https://github.com/donadigo/elementary-ide/blob/master/src/Widgets/EditorView.vala#L51

I could properly re-implement `add_button_tooltip` by having it affect only the text part in a separate PR?

jeremypw:
@peteruithoven Has conflicts but there was some interest.  I suggest you fix the conflicts and then re-request a review.

peteruithoven:
Thanks for checking up on all these PR's @jeremypw! 

I'll try working on this again. 

I'd still love a response to whether removing the `add_button_tooltip` is problematic though.

jeremypw:
I do not think removing the `add_button_tooltip ()` is a good idea for this PR - just focus on showing the (customizable) accels. If the developer overrides the default ones in their code its up to them to set the customized accels on the widget, I guess otherwise they will not match. This is not something exposed to the non-technical user to change.

jeremypw:
I would try building with vala 0.48.  It is possible Granite has dependencies not in Hera - a lot of repos now have - and the `meson.build` file needs updating.  You could also set up Elementary 6 in a VM for development - I did that for a while although I am now running it on bare metal without any problems.

peteruithoven:
I'll try to alter `add_button_tooltip` where it only changes the text part of the tooltip. I noticed it's now also used by Sequeler for example:
https://github.com/Alecaddd/sequeler/blob/b2b00628ccccc32413abffe9a3744da0465aaf96/src/Layouts/DataBaseView.vala#L221

jeremypw (dismissed):
I noticed that this does not show the accelerators in the tab context menu - only on the button tooltips.  Is that something you could fix?

jeremypw:
@peteruithoven I think this now needs to be reworked to coordinate with the custom AccelLabels provided by #502.  So converting to draft.

danrabbit:
Closing since dynamic notebook was dropped for Granite 7

PR #237, opened 2018-10-26
Make accels optional

peteruithoven:
In some cases, like example below, it would be nice if even accels would be optional.
```
private void update_close_button_tooltip () {
    string[] accels = {};
    if (_is_current_tab) accels = {"<Ctrl>W"};
    close_button.tooltip_markup = markup_accel_tooltip (accels, _("Close Tab"));
}
```


lib/Widgets/Utils.vala
```diff
@@ -152,27 +152,32 @@ public static string accel_to_string (string accel) {
...
-    var accel_label = string.joinv (_(", "), unique_accels);
+        ///TRANSLATORS: This is a delimiter that separates two keyboard shortcut labels like "⌘ + →, Control + A"
+        var accel_label = string.joinv (_(", "), unique_accels);
 
-    var markup = """<span weight="600" size="smaller" alpha="75%">%s</span>""".printf (accel_label);
+        var markup = """<span weight="600" size="smaller" alpha="75%">%s</span>""".printf (accel_label);
 
-    if (description != null && description != "") {
-        markup = string.join ("\n", description, markup);
+        var accel_markup = """<span weight="600" size="smaller" alpha="75%">%s</span>""".printf (accel_label);
```
danrabbit: Am I missing something or is this a duplicate of the line above. It looks like the variable `markup` is never used in this branch

peteruithoven: Nop, that's definitely duplicate. I don't know how I've missed that. Maybe got lost in a merge conflict.

PR #240, opened 2018-11-03
port to meson

### LINKED ISSUE #177: Migrate to Meson
<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58523255-migrate-to-meson?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github).
</bountysource-plugin>

decathorpe:
CMake Functionality is preserved 1:1, AFAICT. Building the docs works too.

Fixes #177


decathorpe:
So, I've added back the OS detection, which was - I hope - the last thing that got lost with my port. According to http://mesonbuild.com/Reference-tables.html#operating-system-names, OpenBSD isn't supported by meson, so I didn't add that to the list.

I also removed the config.vapi / config.h files, which aren't actually needed anymore.

decathorpe:
Compiling packages against this meson-built granite works in my tests, too. I checked some projects, including code and music.

decathorpe:
While we're at it, "correct" versioning of the .gir and .typelib should be considered as well (it should have been bumped to 5.0 for versions 5.x.y, AFAICT). While it's too late to fix it for the 5.x series, it should be bumped to 6.0 for the 6.x.y series (with the next soname bump). This ensures that applications using an older version of the gir/ typelib bindings continue to work as expected, and different versions of granite can be installed in parallel, because files won't conflict anymore.

Side note: Is there something blocking this PR (other than four additional human eyes)?

lib/meson.build
```diff
@@ -0,0 +1,129 @@
...
+    ],
+    output: granite_typelib,
+    depends: libgranite,
+    install: true,
+    install_dir: join_paths(get_option('libdir'), 'girepository-1.0'),
+)
+
+libgranite_dep = declare_dependency(
+    link_with: libgranite,
+    dependencies: libgranite_deps,
```
tintou: add an `include_directories` here

demo/meson.build
```diff
@@ -0,0 +1,43 @@
...
+    'Views/StorageView.vala',
+    'Views/ToastView.vala',
+    'Views/UtilsView.vala',
+    'Views/WelcomeView.vala',
+
+    dependencies: [libgranite_dep],
+
+    include_directories: [include_directories('../lib')],
+
+    link_with: libgranite,
```
tintou: you don't need `link_with` and `include_directories` here if `libgranite_dep` is well-defined

doc/meson.build
```diff
@@ -0,0 +1,58 @@
+valadoc = find_program('valadoc', required: false)
+gd_scan = find_program('gtkdoc-scan', required: false)
+
+if not valadoc.found()
+    message('valadoc is not installed. building valadocs is not possible.')
```
tintou: I prefer having an option in `meson_options.txt` specifying if this should be enabled than having an hidden and soft fallback, this helps packagers to do things right ;)

lib/meson.build
decathorpe: Like this?

```
libgranite_dep = declare_dependency(
    link_with: libgranite,
    dependencies: libgranite_deps,
    include_directories: [include_directories('.')],
)
```

demo/meson.build
decathorpe: Will do

lib/meson.build
tintou: exactly :)

doc/meson.build
decathorpe: I pushed a commit containing this change, but it breaks running "ninja install" when documentation is enabled. meson/ninja runs the custom targets for the docs twice (on "ninja" and on "ninja install", and I couldn't figure out how to make them run only once.

lib/meson.build
decathorpe: Done :)

tintou:
@decathorpe https://github.com/decathorpe/granite/compare/master...elementary:meson

decathorpe:
> @decathorpe [decathorpe/granite@master...elementary:meson](https://github.com/decathorpe/granite/compare/master...elementary:meson)

Thanks @tintou , merged. Works as advertised. Don't know why I couldn't figure that out, but the docs are a bit ... vague sometimes.

Is there anything else blocking this PR?

decathorpe:
I knew I was forgetting something - I now also added back a meson post-install script for updating the GTK icon cache.

decathorpe:
- Rebased against the latest git master
- checked that ABI doesn't change

(The added symbols are there because the last version I have on fedora that's built with CMake is 5.1.0, and the results are compared to 5.2.2.)

```
================ changes of 'libgranite.so.5.0'===============
  Functions changes summary: 0 Removed, 0 Changed (9 filtered out), 2 Added functions
  Variables changes summary: 0 Removed, 0 Changed, 0 Added variable

  2 Added functions:

    'function gchar* granite_accel_to_string(const gchar*)'    {granite_accel_to_string}
    'function gchar* granite_markup_accel_tooltip(gchar**, int, const gchar*)'    {granite_markup_accel_tooltip}

================ end of changes of 'libgranite.so.5.0'===============
```

So, ignoring the two functions that were added with granite 5.2.0, ABI doesn't change.

Conan-Kudo:
🎉

PR #242, opened 2018-11-04
move "clock-format" GSettings here from the datetime indicator

decathorpe:
This breaks the circular dependency between granite and the datetime indicator, by moving the shared GSettings key to the root of the elementary dependency tree.

This pull request depends on #240 because I have no idea how to do the same thing with CMake.


danrabbit:
I don't think this should live in Granite because it doesn't really have anything to do with this library. This is more a Pantheon thing than a library thing.

I'd be interested in looking what package supplies the keys at `org.gnome.desktop.interface` and doing something similar. My guess is that there's some kind of gnome-desktop-settings package or something that apps who rely on these settings pull in. In this way, apps can also depend on Pantheon without depending on the library Granite.

decathorpe:
I don't see any future where "Pantheon" won't depend on granite. So I don't understand that argument.

granite is the smallest common denominator here.

Introducing a new package for one GSettings schema (even if there might be more in the future) is over-engineering the solution right now.

danrabbit:
Well we already have The five keys that live at `io.elementary.desktop.wingpanel.datetime` but we could also keep here `io.elementary.desktop.wingpanel.bluetooth bluetooth-enabled` which currently forces the switchboard plug to depend on the indicator

decathorpe:
You know, those three facts alone can explain a lot of weird crash reports that I've been getting.

Conan-Kudo:
@decathorpe Could you please split your port to Meson to a separate PR?

decathorpe:
My branch for this pull request has diverged a bit from the meson port branch, I'll close this one, and maybe I'll open a new one in the future.

jtojnar:
> Introducing a new package for one GSettings schema (even if there might be more in the future) is over-engineering the solution right now.

Surely there are many more desktop-wide settings that should go to `elementary-desktop-schemas`. gsettings-desktop-schemas, which GTK depends on, already contains a [lot of keys](https://github.com/GNOME/gsettings-desktop-schemas/tree/master/schemas) so we might take an inspiration there.

PR #252, opened 2018-12-12
DateTime: strip leading spaces from time formats

### LINKED ISSUE #253: Time strings begin with extra spaces
I noticed while working on #348 that time strings, i.e. `7:00 PM` actually contain a leading space, so become:

```
 7:00 PM
```

I'm not sure where it's coming from.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67714888-time-strings-begin-with-extra-spaces?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Gets rid of extra leading spaces in time strings. Fixes #253.


peteruithoven:
@danrabbit please provide some way of testing, like an aspect of elementary OS that this clearly impacts. 

I couldn't really figure out why this padding exists, the docs only state: 
> -: Do not pad a numeric result. This overrides the default padding for the specifier.

Why not remove the padding from the 24h, no seconds formatting? 
https://github.com/elementary/granite/pull/252/files#diff-d44a0b3e3478bd719b41bd449831b474R53

Are you sure this gives the results you want? It for example also removes leading 0's and introduces a padding in the 24h formatting. 
I wrote a little test:
https://gist.github.com/peteruithoven/947d4bafcd2bb1f038b7a8728cca4193
screenshot from 2018-12-13 14 54 05

cassidyjames:
@peteruithoven well one test case is events in Calendar. :smile:

danrabbit:
@peteruithoven oops you're right. I only wanted to change the 12hr. Fixed

PR #255, opened 2018-12-21
Handle 'Return' key

cassidyjames:
It looks like currently passing "Enter" to `markup_accel_tooltip` results in a blank label, and passing "Return" lists that as the key (even though everywhere else we refer to the key as "Enter". This fixes the latter case.

### DIFF
--- a/lib/Widgets/Utils.vala
+++ b/lib/Widgets/Utils.vala
@@ -130,7 +130,6 @@ public static string accel_to_string (string accel) {
             ///TRANSLATORS: This is a non-symbol representation of the "+" key
             arr += _("Plus");
             break;
-        case Gdk.Key.KP_Enter:
         case Gdk.Key.Return:
             arr += _("Enter");
             break;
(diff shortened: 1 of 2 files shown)


cassidyjames:
Example of usage: Write.as uses Ctrl+Enter for a publish action

image

lib/Widgets/Utils.vala
```diff
@@ -130,6 +130,10 @@ public static string accel_to_string (string accel) {
             ///TRANSLATORS: This is a non-symbol representation of the "+" key
             arr += _("Plus");
             break;
+        case Gdk.Key.KP_Enter:
```
danrabbit: I'm not sure we want to lump this in, do we? Right now it shows "Enter (keypad)". It sounds like for the most part most apps don't distinguish, but there are some apps (like [Pro Tools](https://ux.stackexchange.com/questions/54461/should-i-use-the-name-enter-or-return-key)) which do use the keys for different things.

cassidyjames: @danrabbit I was just being consistent with the plus/minus lines above. But I am fine to remove it in this PR and correct those later (if we want).

PR #257, opened 2019-01-23
Add function to check if history is enabled within the privacy settings

Philip-Scott:
A lot of our apps are not testing if privacy mode is enabled to prevent things like recent files from showing up. But adding a settings dependency to `org.gnome.desktop.privacy` on every app that wants to check if privacy mode is enabled could be prevented by a helper function in Granite. This PR adds said function, and makes is to that if the setting is not found (Different OS or maybe GNOME changed the schema) it will just return `true` and the apps can act as normal
 
## Use cases: 

Terminal: PR to stop Tabs from being saved https://github.com/elementary/terminal/pull/349
Spice-Up: Stop saving the used files while it's enabled: https://github.com/Philip-Scott/Spice-up/pull/260
Videos also has the direct dependency at tihs line on [Window.vala](https://github.com/elementary/videos/blob/59dc6d61b0ad4a7e91cdda076169cd328f58a0c8/src/Window.vala#L407)

### DIFF
--- a/lib/Services/System.vala
+++ b/lib/Services/System.vala
@@ -168,11 +168,12 @@ namespace Granite.Services {
         private static GLib.Settings? privacy_settings = null;
 
         /**
-         * Returns whether privacy mode is enabled in the system settings or not.
+         * Returns whether history is enabled within the Security and Privacy system settings or not. A value of true
+         * means that you should store information such as the last opened file or a history within the app.
          *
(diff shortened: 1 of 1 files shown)


Philip-Scott:
You are totally right! It made more sense last night :sweat_smile: 

Updated the function name, as well as added a description of what the developer should do if it returns true

PR #213, opened 2018-09-07
Refactorings to prep for a later show_seconds PR

FriendOfEntropy:
Related to #208 

Per discussion in a previously canceled [show_seconds PR](https://github.com/elementary/granite/pull/211), this is a set of small refactorings to limit the amount of simultaneous changes for review when a new show_seconds PR is created later.

- Parse_time will become especially complex and busy when second support is added, so making it a bit more succinct here.  
- Going ahead and adding a with_format creation version of TimePicker to the demo view to help with regression testing, demonstrate the usage in code.  When show_seconds is added later, there will be a with_seconds equivalent of this being demonstrated

PR #260, opened 2019-01-29
TimePicker: Use more compact boolean comparison syntax

danrabbit:
Just changes to Timepicker.vala from #213 

Also try to reduce redundant comparisons in `parse_time`

### DIFF
--- a/lib/Widgets/TimePicker.vala
+++ b/lib/Widgets/TimePicker.vala
@@ -307,7 +307,7 @@ namespace Granite.Widgets {
                         }
                     }
 
-                    if (c.to_string ().contains ("m") && is_suffix == true) {
+                    if (c.to_string ().contains ("m") && is_suffix) {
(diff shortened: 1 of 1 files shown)


lib/Widgets/TimePicker.vala
```diff
@@ -289,18 +289,22 @@ namespace Granite.Widgets {
...
-                    if ((c.to_string ().contains ("a") || c.to_string ().contains ("p")) && is_suffix == false) {
-                        is_suffix = true;
-                        current = "%s%c".printf (current, c);
+                        if (c.to_string ().contains ("a") || c.to_string ().contains ("p")) {
+                            is_suffix = true;
+                            current = "%s%c".printf (current, c);
+                        }
                     }
 
                     if (c.to_string ().contains ("m") && is_suffix == true) {
```
ricotz: At least one more

PR #261, opened 2019-01-29
TimePicker: Make parse_time non-void

danrabbit:
Make `parse_time` return a value so that we can add tests for it


tintou:
You can already add tests by using the `time` variable inside of the class here you're making it a little bit complicated

PR #262, opened 2019-01-31
[WIP] Fixes #224 : Implements a reusable way to remember window state

### LINKED ISSUE #224: Reusable logic to remember window position and size
A lot of applications are writing custom logic to store and retrieve the last position and sometimes size of the window. 
It would be great if there was something build into Granite that could handle this.
Maybe `load_state()` and `save_state()` functions that apps could optional call.

naaando:


donadigo:
This probably should go in `lib/Services`.

A couple of additional TODO comments:
* This should probably be named differently as this is an interface, we usually start interface names with capital `I`.
* All `get_` methods can be static / can be just public constants.
* Of course, lacks documentation and examples.

naaando:
Hi @donadigo, about the second point, I don't think it's a good idea to use a static method for this because the idea behind those getters are to be overridden if the gsettings name differs or you have more than one window saving states, about using constants would be better but they can be overridden?

donadigo:
@naaando: oh I didn't think that those could be used to override key names, in this case my comment doesn't apply anymore.

naaando:
Unfortunatelly I misunderstood the gsettings documentation, the delay () method switch gsettings to a unreversible delay-apply state, so any other settings would have to call apply () to commit the change. This happened because using a interface I had to figure out a way to throttle without a state. I fixed it on my project by using a class/object instead.

Issue #265, opened 2019-02-07
Fails to build on Ubuntu 18.04.1 LTS


NotAlexNoyle:
I installed all of the dependencies and get this error from meson:

```
└─ $ ▶ meson build --prefix=/usr
The Meson build system
Version: 0.45.1
Source dir: /home/xubuntu/granite
Build dir: /home/xubuntu/granite/build
Build type: native build
Project name: granite
Native C compiler: cc (gcc 7.3.0 "cc (Ubuntu 7.3.0-27ubuntu1~18.04) 7.3.0")
Native Vala compiler: valac (valac 0.40.8)
Build machine cpu family: x86_64
Build machine cpu: x86_64
Found pkg-config: /usr/bin/pkg-config (0.29.1)
Native dependency gee-0.8 found: YES 0.20.1
Native dependency gio-2.0 found: YES 2.56.3
Native dependency gio-unix-2.0 found: YES 2.56.3
Native dependency glib-2.0 found: YES 2.56.3
Native dependency gobject-2.0 found: YES 2.56.3
Native dependency gtk+-3.0 found: YES 3.22.30
Library m found: YES
Library posix found: YES
Program g-ir-compiler found: YES (/usr/bin/g-ir-compiler)

lib/meson.build:122:0: ERROR: Pkgconfig_gen takes no positional arguments.

```

tintou:
The project requires meson >=0.48.2, it is now specified in master.

PR #266, opened 2019-02-07
meson: require meson >= 0.48.2

codygarver:


codygarver:
Closes #265

codygarver:
I tried digging through the Meson release notes to figure out exactly what version fixed #265 but no luck, so I just used to earliest version I know that is confirmed to work.

PR #269, opened 2019-02-28
System: deprecate utilies provided by GLib

danrabbit:
All these utilities should be deprecated for a couple of reasons:

1. They're provided by GLib.AppInfo in few lines already so there's no real convenience benefit

2. These utilities don't return an Error and apps should really handle their errors in a more graceful way than just terminal output, with at least some kind of feedback in the UI.

Contents of the utilities have been replaced to not internally use deprecate methods so we don't get deprecation warnings when building Granite.

### DIFF
--- a/lib/Services/System.vala
+++ b/lib/Services/System.vala
@@ -18,57 +18,74 @@
  */
 
 namespace Granite.Services {
-
     /**
      * Utility class for frequently-used system-related functions, such as opening files, launching
      * applications, or executing terminal commands.
      */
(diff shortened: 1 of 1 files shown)


jeremypw:
An alternative approach would be for these convenience methods to throw an Error.

danrabbit:
> An alternative approach would be for these convenience methods to throw an Error.

For at least of couple of these, it looks like the method wouldn't have any real utility of using GLib directly expect for they handle their own errors in the command line output

PR #277, opened 2019-03-12
Expose OS-wide dark style preference

### LINKED ISSUE #276: Add way to check if the OS prefers a dark style
To enable system apps/components to more easily go dark when the OS wants to go dark. See: elementary/switchboard-plug-pantheon-shell#171

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/71265927-add-way-to-check-if-the-os-prefers-a-dark-style?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Will fix #276. ~~Right now this is just a method that returns a bool, but I imagine it'd be more useful to be able to bind to this.~~ Updated to be a property. Not sure where it would belong though. Feedback welcome.

### DIFF
--- a/demo/Views/UtilsView.vala
+++ b/demo/Views/UtilsView.vala
@@ -32,7 +32,7 @@ public class UtilsView : Gtk.Grid {
         prefer_dark_label.halign = Gtk.Align.END;
 
         string prefer_dark_string;
-        if (Granite.prefer_dark {
+        if (Granite.Services.System.prefer_dark == true) {
             prefer_dark_string = "Dark";
         } else {
             prefer_dark_string = "Default";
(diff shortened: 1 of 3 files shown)


worldofpeace:
> If the GSetting schema that contains the setting does not
>   * exist, this will not connect and cb will never be called.

But because the GSettings Schema is provided by `switchboard-plug-pantheon-shell` it creates a circular cycle between granite.

It shouldn't cause a failure because of the precaution, but it would be a good practice to avoid that entirely.

cassidyjames:
@worldofpeace yeah I suppose in this case Granite could create the GSettings Schema itself?

donadigo:
@cassidyjames in my opinion ideally for such a setting, no non-core / exclusive package should manage how this setting is created. The setting should be created and built-in on an OS level (e.g. at install). This ensures that we don't add any unnecessary dependencies.

worldofpeace:
> @worldofpeace yeah I suppose in this case Granite could create the GSettings Schema itself?

There's a similar problem with https://github.com/elementary/granite/pull/242 and @danrabbit said that didn't belong in granite either. 
Seemed we mostly agreed on there being an `elementary-desktop-schemas`.

I think we should go with that.

cassidyjames:
Closing in favor of #319

PR #284, opened 2019-05-06
Contrasting foreground color

### LINKED ISSUE #244: Contrasting color util
There are times when an app might want to get a contrasting color, i.e. white or black, in order to overlay onto a dynamically-colored background. For example, I use [a few utilities for this in Palette](https://github.com/cassidyjames/palette/blob/395103b1cd7735782a332c88f37857b8635137c8/src/Widgets/ColorButton.vala#L106), as demonstrated with the foreground text color here: screenshot

This could also be used for saner defaults for foreground colors in branded headerbars, preventing unreadable dark gray text on darker brand colors.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65730169-contrasting-color-util?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Fixes #244. Input welcome.

image

### DIFF
--- a/lib/Widgets/Utils.vala
+++ b/lib/Widgets/Utils.vala
@@ -204,7 +204,7 @@ public static string markup_accel_tooltip (string[]? accels, string? description
     return string.joinv ("\n", parts);
 }
 
-private double contrast_ratio (Gdk.RGBA bg_color, Gdk.RGBA fg_color) {
+private static double contrast_ratio (Gdk.RGBA bg_color, Gdk.RGBA fg_color) {
     var bg_luminance = get_luminance (bg_color);
     var fg_luminance = get_luminance (fg_color);
 
@@ -215,15 +215,15 @@ private double contrast_ratio (Gdk.RGBA bg_color, Gdk.RGBA fg_color) {
     return (fg_luminance + 0.05) / (bg_luminance + 0.05);
 }
 
-private double get_luminance (Gdk.RGBA color) {
+private static double get_luminance (Gdk.RGBA color) {
     var red = sanitize_color (color.red) * 0.2126;
     var green = sanitize_color (color.green) * 0.7152;
     var blue = sanitize_color (color.blue) * 0.0722;
 
-    return (red + green + blue);
+    return red + green + blue;
 }
 
-private double sanitize_color (double color) {
+private static double sanitize_color (double color) {
     if (color <= 0.03928) {
         return color / 12.92;
     }
@@ -232,18 +232,15 @@ private double sanitize_color (double color) {
 }
 
 /**
- * Takes a {@link Gdk.RGBA} background color and returns a suitably-contrasting foreground color, i.e. for determining text color on a colored background.
+ * Takes a {@link Gdk.RGBA} background color and returns a suitably-contrasting foreground color, i.e. for determining text color on a colored background. There is a slight bias toward returning white, as white generally looks better on a wider range of colored backgrounds than black.
  *
- * @param bg_color a {@link Gdk.RGBA} background color
+ * @param bg_color any {@link Gdk.RGBA} background color
  *
- * @return a contrasting {@link Gdk.RGBA} foreground color
+ * @return a contrasting {@link Gdk.RGBA} foreground color, i.e. white ({ 1.0, 1.0, 1.0, 1.0}) or black ({ 0.0, 0.0, 0.0, 0.0}).
  */
 public static Gdk.RGBA contrasting_foreground_color (Gdk.RGBA bg_color) {
-    var gdk_white = Gdk.RGBA ();
-    gdk_white.parse ("#fff");
-
-    var gdk_black = Gdk.RGBA ();
-    gdk_black.parse ("#000");
+    Gdk.RGBA gdk_white = { 1.0, 1.0, 1.0, 1.0 };
+    Gdk.RGBA gdk_black = { 0.0, 0.0, 0.0, 1.0 };
 
     var contrast_with_white = contrast_ratio (
         bg_color,
--- a/lib/Widgets/Utils.vala
+++ b/lib/Widgets/Utils.vala
@@ -236,7 +236,7 @@ private static double sanitize_color (double color) {
  *
  * @param bg_color any {@link Gdk.RGBA} background color
  *
- * @return a contrasting {@link Gdk.RGBA} foreground color, i.e. white ({ 1.0, 1.0, 1.0, 1.0}) or black ({ 0.0, 0.0, 0.0, 0.0}).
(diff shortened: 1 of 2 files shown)


lib/Widgets/Utils.vala
```diff
@@ -204,6 +204,68 @@ public static string markup_accel_tooltip (string[]? accels, string? description
...
+private double get_luminance (Gdk.RGBA color) {
+    var red = sanitize_color (color.red) * 0.2126;
+    var green = sanitize_color (color.green) * 0.7152;
+    var blue = sanitize_color (color.blue) * 0.0722;
+
+    return (red + green + blue);
+}
+
+private double sanitize_color (double color) {
+    if (color <= 0.03928) {
```
cassidyjames: Should probably comment where this magic number is coming from and/or use a const?

lib/Widgets/Utils.vala
```diff
@@ -204,6 +204,68 @@ public static string markup_accel_tooltip (string[]? accels, string? description
...
+        return color / 12.92;
+    }
+
+    return Math.pow ((color + 0.055) / 1.055, 2.4);
```
cassidyjames: Same here with magic numbers.

lib/Widgets/Utils.vala
```diff
@@ -204,6 +204,68 @@ public static string markup_accel_tooltip (string[]? accels, string? description
...
+/**
+ * Takes a {@link Gdk.RGBA} background color and returns a suitably-contrasting foreground color, i.e. for determining text color on a colored background.
+ *
+ * @param bg_color a {@link Gdk.RGBA} background color
+ *
+ * @return a contrasting {@link Gdk.RGBA} foreground color
+ */
+public static Gdk.RGBA contrasting_foreground_color (Gdk.RGBA bg_color) {
+    var gdk_white = Gdk.RGBA ();
+    gdk_white.parse ("#fff");
```
donadigo: Instead of doing additional work here just pre-compile the struct:
`Gdk.RGBA gdk_white = { 1.0, 1.0, 1.0, 1.0 };`

lib/Widgets/Utils.vala
```diff
@@ -204,6 +204,68 @@ public static string markup_accel_tooltip (string[]? accels, string? description
...
+
+    var gdk_black = Gdk.RGBA ();
+    gdk_black.parse ("#000");
```
donadigo: Same here:
`Gdk.RGBA gdk_black = { 0.0, 0.0, 0.0, 1.0 };`

lib/Widgets/Utils.vala
```diff
@@ -204,6 +204,68 @@ public static string markup_accel_tooltip (string[]? accels, string? description
     return string.joinv ("\n", parts);
 }
 
+private double contrast_ratio (Gdk.RGBA bg_color, Gdk.RGBA fg_color) {
```
donadigo: Should be static.
All other private functions should be static too.

lib/Widgets/Utils.vala
```diff
@@ -204,6 +204,68 @@ public static string markup_accel_tooltip (string[]? accels, string? description
...
+
+    return (fg_luminance + 0.05) / (bg_luminance + 0.05);
+}
+
+private double get_luminance (Gdk.RGBA color) {
+    var red = sanitize_color (color.red) * 0.2126;
+    var green = sanitize_color (color.green) * 0.7152;
+    var blue = sanitize_color (color.blue) * 0.0722;
+
+    return (red + green + blue);
```
donadigo: Don't need parentheses here.

lib/Widgets/Utils.vala
```diff
@@ -204,6 +204,68 @@ public static string markup_accel_tooltip (string[]? accels, string? description
...
+        bg_color,
+        gdk_black
+    );
+
+    // Default to white
+    var fg_color = gdk_white;
+
+    // NOTE: We cheat and add 3 to contrast when checking against black,
+    // because white generally looks better on a colored background
+    if ( contrast_with_black > (contrast_with_white + 3) ) {
```
donadigo: Probably worth mentioning in the documentation of this function:

> On a colored background e.g a color value with non-equal RGBA fields, the function will prefer to return the white color, however a black color can still be returned - this is achieved with having a little bias towards the white color.
The reason for this is that intuitively the white color generally looks better on a colored background.

lib/Widgets/Utils.vala
```diff
@@ -204,6 +204,68 @@ public static string markup_accel_tooltip (string[]? accels, string? description
...
+
+    return Math.pow ((color + 0.055) / 1.055, 2.4);
+}
+
+/**
+ * Takes a {@link Gdk.RGBA} background color and returns a suitably-contrasting foreground color, i.e. for determining text color on a colored background.
+ *
+ * @param bg_color a {@link Gdk.RGBA} background color
+ *
+ * @return a contrasting {@link Gdk.RGBA} foreground color
```
donadigo: This should probably also mention that the only colors that can be returned is either completely black or completely white and possible values e.g:

> The return value can be either a white color (`{ 1.0, 1.0, 1.0, 1.0`}, or a black color (`{ 0.0, 0.0, 0.0, 0.0}`) depending on the contrast of the passed {@link Gdk.RGBA} color.

cassidyjames: @donadigo that's true now, but I was wondering if we'd want to do anything more clever with shades of the background color, or not. So I was leaving it open.

lib/Widgets/Utils.vala
cassidyjames: @danrabbit this came from https://github.com/danrabbit/harvey/blob/57dcb07e55b7e33eb0947de720b95117228bab35/src/MainWindow.vala#L268 originally. Do you know how to explain these numbers? :sweat_smile:

donadigo: Not Daniel, but I can say that they're probably not "magic" as in what works or doesn't but are part of an algorithm that dictates those as a required part. You probably don't need to have constants for all these at all.
For example `get_luminance` uses a formula directly from a wiki page: https://en.wikipedia.org/wiki/Relative_luminance which in turn uses `sanitize_color` from https://en.wikipedia.org/wiki/SRGB#The_reverse_transformation.

cassidyjames: @donadigo my problem is that they're not super obvious to passers-by, so I wonder if documenting them better here (even just with a comment) would be worthwhile.

danrabbit: Yeah these numbers come from:
https://www.w3.org/TR/WCAG20/#contrast-ratiodef
https://www.w3.org/TR/WCAG20/#relativeluminancedef

I think it would be worth commenting with links to either WCAG or Wikipedia

PR #285, opened 2019-05-06
Set headerbar foreground color with set_color_primary

cassidyjames:
Builds off of https://github.com/elementary/granite/pull/284. I thought it made sense here since I frequently see brand colors without properly-contrasting foreground colors.

### DIFF
--- a/demo/GraniteDemo.vala
+++ b/demo/GraniteDemo.vala
@@ -33,9 +33,13 @@ public class Granite.Demo : Gtk.Application {
     public override void activate () {
         var window = new Gtk.Window ();
 
+        var headerbar = new Gtk.HeaderBar ();
+        headerbar.get_style_context ().add_class ("default-decoration");
+        headerbar.show_close_button = true;
+
         var alert_view = new AlertViewView ();
         var avatar_view = new AvatarView ();
-        var css_view = new CSSView (window);
+        var css_view = new CSSView (headerbar);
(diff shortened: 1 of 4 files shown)


danrabbit:
I'm kind of thinking this should probably be a different method and it should take the titlebar as an argument instead of the window for a couple reasons:

1. It might be better if we request the dark or light stylesheet for the titlebar widget instead of trying to mimic those styles and maintain them in two places. There's a big difference with the text shadow with, for example, the lightest purple in the color chooser. I think this would make other box shadows look nicer and make more sense as well.

2. For apps with flat titlebars, I think this might introduce unwanted consequences and changing the API in this way is probably unexpected for people already using it. I think it would be better to keep the results very predictable and contained to just setting the titlebar color so developers get only what they expect.

cassidyjames:
The latest commit is where I'm currently stuck. I am running into the same [problem I saw in Ephemeral](https://github.com/cassidyjames/ephemeral/pull/26), where I can't set the variant and also set a custom style, because the two css_providers override one another.

jeremypw:
One work around is to (re)apply the brand theming *after* setting the appropriate variant in `set_brand_color`.  You can do some refactoring to avoid applying the theme css_provider twice.

jeremypw:
@cassidyjames Converting to draft as has unaddressed issues and no recent activity.  Please feel free to close or reopen for review after fixing conflicts as desired.

PR #288, opened 2019-06-02
Provide Badge and Progress support for an Application

### LINKED ISSUE #287: Provide the missing bits from libunity
We are relying on libunity in a few places and the library is over-complicated and using outdated dependencies, it might be the right time to provide our alternative to it:

Alternative API:
```vala
namespace Granite.Application {
    public static bool set_badge (int64 count) throws GLib.Error;
    public static bool set_badge_visible (bool visible) throws GLib.Error;
    public static bool set_progress (double progress) throws GLib.Error;
    public static bool set_progress_visible (bool visible) throws GLib.Error;
    public static bool set_dynamic_menu (GLib.Menu menu) throws GLib.Error;
}
```
This API would internally use the `GLib.Application.get_default ().application_id`

The `dynamic menu` one might be questionable : Only Switchboard is using it and it's not persistent.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/74876951-provide-the-missing-bits-from-libunity?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github).
</bountysource-plugin>

tintou:
It is usign the same DBus API as libunity and is here to replace the tiny part that we are still using.
Fixes #287


lib/Services/Application.vala
```diff
@@ -0,0 +1,124 @@
...
+                properties["progress"] = new GLib.Variant.double (0.0);
+                properties["progress-visible"] = new GLib.Variant.boolean (false);
+            }
+
+            private string app_uri;
+            private GLib.HashTable<string,GLib.Variant> properties;
+
+            internal void set_app_property (string property, GLib.Variant var) {
+                var updated_properties = new GLib.HashTable<string,GLib.Variant> (str_hash, str_equal);
+                updated_properties[property] = var;
```
jeremypw: updated_properties seems to have no function?

lib/Services/Application.vala
```diff
@@ -0,0 +1,124 @@
...
+ *  Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ *  Boston, MA 02110-1301 USA.
+ */
+
+[CCode (gir_namespace = "GraniteServices", gir_version = "1.0")]
+namespace Granite.Services {
+    /**
+     * Utilities for Applications
+     */
+    namespace Application {
```
donadigo: It seems like we moved from namespacing anything altogether in Granite. All recently merged classes are located in `Granite` rather than in sub namespaces. Do we still want that here?

lib/Services/Application.vala
tintou: Oops, updated

demo/Views/ApplicationView.vala
```diff
@@ -0,0 +1,81 @@
...
+        attach (progress_visible_switch, 1, 0, 1, 1);
+        attach (progress_scale, 0, 1, 2, 1);
+        attach (badge_visible_label, 0, 2, 1, 1);
+        attach (badge_visible_switch, 1, 2, 1, 1);
+        attach (badge_spin, 0, 3, 2, 1);
+
+        progress_scale.value_changed.connect (() => {
+            Granite.Services.Application.set_progress.begin (progress_scale.get_value (), (obj, res) => {
+                try {
+                    Granite.Services.Application.set_progress.end (res);
```
donadigo: I don't think we need to have the async callback here at all. I believe Vala will automatically complete the callback with just calling `begin`.

demo/Views/ApplicationView.vala
```diff
@@ -0,0 +1,81 @@
...
+                } catch (GLib.Error e) {
+                    critical (e.message);
+                }
+            });
+        });
+
+        progress_visible_switch.notify["active"].connect (() => {
+            Granite.Services.Application.set_progress_visible.begin (progress_visible_switch.active, (obj, res) => {
+                try {
+                    Granite.Services.Application.set_progress_visible.end (res);
```
donadigo: Same here.

demo/Views/ApplicationView.vala
```diff
@@ -0,0 +1,81 @@
...
+                } catch (GLib.Error e) {
+                    critical (e.message);
+                }
+            });
+        });
+
+        badge_spin.value_changed.connect (() => {
+            Granite.Services.Application.set_badge.begin ((int64)badge_spin.value, (obj, res) => {
+                try {
+                    Granite.Services.Application.set_badge.end (res);
```
donadigo: Same here.

demo/Views/ApplicationView.vala
```diff
@@ -0,0 +1,81 @@
...
+                } catch (GLib.Error e) {
+                    critical (e.message);
+                }
+            });
+        });
+
+        badge_visible_switch.notify["active"].connect (() => {
+            Granite.Services.Application.set_badge_visible.begin (badge_visible_switch.active, (obj, res) => {
+                try {
+                    Granite.Services.Application.set_badge_visible.end (res);
```
donadigo: Same here.

demo/Views/ApplicationView.vala
tintou: we want the callback because I'm using the error that might be throwed

lib/Services/Application.vala
tintou: I think that we are moving from namespacing regarding the Gtk widgets but here it's an utility and Granite.Application is already a class unfortunately

donadigo: Ah okay, then it makes sense, but since the class name is `UnityLauncherEntry`, can't just put that into `Services`?

demo/Views/ApplicationView.vala
donadigo: Okay then.

tintou:
> Have come diff comments.
> Also one question: do you know what package / service provides `/com/canonical/unity/`? Or is there even a need for it to be provided at all by unity?

We are registering this path here by the application so there is no need for any extra package, the dock is responsible for watching to the addition/removal of interfaces

lib/Services/Application.vala
tintou: UnityLauncherEntry is the private DBus interface, the static functions aren't in a class :slightly_smiling_face:

PR #295, opened 2019-07-04
Fix #294: Rapidly closing tabs with three-finger/middle click button re-opens closed tabs

### LINKED ISSUE #294: Dynamic notebook: Rapidly closing tabs with three-finger/middle click button re-opens closed tab
@jlnr already reported this issue in https://github.com/elementary/code/issues/485, but as it is a problem of Granite's DynamicNotebook I think It should also be mentioned here!

@jlnr also reported the undesired opening of new tabs when double clicking in #221. Even though both issues are caused by wrong event handling of the `DynamicNotebook` class, I think these are two different bugs. So I have opened this separate issue.

**The Problem**:
When rapidly closing tabs with a three-finger/middle click button, instead of closing these tabs the last closed tabs get restored.

RAPID_CLICKING

**Cause**
I think the problem is that a tab can only be closed if the `close_button` is visible. But the Gtk.Revealer `close_button_revealer`, which is the parent of the `close_button`, has a transition duration of 250ms.

After closing a tab, the cursor enters a new tab and toggles the visibility of the `close_button` of this tab. As the reveal animation takes 250ms this tab cannot be closed by a middle click during this time!

Furthermore as `e.button == 2 && close_button_is_visible ()` evaluates to `false` the middle click event is not consumed by the `Tab` and instead it gets handled by the `DynamicNotebook`, which leads to unwanted restoring of the last closed tab.

---
**The Tab:**
```vala
public Tab (string? label = null, GLib.Icon? icon = null, Gtk.Widget? page = null) {
    ...
    this.enter_notify_event.connect ((e) => {
        cursor_over_tab = true;
        update_close_button_visibility (); // <-- triggers 250ms animation
        return false;
    });
    ...
    this.button_press_event.connect ((e) => {
        if (e.button == 2 && close_button_is_visible ()) { // <-- The close button of the next tab is not visible when clicking fast enough
            e.state &= MODIFIER_MASK;

            if  (e.state == 0) {
                this.closed ();
            } else if (e.state == Gdk.ModifierType.SHIFT_MASK) {
                this.close_others ();
            }
        } else if {
          ...
        } else {
            return false; // <-- button_press_event is handeled by Dynamic notebook, which restores last closed tab!
        }
return true;
```
**DynamicNotebook:**
```vala
button_press_event.connect ((e) => {
    if (e.type == Gdk.EventType.2BUTTON_PRESS && e.button == 1) {
        new_tab_requested ();
    } else if (e.button == 2 && allow_restoring) {
        restore_last_tab (); // <-- Middle click opens new tab
        return true;
    } else if (e.button == 3) {
        menu.popup_at_pointer (e);
    }

    return false;
});
```

**Check transition duration:**
```vala
debug("%u", close_button_revealer.get_transition_duration());`
```

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/76472393-dynamic-notebook-rapidly-closing-tabs-with-three-finger-middle-click-button-re-opens-closed-tab?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github).
</bountysource-plugin>

andreasfelix:
Fix for #294 and https://github.com/elementary/code/issues/485: Dynamic notebook: Rapidly closing tabs with three-finger/middle
click button re-opens closed tabs

Changing `child_revealed` to `reveal_child` to make tab close-able by middle/three-finger click during
reveal animation. This prevents the creation of new tabs when trying to rapidly close multiple tabs.

**Before**

RAPID_CLICKING

**After**

RAPID_CLICKING_FIX

### DIFF
--- a/lib/Widgets/DynamicNotebook.vala
+++ b/lib/Widgets/DynamicNotebook.vala
@@ -301,9 +301,8 @@ namespace Granite.Widgets {
             });
 
             this.button_press_event.connect ((e) => {
-                if (e.button == 2 && close_button_is_visible ()) {
+                if (e.button == 2) {
                     e.state &= MODIFIER_MASK;
-
                     if  (e.state == 0) {
                         this.closed ();
                     } else if (e.state == Gdk.ModifierType.SHIFT_MASK) {
@@ -386,10 +385,6 @@ namespace Granite.Widgets {
 
             close_button_revealer.reveal_child = _closable && !_pinned
                 && (cursor_over_tab || cursor_over_close_button || _is_current_tab);
-        }
-
-        private bool close_button_is_visible () {
-            return close_button_revealer.visible && close_button_revealer.reveal_child;
         }
     }
 
--- a/lib/Widgets/DynamicNotebook.vala
+++ b/lib/Widgets/DynamicNotebook.vala
@@ -389,7 +389,7 @@ namespace Granite.Widgets {
         }
 
         private bool close_button_is_visible () {
-            return close_button_revealer.visible && close_button_revealer.child_revealed;
+            return close_button_revealer.visible && close_button_revealer.reveal_child;
         }
     }
 


jlnr:
@andreasfelix Definitely better than before! It still happens when I double-wheelclick on the border between two tabs, though. Do you think this is can be fixed within the same PR?

andreasfelix:
Unfortunately I think this is a different Issue. I have highlighted the size of the GranitsWidgetsTab and logged the enter and leave events:

Border_event

As one can see, the size of the GranitsWidgetsTab is slightly smaller than its visual size. This can also be seen at the event log: The signal get emitted when entering/leaving the blue highlighted area.

andreasfelix:
I have a other question: What is the rationale behind the `close_button_is_visible` method in the first place? Why shouldn't a tab be close-able by middle/three-finger when the `close_button` is not visible? Is this in any way needed for pinned tabs? I ask because of the line `close_button_revealer.visible = !_pinned`.

Otherwise it would be simpler to just remove the `private bool close_button_is_visible ()` from the if statement:

```vala
if (e.button == 2 && close_button_is_visible ()) { // <-- remove close_button_is_visible ()
    e.state &= MODIFIER_MASK;

    if  (e.state == 0) {
        this.closed ();
    } else if (e.state == Gdk.ModifierType.SHIFT_MASK) {
        this.close_others ();
    }
} 
```

jlnr:
> What is the rationale behind the close_button_is_visible method in the first place?

I have no clue. git blame leads to this commit: https://github.com/elementary/granite/commit/13f290aa53814175b55bf8119a3bf24fe6a33c00

The referenced launchpad issue is about hiding the (x) buttons until you hover the tab: https://bugs.launchpad.net/granite/+bug/1205236

I would guess that it was added out of unspecific defensiveness. Have you tried removing it? Does anything break? Less bugs *and* less code sounds good to me :)

andreasfelix:
I have completely removed the `close_button_is_visible` method as it always returns true in the if statement and isn't used anywhere else.

Issue #18, opened 2017-04-21
[DynamicNotebook] Close Tabs to the Right


elementaryBot:
Would the "Close Tabs to the Right" feature of Google Chrome, Midori and others be interesting to have in DynamicNotebook?

Launchpad Details: [#LP1269617](https://bugs.launchpad.net/bugs/1269617) Niclas Lockner - 2014-01-15 22:15:52 +0000

PR #296, opened 2019-07-07
Close Tabs to the Right

andreasfelix:
Fixes: #18 

Close Tabs to the right

As I am relatively new to vala/gtk I could use some feedback:

Is there a better way than 

https://github.com/andreasfelix/granite/blob/2d9637c2e798594aa231b5fe3b726146d5dc67f2/lib/Widgets/DynamicNotebook.vala#L320-L322

to get the positon of a tab?

Is 

https://github.com/andreasfelix/granite/blob/2d9637c2e798594aa231b5fe3b726146d5dc67f2/lib/Widgets/DynamicNotebook.vala#L1205-L1214

an appropiate way to close all tabs to the right?

### DIFF
--- a/lib/Widgets/DynamicNotebook.vala
+++ b/lib/Widgets/DynamicNotebook.vala
@@ -1194,22 +1194,22 @@ namespace Granite.Widgets {
             }
         }
 
-        private void on_close_others (Tab clickedTab) {
+        private void on_close_others (Tab clicked_tab) {
             tabs.copy ().foreach ((tab) => {
-                if (tab != clickedTab) {
+                if (tab != clicked_tab) {
                     tab.closed ();
                 }
             });
         }
 
-        private void on_close_others_right (Tab clickedTab) {
-            var toTheRight = false; 
+        private void on_close_others_right (Tab clicked_tab) {
+            var is_to_the_right = false; 
 
             tabs.copy ().foreach ((tab) => {
-                if (toTheRight)
+                if (is_to_the_right)
                     tab.closed ();
-                if (tab == clickedTab) 
-                    toTheRight = true;
+                if (tab == clicked_tab) 
+                is_to_the_right = true;
             });
         }
 
--- a/lib/Widgets/DynamicNotebook.vala
+++ b/lib/Widgets/DynamicNotebook.vala
@@ -1206,10 +1206,12 @@ namespace Granite.Widgets {
             var is_to_the_right = false; 
 
             tabs.copy ().foreach ((tab) => {
-                if (is_to_the_right)
+                if (is_to_the_right) {
                     tab.closed ();
-                if (tab == clicked_tab) 
-                is_to_the_right = true;
+                }
+                if (tab == clicked_tab) {
+                    is_to_the_right = true;
+                }
             });
         }
 
--- a/lib/Widgets/DynamicNotebook.vala
+++ b/lib/Widgets/DynamicNotebook.vala
@@ -322,7 +322,7 @@ namespace Granite.Widgets {
                     uint tab_position = dynamic_notebook.get_tab_position (this);
                     close_other_m.label = ngettext (_("Close Other Tab"), _("Close Other Tabs"), num_tabs - 1);
                     close_other_m.sensitive = (num_tabs != 1);
-                    close_other_right_m.label = ngettext (_("Close Tab to the right"), _("Close Tabs to the right"), num_tabs - 1 - tab_position);
+                    close_other_right_m.label = ngettext (_("Close Tab to the Right"), _("Close Tabs to the Right"), num_tabs - 1 - tab_position);
                     close_other_right_m.sensitive = (tab_position < num_tabs - 1);
                     new_window_m.sensitive = (num_tabs != 1);
                     pin_m.label = "Pin";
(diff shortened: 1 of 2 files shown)


lib/Widgets/DynamicNotebook.vala
```diff
@@ -1184,18 +1194,23 @@ namespace Granite.Widgets {
...
+            });
+        }
 
-                num = n_tabs;
-            }
+        private void on_close_others_right (Tab clickedTab) {
+            var toTheRight = false; 
+
+            tabs.copy ().foreach ((tab) => {
+                if (toTheRight)
```
jlnr: One-line `if` statements should also have curly braces: https://elementary.io/docs/code/reference#indentation

lib/Widgets/DynamicNotebook.vala
```diff
@@ -1184,18 +1194,23 @@ namespace Granite.Widgets {
...
 
-        private void on_close_others (Tab tab) {
-            var num = 0; //save num, in case a tab refused to close so we don't end up in an infinite loop
-
-            for (var j = 0; j < tabs.length (); j++) {
-                if (tab != tabs.nth_data (j)) {
-                    tabs.nth_data (j).closed ();
-                    if (num == n_tabs) break;
-                    j--;
+        private void on_close_others (Tab clickedTab) {
```
ryonakano:
```suggestion
        private void on_close_others (Tab clicked_tab) {
```

lib/Widgets/DynamicNotebook.vala
```diff
@@ -1184,18 +1194,23 @@ namespace Granite.Widgets {
...
+            tabs.copy ().foreach ((tab) => {
+                if (tab != clickedTab) {
+                    tab.closed ();
                 }
+            });
+        }
 
-                num = n_tabs;
-            }
+        private void on_close_others_right (Tab clickedTab) {
```
ryonakano:
```suggestion
        private void on_close_others_right (Tab clicked_tab) {
```

lib/Widgets/DynamicNotebook.vala
```diff
@@ -1184,18 +1194,23 @@ namespace Granite.Widgets {
...
+            var toTheRight = false;
```
ryonakano:
```suggestion
            var is_to_the_right = false; 
```

andreasfelix: done

lib/Widgets/DynamicNotebook.vala
andreasfelix: done

lib/Widgets/DynamicNotebook.vala
andreasfelix: done

andreasfelix:
@ryonakano, @jlnr thanks for your suggestions! I have implemented them.

andreasfelix:
Should it be the words *Tab* and *Right* be uppercase?

Which one do you prefer?
- *Close tabs to the right* 
- *Close Tabs to the right*
- *Close Tabs to the Right*

ryonakano:
> Should it be the words Tab and Right be uppercase?

> Which one do you prefer?

> * Close tabs to the right
> * Close Tabs to the right
> * Close Tabs to the Right

Maybe we may want to use [AP Style Title Case](https://www.bkacontent.com/how-to-correctly-use-apa-style-title-case/) as encouraged to do in https://elementary.io/docs/human-interface-guidelines#title-case, thus "Close Tabs to the Right"?

lib/Widgets/DynamicNotebook.vala
```diff
@@ -1179,18 +1189,25 @@ namespace Granite.Widgets {
...
-        private void on_close_others (Tab tab) {
-            var num = 0; //save num, in case a tab refused to close so we don't end up in an infinite loop
-
-            for (var j = 0; j < tabs.length (); j++) {
-                if (tab != tabs.nth_data (j)) {
-                    tabs.nth_data (j).closed ();
-                    if (num == n_tabs) break;
-                    j--;
+        private void on_close_others (Tab clicked_tab) {
+            tabs.copy ().foreach ((tab) => {
```
jlnr: It looks like `tabs` is recreated every time the property is being accessed: https://github.com/andreasfelix/granite/blob/close_tabs_to_the_right/lib/Widgets/DynamicNotebook.vala#L679-L688

I'm not sure if `copy` is necessary or wasteful in this situation. I also don't quite understand why there's a `_tabs` field to hold the value if it's being recreated anyway :thinking:

andreasfelix:
Ok sorry to bother you. I will read the guidelines!

lib/Widgets/DynamicNotebook.vala
jlnr: Maybe the reasoning is similar to this temporary (seven year old :grimacing:) field?

https://github.com/andreasfelix/granite/blob/close_tabs_to_the_right/lib/Widgets/DynamicNotebook.vala#L671

andreasfelix: I added the `copy`, because otherwise I get a segmentation fault.

jlnr: Ah right, because returned properties are not owned, and the next "call" to `tabs` will destroy the old list you're iterating over. All good then :+1:

Issue #241, opened 2018-11-04
Python bindings issues


ivankra:
I'm trying to use libgranite from python via gir bindings and running into a couple of issues with them.

First, .gir file that's produced during the build is missing a `shared-library=` attribute, as a result you can't instantiate any Granite.* classes:

```
$ python3 -c 'from gi.repository import Granite; Granite.WidgetsSourceList()'
...
TypeError: could not get a reference to type class
```

I reported this at first to Debian ([bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=911910), [patch](https://sources.debian.org/patches/granite/5.1.0-2/set-shared-library-attribute.patch/)) as I thought it was distro problem, but the culprit seems to build scripts in this repository and seems to also affect Elementary OS. Please consider cherrypicking their patch.

Second issue I ran into is that overriding `get_context_menu` virtual method for Granite.Widgets.SourceList.Item doesn't work, and as a result I can't create a custom menu for SourceList:

```
#!/usr/bin/env python3
import gi
gi.require_version('Gtk', '3.0')
gi.require_version('Granite', '1.0')
from gi.repository import Granite, Gtk

class Item(Granite.WidgetsSourceListItem):
    def __init__(self):
        super().__init__(name='item')

    def do_get_context_menu(self):
        print('do_get_context_menu called')
        return Gtk.Menu()

item = Item()
print(item.get_context_menu())
```
Output: None, and my virtual method override (with `do_` prefix) never gets called.

Not sure if the bug is in granite or vala or (py-)gobject. Do you have any ideas how make it work from python?

jmc-88:
The [Debian patch](https://salsa.debian.org/pkg-deepin-team/granite/commit/ea031175f804fd0b88c300d126444cad1aad0635) added this flag to the CMake build, which has been since removed in favour of Meson, and that does have the `--shared-library` flag as expected: https://github.com/elementary/granite/blob/f1b29f52e3aaf0f5d6bba44c42617da265f679c8/lib/meson.build#L103-L104

So, the first reported issue should be fixed by now and just needs a new release. Building it locally from head and running the following completes without issue:

```shell
$ python3 -c 'import gi; gi.require_version("Granite", "1.0"); from gi.repository import Granite; Granite.WidgetsSourceList()'
```

As for the second part of the report, I don't understand the need for the `do_` prefix. Overriding the `get_context_menu()` method works as expected.

ivankra:
[...]

PR #297, opened 2019-07-07
Remove path from --shared-library flag

jmc-88:
See issue #241 for context.
Passing libgranite.full_path() to g-ir-compiler during the build
results in a typelib which references the local build path, which
doesn't correspond to the actual build path in the binary that's
shipped to the users in the packaged version of libgranite.
Removing the path component and leaving only the output library
filename here still works as expected.

### DIFF
--- a/lib/meson.build
+++ b/lib/meson.build
@@ -102,11 +102,12 @@ custom_target(
     command: [
         g_ir_compiler,
         '--shared-library',
-        libgranite.full_path(),
+        '@PLAINNAME@',
         '--output',
         '@OUTPUT@',
         join_paths(meson.current_build_dir(), granite_gir),
     ],
+    input: libgranite,
     output: granite_typelib,
     depends: libgranite,
     install: true,


lib/meson.build
```diff
@@ -101,7 +101,7 @@ custom_target(
     command: [
         g_ir_compiler,
         '--shared-library',
-        libgranite.full_path(),
+        libgranite.full_path().split('/')[-1],
```
jlnr: This is not my area of expertise, but I've taken a look to see how other libraries handle this: https://github.com/search?q=language:meson+"--shared-library"&type=Code

I think that splitting the string is a bit unconventional, maybe better to use one of the libraries from above (granite_gi?) to build this argument?

jmc-88: I did check roughly the same source, and I resorted to splitting the path because the [`library()`](https://mesonbuild.com/Reference-manual.html#library) rule can potentially alter the output file name in several ways (e.g. through custom `name_prefix` or `name_suffix`).

Though, checking the same search again, though, I noticed a neat way of achieving the same in [`saga-glib`](https://github.com/major-lab/saga-glib/blob/9b7b5618e259bb88e84b137d339c3520790db200/src/meson.build#L36-L37): adding the library as `input` and referring to it via `@PLAINNAME@` which does exactly what I need here without having to resort to string operations ([documentation](https://mesonbuild.com/Reference-manual.html#custom_target)).

I'll try to update the commit accordingly.

jmc-88: After a bit of fight with Git, I think I managed to update it cleanly (5fa5bbbcb0075796b2667a5ee6098221ebc935ab). PTAL.

jmc-88: Ping?

jlnr: The new commit looks much cleaner :+1:, but I'd prefer someone with more knowledge about the Debian packaging to take a look at this.

jmc-88:
> it looks good to me

Thanks. Would you mind rubberstamping again?
Sorry, since I rebased it killed the previous commit. I guess next time I'll let it pollute the history with merge commits :(

jmc-88:
Ping?

PR #299, opened 2019-07-08
More efficient code for closing to the right

donadigo:
A follow up to #296.

There is no need to create a new variable and copy the list to remove all tabs to the right.
We find where the `closed_tab` in the tab list is, get the next position and iterate from there until we hit the end of the list.


andreasfelix:
After installing this branch, I get a segmentation fault when closing others in elementary-code.

donadigo:
@andreasfelix oops, thanks for noticing, I created a revert branch for now.

andreasfelix:
No problem! I would also be interested to find a solution where the copying is not needed...

jlnr:
The easiest way would be to move the body of the `tabs` property into a method (which, unlike a property, transfers ownership to the caller):

```vala
        private List<Tab> copy_tabs () {
            List<Tab> tabs = new List<Tab> ();
            for (var i = 0; i < n_tabs; i++) {
                _tabs.append (notebook.get_tab_label (notebook.get_nth_page (i)) as Tab);
            }
            return tabs;
        }

        List<Tab> _tabs;
        public List<Tab> tabs {
            get {
                return _tabs = copy_tabs ();
            }
        }
```

The `tabs` property would still behave as before, keeping existing code intact, but internally you can now use `copy_tabs ()` instead of `tabs.copy ()`.

I don't think the extra code complexity is worth the collective CPU time this will save (it halves the linked list allocations, but why are we using a linked list in the first place?). But it's probably a good reason to make anything that allocates memory a method instead of a property in the future.

andreasfelix:
After reading the parts of the doc about ownership and properties I am still a bit confused as they are introduced as an more elegant/concise notation of setter and getter methods (also to hide implementation details). So what is the rational to make them behave differently then setters and getters?

Does this not also make (unowned) properties practically useless for all cases of where you want to do something with the data (like iterating over a list) as you can never be sure that they are not freed by the owner?

Also in the docs they say to achieve a strong reference on:
> The keyword owned can be used to specifically ask a property to return a owned reference of the value ...
>```vala
>public owned Object property { owned get { return new Object(); } }
>```
I tested this for the dynamic notebook and the `owned` keyword would also prevent seg faults. However in the next sentence, it is actively discourged:
> ...  in general they should not be widely used in the programs unless you know what you are doing.

jlnr:
In theory, the `owned` property solves this issue, but it would introduce leaks in existing binaries that read `DynamicNotebook.tabs` without freeing the returned value.

I think the rationale behind these rules is this:

* When an object has a persistent instance variable like `string _title`, then a property `public string title { get { return _title; } }` does the right thing: outside code can access `some_object.title` without any extra allocation or cleanup.
* When an object needs to compute a value, say `string build_title_from_components ()`, Vala also does the right thing: the method can create a temporary object without storing it in an instance variable, and it's up to the caller when to free the result.

But when we use a _property_ for a _computed_ value, we end up with this fragile construct where the object needs a backing `_tabs` variable just to make things work. And in the other direction, when we use a method to retrieve a `string` or `List` that already exists, there will be an extra copy (because these basic data types are not reference counted, grrrrrrrr).

tl;dr I think `tabs` should have been a method all along :)

andreasfelix:
So we should use properties just for persistent variables and methods for computed variables?

Wouldn't it be possible to make the `tabs` list persistent, so that instead of creating a new list whenever `tabs` is accessed, the old list gets updated whenever a tab gets inserted/deleted. To my understanding this would also not break existing code which uses the `tabs` variable.

jlnr:
Yes to both, but keeping two lists in sync adds even more complexity and potential for bugs IMHO.

PR #300, opened 2019-07-08
Add AccelLabel

danrabbit:
Opening as a draft for comment

Screenshot from 2019-07-09 08 02 52@2x
Screenshot from 2019-07-09 08 03 24@2x
Screenshot from 2019-07-09 08 13 49@2x
Screenshot from 2019-07-09 08 14 44@2x


cassidyjames:
@danrabbit is it unreasonable to try and use keycap markup here? It would look nice plus might take up less space since it removes the `+`

cassidyjames:
(I think we skipped that for tooltips because there wasn't a way to style it properly.)

danrabbit:
@cassidyjames that's interesting.

Screenshot from 2019-07-08 15 24 47@2x

Maybe a little busy? Could make it flat or something:

Screenshot from 2019-07-08 15 29 11@2x

cassidyjames:
@danrabbit it definitely adds a bit of weight, but I suppose any Accel labels will. I like it being flatter since it looks less like a pressable button.

danrabbit:
Screenshot from 2019-07-08 16 40 50@2x
Screenshot from 2019-07-08 16 43 16@2x

peteruithoven:
Any change you could use this somewhere in the `granite-demo`? That would make it easier to experiment with the style

jeremypw:
If there is uncertainty about styling, it might be worth merging a Granite widget with current styling so that it can be used in apps now simplifying PRs such as https://github.com/elementary/terminal/pull/387.  New styling can be introduced later.

danrabbit:
Keycap style was merged here: https://github.com/elementary/stylesheet/pull/524

jeremypw:
Unfortunately you cannot construct with an action name. Maybe an alternate constructor accepting an Application and and action name as parameters?

danrabbit:
@jeremypw there is a constructor `from_action_name` here https://github.com/elementary/granite/pull/300/files#diff-cdea9748cbaa0ab810ca26680d8a2856R71

Is that not what you wanted?

PR #301, opened 2019-07-09
More efficient closing other tabs

donadigo:
No need for copying the tab list. Just iterate and remove all other tabs.

### DIFF
--- a/lib/Widgets/DynamicNotebook.vala
+++ b/lib/Widgets/DynamicNotebook.vala
@@ -1191,11 +1191,12 @@ namespace Granite.Widgets {
         }
 
         private void on_close_others (Tab clicked_tab) {
-            tabs.copy ().foreach ((tab) => {
+            for (unowned List<Tab> list = tabs; list != null; list = list.next) {
+                var tab = list.data;
                 if (tab != clicked_tab) {
                     tab.closed ();
                 }
-            });
+            }
         }
 
         private void on_close_others_right (Tab clicked_tab) {


jlnr:
If any listener of the `closed ()` signal happens to access (just reading) the `tabs` property, the list you are iterating over will be destroyed. FWIW, [@andreasfelix was indeed reporting segmentation faults](https://github.com/elementary/granite/pull/296#discussion_r300870338).

donadigo:
@jlnr this is different than iterating over with an index. Removing the element from `GList` will cause `.next` to update so there is no worry about accessing elements that were deleted. See: https://github.com/elementary/granite/pull/299

jlnr:
Every time the `tabs` property is accessed, the `_tabs` list will be destroyed and recreated, with no reference counting because it's a GList. If `on_close_others` iterates over `tabs`, but (for example) `closed ()` internally writes `tabs.size ()` to a log file, then that will destroy the GList while an iteration is in progress. The *elements* will still be there, but the linked-list nodes won't, and accessing the `next` pointer in the loop can crash. At least that is my understanding from following the C source code.

andreasfelix:
Doesn't the `unowned` keyword keep the list alive?

*From Vala-doc:*
> assign the reference to a variable that is declared with the unowned keyword and Vala will know that it should not free the object on your side.

I can confirm that I don't experienced any seg faults with this branch ...

jlnr:
I understand this paragraph the other way around: `unowned` here means that when the local `list` variable in `on_close_others` goes out of scope, the GList it points to is not being freed (on the "side" of this function), instead the `_tabs` instance variable keeps it alive.

Maybe there is no error right now, but this change makes it easy to introduce one. For example, if I add this line to the top of `on_tab_closed`:

```vala
            warning (@"$((int) tabs.length) tabs left");
```

...then the "close other tabs" functionality breaks. This will be really hard to debug, and the `copy ()` is not happening in a loop, but in response to user action. I know it feels wrong to `copy` a list that is only being used once, which is why I brought the topic up during review, but I'm now convinced that it's the best option. (Unless we can somehow refactor the `tabs` property.)

andreasfelix:
Ok your interpretation of paragraph makes more sense, especially with regard to the term `unowned`. I was confused because to me it sounded like an equivalent alternative to the copy method:

> So when you call a method  that returns an `unowned` reference to an object that you are interested in, then you have two choices: either copy the object if it has a copy method, then you can have your own single strong reference to the new copied object, or assign the reference to a variable that is declared with the `unowned` keyword ...

jlnr:
Right. I think the docs on this could really be a bit more clear. To that end, thanks to you two for looking into these optimizations in the first place! I always avoided memory management in Vala, but after navigating the generated DynamicNotebook.c for a bit, it finally starts to make sense :)

jeremypw:
This fails if the `tabs` property is accessed during the loop.  You can test this by inserting the a line such as:

`warning ("tabs length %u", tabs.length ());`

Inside the for loop in on_close_others.

If e.g. Files is compiled against this modified library then choosing "close other tabs" from the tab context menu only results in one or none tabs closing (depending on which tab is active).
If the above warning line is removed, and Granite/Files recompiled then all but the active tab closes.

I am not sure how likely it is that some other process would access the tabs property in the middle of closing other tabs (this is harder to test) but is the performance gain here worth the risk?

Issue #239, opened 2018-10-18
Expand tab width to fill tab bar


D0J0P:
I started a [post](https://plus.google.com/photos/photo/104362843500611550798/6613365928764600546?sqid=104613975513761463450&ssid=f519a7cd-31d2-4c2a-a6a1-f3bb617eb0fd) on elementary's Google+ page after playing with dconf settings for Epiphany, and found I could expand the tab width to fill the tab bar. Here's some points I made in the conversation: 

"+Cassidy James Blaede Other than Grzegorz G.'s suggestion to wait until the mouse is off the tab bar to change width, there's also the way Mac OS does it here: [youtube.com - Use Tabs in macOS Sierra](https://www.youtube.com/watch?v=3jzb4FtVulo)

You can see there's a moment when he closed a tab, the close button for the next one landed where his mouse was. During this time, I'm sure he could have closed the next tab. Doing things this way keeps all the functionality. So we don't have to do it the exact way Epiphany does it, but we can do it in a way that doesn't make closing lots of tabs rapidly any worse. "
...
"It would even give a potential affordance. By keeping the tabs at the same minimum width when you have a lot of them, it doesn't make the tab description any shorter. When you have less tabs, however, you can read more of the tab description."

Can we make this the new look for tabs bar in elementary?

andreasfelix:
Expanding the tabs to fill the tab-bar would remove the functionality to create a new tab (restore the last closed tab) when double-clicking (middle/three-finger clicking) on an empty part of tab-bar.

But this would also solve #221, which is caused by the event propagation of the `button_press_event` from the left/right arrows to the dynamic notebook.

As the post is no longer available, I made two screenshots:
![Filled_tabs_dark](https://user-images.githubusercontent.com/24651767/60687821-51c86b80-9eb1-11e9-941c-90dad6b54714.jpeg)
![Filled tabs](https://user-images.githubusercontent.com/24651767/60687819-51c86b80-9eb1-11e9-80da-ec5a6866af50.jpeg)

andreasfelix:
![filled](https://user-images.githubusercontent.com/24651767/60693289-0c6c6400-9eda-11e9-811a-a9077e7f43f8.gif)


I did a first implementation of a tab-bar with filled tabs. Similar to macOS/Chrome the tabs keep their witdh until the cursor leaves the tab-bar, which is convenient when rapidly closing multiple tabs.

jlnr:
[...]

Issue #17, opened 2017-04-21
Delay resizing tabs when closed


elementaryBot:
Firefox and Chromium provides "smart tabs". That is; when you close a tab the close button of the right most tab is getting in a position where the cursor is hovering over it. If you click then same thing happens again, if you wait more than 2 or 3 seconds or move away the cursor then the tabs sizes gets adjusted, if needed.

Thanks for reading.

Launchpad Details: [#LP1214933](https://bugs.launchpad.net/bugs/1214933) Karl-Oskar "klrr" Rikås - 2013-08-21 14:23:10 +0000

elementaryBot:
Might be a bit tricky to implement but I love the idea and I had never noticed it even though I use it every day on Firefox.

Launchpad Details: [#LPC](https://bugs.launchpad.net/granite/+bug/1214933/comments/1) David Gomes - 2013-08-22 14:54:03 +0000

elementaryBot:
Getting ~elementary-design to confirm this.

Launchpad Details: [#LPC](https://bugs.launchpad.net/granite/+bug/1214933/comments/2) David Gomes - 2013-08-22 14:54:24 +0000

elementaryBot:
IMO the expand should not be timed but triggered when the cursor leaves the tab bar

Launchpad Details: [#LPC](https://bugs.launchpad.net/granite/+bug/1214933/comments/3) Santiago Ocamica - 2013-09-05 00:51:44 +0000

elementaryBot:
Here's a patch changing the timeout for a connection to the cursor leave signal. I hope it helps.

Launchpad Details: [#LPC](https://bugs.launchpad.net/granite/+bug/1214933/comments/4) Santiago Ocamica - 2013-09-05 01:25:22 +0000

elementaryBot:
https://www.bountysource.com/issues/922295-delay-resizing-tabs-when-closed-50

Launchpad Details: [#LPC](https://bugs.launchpad.net/granite/+bug/1214933/comments/5) Daniel Fore - 2015-06-17 23:50:39 +0000

cassidyjames:
In looking at the behavior of both Chrome and Firefox, there's no time delay, they just delay the resizing until the cursor is not hovering the tab bar. I really like that behavior and think that's the way to go.

danrabbit:
This is now fixed

Issue #82, opened 2017-07-05
Middle-click closes tabs on press instead of press + release


cassidyjames:
Similar to #39, it looks like we're using middle-click press to fire the tab close instead of middle-click press + release (which is the standard way to do clicks in GTK+).

danrabbit:
It looks like moving this to on_release causes a weird tab duplication issue for some reason. We'll need to solve that before re-introducing that change

ls42:
Hi, coming because of the `Bitesize` tag. @danrabbit would you like to point in me in a direction on how to debug this? Been trying to reproduce this behavior but didn't get any weird issues so far. But I've been only tinkering with Code tabs for a few minutes. Absolute newbie on granite, vala and GTK dev. in general.

Where would be the issue located in your opinion? Do you think it's somewhere in diff a3c083e or somewhere up the library chain? Maybe `button_release_event`? `button_release_event.connect`?

Sorry, trying to be helpful but don't really know what to ask exactly.

feoh:
Hi. Just FWIW I can't repro this either. Tried in Terminal and Code, and in both cases Middle click still leaves the tab open, but when you release it vanishes as desired.

Don't want to just resolve in case there's more evidence required here.

danrabbit:
Closing since we removed DynamicNotebook in favor of Hdy.Tabbar

Issue #221, opened 2018-10-03
Dynamic notebook: Double-clicking on horizontal arrows should not open new tabs


jlnr:
* Open enough tabs to make the the left/right arrows appear
* Double-click either arrow

Expected:

* Focus moves two tabs to the left/right

Actual:

* Focus moves by one tab, then a new tab is opened (at least in Files and Code)

Maybe this is also what causes https://github.com/elementary/code/issues/485 (which I can also reproduce in Files): All double-clicks, even on arrows/even with the middle mouse button, are interpreted as a "open new tab" gesture?

andreasfelix:
I think the double click on the left/right arrows opens a new tab because the `button_press_event` gets handled by the `DynamicNotebook` class:
```vala
button_press_event.connect ((e) => {
    if (e.type == Gdk.EventType.2BUTTON_PRESS && e.button == 1) { // <-- double click opens new tab!
        new_tab_requested ();
    } else if (e.button == 2 && allow_restoring) {
        restore_last_tab ();
        return true;
    } else if (e.button == 3) {
        menu.popup_at_pointer (e);
    }

    return false;
});
```

I think the `left/right-arrow` buttons have to consume the `button_press_event` explicitly, similar to the `add_button` (the button with the + symbol which creates new tabs):
```vala
add_button.button_press_event.connect ((e) => {
    // Consume double-clicks
    return e.type == Gdk.EventType.2BUTTON_PRESS && e.button == 1;
});
```
These lines prevent the creation of a new tab when double clicking the `add_button`. So I think a similar line is needed for the `left/right arrow` buttons. Unfortunately I can't find the declaration of these buttons in the `DynamicNotebook.vala` file ...

danrabbit:
Closing since we removed DynamicNotebook in favor of Hdy.Tabbar

PR #302, opened 2019-07-10
implement filled tab-bar

andreasfelix:
Fixes: #239
Also fixes: #17, #220, #221

Removes the functionality to create a new tab (restore the last closed tab) when double-clicking (middle/three-finger clicking) on an empty part of tab-bar.

Files
Code
Terminal

Suggestions welcome!

**Open Design Decisions:**
- [ ] Should hovering over the `add_button` resize tabs?
- [ ] Should a single tab be resized instantly (while still hovering over tab-bar)?
- [ ] Should the `close_tab_button` be visible for a single tab (without hovering)?
- [ ] Should there be a resize animation?

**Implementation:**
- closing a tab with a `close_button` or `middle_click` freezes the current
  tab width and sets the `need_to_expand` flag
- leaving the tab-bar will expand the tabs again, if the `need_to_expand` flag is set
- the current `PageTabContainer`, the `DynamicNotebook` and the `add_tab_button` are used to signal the `leave_tab_bar_event`
- this can be done as:
        *tab_bar_area* **=** *dynamic_notbook_area* **-** *current_page_tab_container_area* **-** *add_tab_button_area*
- so leaving the `DynamicNotebook`, entering the current `TabPageContainer` or entering the `add_tab_button`  will cause a check if the tabs have to be expanded again (equivalent to leave_tab_bar_event)


andreasfelix:
last two commits also fix #82

The problem of #82 was that the middle-click `button_press_event` was not consumed by the tab and the event was propagated to notebook. 

Even though the middle/click on tab can't conflict with middle-click on notebook for the filled tab bar, I added a line to explicitly consume this event.

andreasfelix:
I played around with the style in the inpsector, so that a single filled tab has no border. What do you think?

changes style
change_style code

**CSS:**
```css
notebook tab:nth-child(2):nth-last-child(2) {
    border: none;
    border-width: none;
    background: none;
    box-shadow: none;
}
```

donadigo:
I personally don't like this change. It seems way harder to scan what tabs you have currently open because they are much wider and have the same styling. I believe that we're aiming for having tabs behave as closely as possible to how Chrome does it.

I think there's two changes from this PR that would be definitely accepted:
* increased default maximum width for tabs
* do not resize tabs when closing them on hover

andreasfelix:
@donadigo A counterargument would be, that you only  have to really scan the tab-bar if more than 3 tabs are open. And in this case the tabs are not much wider anyway.

The main argument for the filled tabs is, that it would allow display longer labels. This would be especially helpful for code/files, in the case you two files with the same name. (see below for gala example)

**Currently**
code small

**Filled tabs** (optional folder name has to be implemented)
filled

Another possibility would be to make the tab width depend on the current label. (look at vscode)

A fourth option would be a combination be a filled tab bar, where tabs which need more space would be wider than others.

glimpster:
Great stuff Felix!

andreasfelix:
As I see it, there are 4 different options, which would solve the problem of the non-readable tabs:

1. non-filled equal-spaced tabs with larger minimum width
2. non-filled variable-spaced tabs (similar to vscode, where tabs can have different width)
3. filled equal-spaced tabs (this PR)
4. filled variable-spaced tabs

I would be ok to change this PR to any of them, but I personally prefer 3 or 4.

Another question is if the tab size should depend on the window width. Currently the tab-width gets updated whenever the window size changes:

https://github.com/elementary/granite/blob/f5590325f92430d8b8c88d175a99119306c76bc3/lib/Widgets/DynamicNotebook.vala#L841-L843

What is the point of seeing all tabs at once when you can't read the label? I would rather click the arrow-button then having tab labels with "Do...ds" and "Do..nts". At least, I think there should be a minimum width so that normal length labels don't get truncated.

---

How 4. would look like:

filled_with_variable_width

jlnr:
I like the GIF variable-spaced tabs in Code (approach 2), but I think it would be distracting in Epiphany, Terminal, and Files, where tabs change their titles while navigating around.

Since Code already uses a different tab style, I wouldn't mind if it adopted variable-spaced tabs (probably implemented as an option here in Granite), but I'm not sure if it should be the default.

andreasfelix:
Good point, I didn't thought about this! So variable-spaced tabs only make sense for static labels and changing labels should always be same width.

I could implement an option, so that the different applications could chose between the 4 different tab-styles. This could be simple done with two bools `filled_tab_bar` and `equal_spaced_tabs`.

jlnr:
Hey @andreasfelix, I've just closed #239. It was good to have an implementation of filled tabs to play around with, but the UX team consensus was that filling the whole tab bar is too much.

I would still love to see #17 and the other issues fixed, and maybe you can also re-target this PR to increase the initial tab width? The current default size can definitely feel cramped, especially with "Larger" fonts. (Could the font size maybe be taken into account for the initial tab width?)

Variable tab width seems like a separate feature request that would also need consideration from the UX team. Before this PR gets too messy, I would suggest opening a new issue for that in granite, even if it mostly applies to Code for now.

andreasfelix:
Should this initial tab width depend on the the current window width? If not there would be one static tab width, which would automatically solve #17. 

Also #221 was only solved as a side effect when removing the functionality to create a new tab (restore the last closed tab) when double-clicking (middle/three-finger clicking) on an empty part of tab-bar. So it will remain for the non-filled tab bar.

jlnr:
@andreasfelix I don't think the initial tab width should depend on the window width (unless the window is like 50px wide). #17 is still relevant because, even though tabs can't _grow_ indefinitely, they do still _shrink_, and that's when the same-width magic can kick in.

So basically, tab widths should be dynamic, but with an upper limit that is somewhere above the current default size (not sure what feels reasonable, maybe 150% of what it is now?).

andreasfelix:
@jlnr I did moved the delay and the larger min/max tab width to a new PR (#311).

PR #303, opened 2019-07-10
Allow to change active tab by scrolling on tab bar 

andreasfelix:
Fixes: #250 

Has Same Problem as https://github.com/elementary/stylesheet/issues/523.

The feature was already implemented, just had to set the `Gdk.EventMask.SCROLL_MASK` ...

PR #311, opened 2019-08-16
Delay tab resizing when closing tabs and larger min/max tab width

andreasfelix:
follow up to #302 

#302 was rejected but some of its changes could also be used for non-filled tabs:
- delay resize tabs
- introduce minimum width for tabs
- larger maximum width for tabs
- activate scroll mask to switch tabs
- use release event of middle click to close tabs


lib/Widgets/DynamicNotebook.vala
```diff
@@ -818,6 +839,11 @@ namespace Granite.Widgets {
             notebook.set_action_widget (add_button_box, Gtk.PackType.START);
             notebook.set_action_widget (restore_button, Gtk.PackType.END);
 
+            //  delay tabs resizing until cursor leaves tab-bar
+            //  tab_bar-area = DynamicNotebook-area - TabPageContainer-are
```
jlnr: Typo: are->area
But maybe better to explain the formula in `recalc_size` instead?

lib/Widgets/DynamicNotebook.vala
```diff
@@ -281,42 +296,32 @@ namespace Granite.Widgets {
             duplicate_m.activate.connect (() => duplicate () );
             pin_m.activate.connect (() => pinned = !pinned);
 
+            add_events (Gdk.EventMask.SCROLL_MASK);
```
jlnr: Just for clarification -- does that mean we had code to handle scroll_events, but they never arrived because of the missing mask?

lib/Widgets/DynamicNotebook.vala
```diff
@@ -336,6 +341,20 @@ namespace Granite.Widgets {
...
 
+            this.button_release_event.connect ((e) => {
+                if (e.button == 2 && cursor_over_tab) {
+                    e.state &= MODIFIER_MASK;
+                    if  (e.state == 0) {
+                        dynamic_notebook.close_tab_and_keep_width (this);
+                    } else if (e.state == Gdk.ModifierType.SHIFT_MASK) {
+                        this.close_others ();
+                    }
+                } else {
```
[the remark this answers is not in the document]
andreasfelix: I think it is reachable, if `e.button == 2 && cursor_over_tab` evaluates to `True`?

lib/Widgets/DynamicNotebook.vala
andreasfelix: as far as I understand it, yes! Would it maybe better to remove it in this PR and merge the other PR for cleaner commit history?

andreasfelix: I removed the scroll-able tabs from this PR as there is already a separate PR (#303) which will enable this behavior.

andreasfelix: Never mind, I think it is better to include the scroll-able tabs in this PR, because #303 would raise a merge conflict as does not has the new `dynamic_notebook` property of this PR.

lib/Widgets/DynamicNotebook.vala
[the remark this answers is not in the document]
andreasfelix: However, it was possible to simplify the statement: 8ebb325

jlnr:
Thanks for the tweaks, very happy to merge this :)

Issue #313, opened 2019-08-26
Add the middle-split window as a special type of Window


aggalex:
Many apps are already using it, and adding it to granite will definitely make it more accessible to developers and more readable (by hiding the code of just constructing the window into eg. a class that inherits from Gtk.Window)

## Examples of apps using it:
- [Cyfrif](https://appcenter.elementary.io/com.github.aimproxy.cyfrif)
- [Ion Color Generator](https://appcenter.elementary.io/com.github.alainm23.ioncolorgenerator)
- [Harvey](https://appcenter.elementary.io/com.github.danrabbit.harvey.desktop)
- [Tasks](https://github.com/elementary/tasks)

danrabbit:
As discussed on the associated pull request Hdy.Window is now available in elementary OS 6 and provides everything you need to make split headerbar windows easily

PR #315, opened 2019-08-30
Add Split Window

aggalex:
This pull request adds a new origin to granite (called "split-window"), which has all the code that is required to add the capability to Granite to have a Tasks-style window, similar to the layout of Harvey, Ion Colour generator, Cyfrif, etc. I have taken care of:

- Adapting the implementation that Dan's Tasks has into Granite
- Documenting stuff
- Updating Granite Demo app to include a demo of a split window
- Updating the elementary OS stylesheet to make sure things look fine (https://github.com/elementary/stylesheet/pull/540)
- Raising @cassidyjames' hype for this at GUADEC (lel)


aggalex:
Btw this pull request also fixes #313.

aggalex:
Btw this pull request also fixes #313.

jlnr:
I really like this style, but two issues when I tried it:

There's a black artifact in the header bar, which might be caused by my NVIDIA drivers or not installing this branch properly(?).

And I can drag the top divider further to the left/right than the bottom divider will follow, which breaks the illusion of it being a single tall divider. It jumps back when I focus the window.

splitwindow2

jlnr:
(Deleted earlier comment, I had installed the wrong stylesheet branch)

I really like this style, one minor issue: I can drag the top divider further to the left/right than the bottom divider will follow, which breaks the illusion of it being a single tall divider. It jumps back when the window loses and regains focus.

splitwindow3

Is it intentional that the sidebar header turns white (unlike the sidebar) when the window is not focused?

And for the granite-demo: What do you think about adding more sidebar items, so that the user can see how scrolling looks in the sidebar? And should the split window demo be shown modally so that the user cannot "lose" it behind the main granite-demo window?

danrabbit:
Alright, so I'm inclined to close this pull request now that we have Hdy.Window in 6.0. The Hdy.Window handles clipping of the corners but it also exists as a single container with no dedicated titlebar area, so you can easily pack a single Gtk.Paned into the window and we don't have to worry about keeping two panes in sync etc.

aggalex:
Does Hdy.Window handle the window moving nicely? If it doesn't have a titlebar, how can you move the window around?

aggalex:
Btw, I've been waiting for this to get settled to redesign contacts and fix some of its problems. If Hdy.Window is what you suggest, I might have a look at it after uni exams are over (they were supposed to end in late June, but covid yeeted the schedules)...

danrabbit:
Yes there are two approaches:  Hdy.WindowHandle which you can pack any widget into to create a draggable area with the window context menu

Hdy.HeaderBar which is pretty much Gtk.HeaderBar pre-packed into a Hdy.WindowHandle

aggalex:
Is libhandy going to be packaged with elementary OS? The package in the repos in Hera is too old, it doesn't have Hdy.Window.

danrabbit:
Yes, LibHandy will be available in elementary OS 6

PR #319, opened 2019-09-13
Get prefers color scheme from accountsservice

### LINKED ISSUE #276: Add way to check if the OS prefers a dark style

danrabbit:
Create a Granite.Settings class like Gtk.Settings. Fixes #276 

Read-only is intentional as we don't want apps setting this for the user


lib/Widgets/Settings.vala
```diff
@@ -0,0 +1,122 @@
...
+        public abstract string prefers_color_scheme { owned get; set; }
+        public abstract string time_format { owned get; set; }
+    }
+
+    [DBus (name = "org.freedesktop.Accounts")]
+    interface FDO.Accounts : Object {
+        public abstract string find_user_by_name (string username) throws GLib.Error;
+    }
+
+    public class Settings : Object{
```
cassidyjames: Lint

```suggestion
    public class Settings : Object {
```

cassidyjames:
We'll want to switch the AccountsService side to an enum since https://github.com/elementary/default-settings/pull/166 was merged.

danrabbit:
@cassidyjames Updated. This should work with the new accountsservice stuff.

Since there's no system UI for this yet, you can send this in dfeet `"io.elementary.pantheon.AccountsService", "PrefersColorScheme", GLib.Variant("u", 1)`

I guess there's a question of how should we demo this? I think we want the override here for the demo maybe? Or should we remove that toggle?

Also, we currently only include this for accountsservice in Odin. Should we add this in Hera too? Maybe not necessarily exposing the setting yet, but if it exists in Granite and AccountsService we could start testing it

Is it okay to make this a boolean property or should we make this enum out of the box? The boolean is nice because you can super easily bind this to app-prefers-dark right now. But this doesn't account for any prefers-light. I guess if we decide to support that we can come back and deprecate the boolean property?

cassidyjames:
>I guess there's a question of how should we demo this? I think we want the override here for the demo maybe? Or should we remove that toggle?

I think we default to the user preference but still allow a local override (so devs can still test how widgets look without changing their whole system preference)? But we can think about it. Another option would be to implement what we expect to see in apps like Code or Terminal where we might want to allow for a tri-state of light/dark/follow system.

>Also, we currently only include this for accountsservice in Odin. Should we add this in Hera too? Maybe not necessarily exposing the setting yet, but if it exists in Granite and AccountsService we could start testing it

Yeah I think we include it in Hera. I think I'd like to ship the system UI stuff for 5.2 if possible, because it has been a long time coming and would be a nice preview of the bigger dark style stuff for 6.0.

>Is it okay to make this a boolean property or should we make this enum out of the box? The boolean is nice because you can super easily bind this to app-prefers-dark right now. But this doesn't account for any prefers-light. I guess if we decide to support that we can come back and deprecate the boolean property?

Yeah I think boolean is fine and we can deprecate later if/when we expose a light preference.

cassidyjames:
Hmm, I guess this raises the question of how apps should bind to it. If it changes live, I imagine we'd want the app to change instantly, too. But if they offer an override, how does that interact? I think that's where it has to become a tri-state of:

- Follow system preference
- Always light
- Always dark

Which is similar to what I see in Android apps fwiw. I think _most_ apps that support a dark style should just automatically follow along with the setting, though.

cassidyjames:
I guess I'm fine in general with this as a starting point, we can always iterate. I would like to get a review of the code side from someone more familiar and opinionated about Granite and namespacing and everything, though.

lib/Widgets/Settings.vala
```diff
@@ -0,0 +1,94 @@
...
+                    "org.freedesktop.Accounts",
+                    user_path,
+                    GLib.DBusProxyFlags.GET_INVALIDATED_PROPERTIES
+                );
+
+                user_prefers_dark = pantheon_act.prefers_color_scheme == 1;
+
+                ((GLib.DBusProxy) pantheon_act).g_properties_changed.connect ((changed_properties, invalidated_properties) => {
+                    uint prefers_color_scheme;
+                    changed_properties.lookup ("PrefersColorScheme", "u", out prefers_color_scheme);
```
tintou, line 86:
```suggestion
                    int prefers_color_scheme;
                    changed_properties.lookup ("PrefersColorScheme", "i", out prefers_color_scheme);
```

tintou:
It would probably make it clearer indeed to return an enum instead of having a boolean, that way we can extend it without fearing API breaks

cassidyjames:
@tintou see: https://github.com/elementary/default-settings/pull/167#discussion_r414022506

We should decide on unsigned or signed and then make sure it matches in the master and hera branches of default-settings as well as here.

PR #333, opened 2019-11-06
Use async mutex for badge synchronisation

davidmhewitt:
The `lock` keyword only locks across different threads, rather than different calls to `async` methods on the same thread.

We were seeing an issue with AppCenter where the badge wouldn't appear and we'd get the following messages in the console:
```
Client.vala:95: An object is already exported for the interface com.canonical.Unity.LauncherEntry at /com/canonical/unity/launcherentry/3980099301
```

This was due to a race condition where `set_badge` and `set_badge_visible` were called in quick succession to each other.

In this case, the first call is still yielding on `GLib.Bus.@get` with `instance = null`, the second call sees that `instance == null` and continues to try and create a new instance and register on the bus again, which fails.

I've created this generalized async locking class that solves these kinds of issues. I've made it an internal API of Granite for now, but maybe we want to document this and make it public for other applications to use?

### DIFF
--- a/lib/Services/Application.vala
+++ b/lib/Services/Application.vala
@@ -25,33 +25,44 @@ namespace Granite.Services {
     namespace Application {
         [DBus (name = "com.canonical.Unity.LauncherEntry")]
         private class UnityLauncherEntry : GLib.Object {
+            private static AsyncMutex? instance_mutex = null;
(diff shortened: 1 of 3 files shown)


tintou:
It's nice to have it private, I think that it's better to not have any API stability promise on it

PR #369, opened 2020-01-08
Update time-based strings to be SI-compliant

getsnoopy:


danrabbit:
I don't think "8 h ago" is correct. This seems to be contrary to how other platforms handle this.

Here are screenshots from iOS and Android that seem to agree with how we do things already:

iu
notification-center-ios10-5802176a5f9b5805c2ed7163

cassidyjames:
I agree with @danrabbit, time strings are weird but it's best to stick with convention here.

getsnoopy:
It is correct, but iOS and Android all use ICU which has this bug. If you look at the Messages or Hangouts app on Android, however, it will show it correctly. I already have a ticket with ICU to have it fixed there, we should similarly fix it here as well.

cassidyjames:
Messages on Android says, “38 min” which could be ambiguous in certain contexts. Granite needs to retain the “ago” imho. I could be convinced of “38 mins ago” but that's a much longer string.

danrabbit:
I'm not really in favor of increasing the length of this string. It's typically used in places where space is at a premium and it's secondary or tertiary information

getsnoopy:
"38 min ago" would be correct, since the "min" and "h" and stuff are all symbols (not abbreviations), and they have the same form regardless of number (i.e., "1 min", "5 min"). But I think seeing as this is a grammatical error, we should really fix it. I've checked the places where the string is used, and it fits in all those places.

danrabbit:
It looks like this is the issue in ICU: https://unicode-org.atlassian.net/browse/CLDR-13511

I'd be interested in seeing what they have to say about this, but until it looks like other platforms may change this formatting I think it's best to stick with the more compact and familiar format we're already using

getsnoopy:
Right. Well there are already many platforms which are using the correct format, so there's precedence for this if that's what we're waiting for, let alone the fact that it's grammatically correct. It's actually not familiar because "5m" normally (ignoring the grammatical error) means "5 metres", and in this case it can mean "5 months" as well. It's quite confusing; there's really no reason to continue doing this.

PR #382, opened 2020-04-15
Added a flag to define accuracy of get_relative_datetime

### LINKED ISSUE #381: get_relative_datetime: Option to define accuracy of returned string
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Feature

`get_relative_datetime` returns a natural-language formatted time string like `Now`, `5m ago` or `Yesterday` which is great for any `DateTime` representation which is somewhen "Today".

But if it comes down to planning, a simple `Yesterday` or `Tomorrow` may not be enough - a `Yesterday at 10am` or `Tomorrow, 10:00` would be better.

**Describe the solution you'd like**

I'd love the possibility to pass a flag to `get_relative_datetime` which defines the returned string's accuracy in the users preferred formatting:

```
enum Granite.RelativeDateTimeAccuracy {
    AUTO,
    DATE,
    TIME
}

public string get_relative_datetime (
    DateTime date_time,
    Granite.RelativeDateTimeAccuracy accuracy = Granite.RelativeDateTimeAccuracy.AUTO
)
```

This way we can fine tune the accuracy of the returned natural-language formatted string where needed. With the flag passed the behaviour of the function would change like this:

**`get_relative_datetime (date_time)` - defaults to `AUTO`, same behaviour as before:**

- `2012-09-29 18:30` => `09/29/12`
- `2020-05-23 10:00` => `May 23`
- `2020-04-16 14:00` => `Tomorrow`
- `2020-04-15 10:55` => `5m ago`
- `2020-04-15 11:00` => `Now`

**`get_relative_datetime (date_time, Granite.RelativeDateTimeAccuracy.DATE)`**

- `2012-09-29 18:30` => `09/29/12`
- `2020-05-23 10:00` => `May 23`
- `2020-04-16 14:00` => `Tomorrow`
- `2020-04-15 10:55` => `Today`
- `2020-04-15 11:00` => `Today`

**`get_relative_datetime (date_time, Granite.RelativeDateTimeAccuracy.TIME)`**

- `2012-09-29 18:30` => `09/29/12 at 18:30`
- `2020-05-23 10:00` => `May 23 at 10:00`
- `2020-04-16 14:00` => `Tomorrow at 14:00`
- `2020-04-15 10:55` => `5m ago`
- `2020-04-15 11:00` => `Now`

**Additional context**

It would be great to have the proposed (or similar) functionality in the elementary Tasks app:

Screenshot from 2020-04-15 09-23-46

Screenshot from 2020-04-15 09-23-50

**EDIT:** Replaced the initial idea of passing a `bool` with passing an `enum`.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/91041590-get_relative_datetime-option-to-define-accuracy-of-returned-string?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github).
</bountysource-plugin>

marbetschar:
This pull requests adds the ability to define a relative accuracy for get_relative_datetime which fixes #381.

### DIFF
--- a/demo/Views/DateTimePickerView.vala
+++ b/demo/Views/DateTimePickerView.vala
@@ -26,6 +26,8 @@
 
 public class DateTimePickerView : Gtk.Grid {
     private Gtk.Label relative_datetime;
+    private Gtk.Label relative_datetime_day;
+    private Gtk.Label relative_datetime_time;
     private Granite.Widgets.DatePicker datepicker;
     private Granite.Widgets.TimePicker timepicker;
 
@@ -69,11 +71,23 @@ public class DateTimePickerView : Gtk.Grid {
         current_date.tooltip_text = date_format;
         current_date.xalign = 0;
 
-        var relative_datetime_label = new Gtk.Label ("Relative datetime:");
+        var relative_datetime_label = new Gtk.Label ("Relative datetime (auto):");
         relative_datetime_label.halign = Gtk.Align.END;
 
         relative_datetime = new Gtk.Label ("");
         relative_datetime.xalign = 0;
+
+        var relative_datetime_day_label = new Gtk.Label ("Relative datetime (day):");
+        relative_datetime_day_label.halign = Gtk.Align.END;
+
+        relative_datetime_day = new Gtk.Label ("");
+        relative_datetime_day.xalign = 0;
+
+        var relative_datetime_time_label = new Gtk.Label ("Relative datetime (time):");
+        relative_datetime_time_label.halign = Gtk.Align.END;
+
+        relative_datetime_time = new Gtk.Label ("");
+        relative_datetime_time.xalign = 0;
 
         set_selected_datetime ();
         datepicker.changed.connect (() => set_selected_datetime ());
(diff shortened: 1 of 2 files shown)


tintou:
We can't do this as it is an API break, please create another function like `get_relative_datetime_with_accuracy`

marbetschar:
@tintou done.

Created new `get_relative_datetime_with_accuracy` which is called by `get_relative_datetime` with `RelativeAccuracy.DEFAULT`.

danrabbit:
This seems a bit more API than what you actually need in your example. As far as I can see in your example, you only need `get_relative_date`, which ignores time.

I think we should continue to make sure we're only adding functions which have a solid use case in multiple apps

marbetschar:
@danrabbit I suspect the idea is that `get_relative_datetime` as well as `get_relative_datetime` should behave the same way regarding dates?

If this assumption is correct, can we change the behaviour of `get_relative_datetime` to return `Fr, Apr 17` instead of `Fri` for the `diff < 6 * TimeSpan.DAY && diff > -6 * TimeSpan.DAY` case because it is less ambiguous?

Right now, both methods behave different at this occasion:

**`get_relative_datetime`**

https://github.com/elementary/granite/blob/07fcb57929fa93dbd35e75d01288ee5eb90b505b/lib/DateTime.vala#L98

**`get_relative_date`**

https://github.com/elementary/granite/blob/07fcb57929fa93dbd35e75d01288ee5eb90b505b/lib/DateTime.vala#L125

marbetschar:
@danrabbit Added a Screenshot for the elementary Tasks app, to showcase how I think the relative date format strings should should look like: https://github.com/elementary/tasks/pull/98#issuecomment-615072657

Is this a suitable use case for Granite, or should we handle `get_relative_date` within Tasks as an app specific feature?

marbetschar:
Since Tasks is the only app with this specific use case, we handle it there.

PR #386, opened 2020-04-29
Add tab_bar_position->Gtk.PositionType to Widgets/DynamicNotebook.vala

jgoday:
Expose Gtk.Notebook->tab_pos through DynamicNotebook.
We can use it in io.elementary.terminal like: 
Captura realizada el 2020-04-29 19 42 51


ryonakano:
I personally don't think we want to add this property to Granite, because this property just accesses Gtk.Notebook.tab_pos. If developers want to change the position of tabs in their app, they can set a wanted value to `tab_pos` because Granite.Widgets.DynamicNotebook inherits Gtk.Notebook

jgoday:
Excuse my vala/granite ignorance but DynamicNotebook seems to inherits Gtk.EventBox, I think that I don't fully understand that ...

ryonakano:
Ah, I was dumb and totally wrong. Gtk.Notebook used in DynamicNotebook seems to be a private member so developers can't access. Ignore my comments :sweat_smile:

jgoday:
@ryonakano Thank you again for your review and comments ! :+1:

danrabbit:
I'm inclined to close this as "Design Conflict". One of the goals of Granite widgets is to ensure consistent design. I think it would go against the goals here to introduce an inconsistency in tabbar placement

@cassidyjames thoughts?

cassidyjames:
Yeah, I think a bottom tab design would better be served by a different widget, anyway. And a developer is always welcome to use their own custom widgets for their own uses; since we don't use this pattern anywhere in elementary OS or our apps, I would also be inclined to close it.

danrabbit:
Alright, sorry closing as "Design Conflict"! Thanks for the PR though :)

PR #388, opened 2020-05-02
Expose tab label tooltip

jeremypw:
Allow apps to set a tooltip with additional information rather than just repeating information that is already visible in the tab label (although the label could be ellipsized)


danrabbit:
Seems reasonable :) Can we add this to the demo?

jeremypw:
Sure, I'll do that.

danrabbit (approved):
Looks good :+1:

danrabbit (changes requested):
Oh I almost forgot. It needs a comment for generating docs :)

danrabbit (approved):

PR #389, opened 2020-05-03
build: add gsettings-desktop-schemas to pkgconfig Requires

worldofpeace:
While there's code in lib/Services/System.vala#history_is_enabled, which will return
if we can't find org.gnome.desktop.privacy, it still seems like we require this, and
we need to make sure applicaitons that use granite have it installed.
Otherwise, methods like is_clock_format_12h will crash the application.

### DIFF
--- a/lib/meson.build
+++ b/lib/meson.build
@@ -123,11 +123,17 @@ libgranite_dep = declare_dependency(
     include_directories: [include_directories('.')],
 )
 
+gsettings_desktop_schemas_dep = [
+    dependency('gsettings-desktop-schemas', version: '>= 3.27.1')
+]
+
+pc_deps = libgranite_deps + gsettings_desktop_schemas_dep
+
 # generate pkgconfig file
 granite_pc = pkgconfig.generate(
     libgranite,
     name: 'granite',
-    requires: libgranite_deps,
+    requires: pc_deps,
     subdirs: ['granite'],
     description: 'elementary\'s Application Framework',
     version: meson.project_version(),


tintou:
I don't think that it belongs here as it is not a build dependency, it's up to the packaging to provide the right run-time packages.

worldofpeace:
> I don't think that it belongs here as it is not a build dependency, it's up to the packaging to provide the right run-time packages.

The same could be said for a python project, where for example, pygobject3 is required at runtime, but all meson is going to do is install the project. It doesn't need this dependency in the build environment. However, without this, anyone using those methods are going to have an issue because there's no way to know their packaging is incorrect unless they read the source code (what I had to do and had a problem because of). So it's imperative for packagers to get a failure at build-time, because most likely, it's the same build expression that runs the build that will provide the runtime dependencies when installed. Not doing this will create a situation where end-users of granite will get bug reports because their app "crashes" with a missing settings schemas when using the `is_clock_format_12h` method. To prevent this using `Requires` is the most sensible option, this doesn't have to be packaging bug. It is true, it's not sensible to put it there because we don't link against it, but it isn't an uncommon practice.
For example, gnome-music is full of dependency objects https://gitlab.gnome.org/GNOME/gnome-music/-/blob/master/meson.build#L38, and these are completely unneeded for the build environment of the package, because they're all runtime dependencies on the typelibs.
That is their only way to keep track of these dependencies, and I know certainly if it wasn't there we'd see gnome-music packaged improperly in downstreams. Since granite is at the top of your dependency true, I think it's very important that people are packaging it correctly.

---

An argument could also be made that these methods without their proper runtime dependencies could not create a fatal error, and just present a warning.
Though it's of my personal opinion that this wouldn't be a great solution, as it's now a tradeoff of a degraded functionality in certain environments. (that aren't packaged correctly)

tintou:
No really, a pkg-config file isn't related to the packaging of a software. That's really something that is broken in the way people package software, they shoudn't be packaging blindly. If people have crashes outside of elementary then they should be able to debug it themselves, our role is to always make sure that the dependencies are there on our end.

PR #390, opened 2020-05-03
debian: add gsettings-desktop-schemas

worldofpeace:
Not sure if this is necessary for debian packaging.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -8,7 +8,8 @@ Build-Depends: debhelper (>= 10.5.1),
                libgirepository1.0-dev,
                libgtk-3-dev (>= 3.22),
                meson,
-               valac (>= 0.40)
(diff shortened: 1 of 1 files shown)


debian/control
```diff
@@ -8,7 +8,8 @@ Build-Depends: debhelper (>= 10.5.1),
                libgirepository1.0-dev,
                libgtk-3-dev (>= 3.22),
                meson,
-               valac (>= 0.40)
+               valac (>= 0.40),
+               gsettings-desktop-schemas (>= 3.27.1)
```
tintou, line 12: Thanks but it's not a Build-Depends but more a Depends as it is a run-time dependency. Also the GSettings schemas and keys we are using are there since at least 3.15

PR #399, opened 2020-05-27
Add new Granite.Avatar

danrabbit:
Screenshot from 2020-05-27 13 15 39@2x

This creates a new `Granite.Avatar` and deprecates the old `Granite.Widgets.Avatar`.

Use with https://github.com/elementary/stylesheet/pull/721

Major differences are:
* Two CSS Nodes `avatar` and `avatar-overlay`, which put shadows and highlights in the right place. Finally solving #398 
* Falls back to initials instead of a generic icon
* No internal margins so the widget is the actual size you requested, whether or not it is using the fallback
* Properties are all settable instead of construct-only
* No creation directly from a pixbuf

### DIFF
--- a/demo/Views/AvatarView.vala
+++ b/demo/Views/AvatarView.vala
@@ -20,24 +20,24 @@
 
 public class AvatarView : Gtk.Grid {
     construct {
-        var username = GLib.Environment.get_user_name ();
-        var iconfile = @"/var/lib/AccountsService/icons/$username";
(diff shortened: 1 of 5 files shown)


tintou:
I'd prefer not to bring Folks with Granite

danrabbit:
After some discussion in Slack, we're gonna try to keep the same API and just update the drawing for the old avatar

PR #410, opened 2020-06-14
meson: add introspection option

ffontaine:
This will allow the user to build granite without gobject-introspection

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


paledega:
good commit 👍🏻. we must merge

PR #422, opened 2020-08-28
Stop the timeout when hovering the toast

### LINKED ISSUE #321: Hovering toast should stop the timeout
If you go to perform the action on the toast and it times out while you're hovering, it hides out from under your cursor. Instead, hovering it should stop the timeout like we do for desktop notifications.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/80468295-hovering-toast-should-stop-the-timeout?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github).
</bountysource-plugin>

dar5hak:
Closes #321.

I added signal handlers so that when a mouse pointer enters the toast, the timeout is stopped and when it leaves, a new timeout is started.

While I was at it, I also abstracted the start and stop logic for timeouts into separate functions.

---

One edge case that I discovered and fixed was that if you:

1. close a timeout by clicking the close button or the action button
2. remove your mouse pointer from the toast
3. immediately retrigger the notification

then the old timeout started by the `leave_notify_event` will kick in and hide the toast prematurely. To fix this, I stopped the timeout unconditionally inside `send_notification ()`.

### DIFF
--- a/lib/Widgets/Toast.vala
+++ b/lib/Widgets/Toast.vala
@@ -130,7 +130,7 @@ namespace Granite.Widgets {
             timeout_id = GLib.Timeout.add (duration, () => {
                 reveal_child = false;
                 timeout_id = 0;
-                return false;
+                return GLib.Source.REMOVE;
             });
         }
 
--- a/lib/Widgets/Toast.vala
+++ b/lib/Widgets/Toast.vala
@@ -78,10 +78,7 @@ namespace Granite.Widgets {
             default_action_button.no_show_all = true;
             default_action_button.clicked.connect (() => {
                 reveal_child = false;
-                if (timeout_id != 0) {
-                    Source.remove (timeout_id);
-                    timeout_id = 0;
-                }
+                stop_timeout ();
                 default_action ();
             });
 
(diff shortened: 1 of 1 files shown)


lib/Widgets/Toast.vala
```diff
@@ -104,13 +98,49 @@ namespace Granite.Widgets {
...
+            if (default_action_button.visible) {
+                duration = 3500;
+            } else {
+                duration = 2000;
+            }
+
+            timeout_id = GLib.Timeout.add (duration, () => {
+                reveal_child = false;
+                timeout_id = 0;
+                return false;
```
tintou, line 133:
```suggestion
                return GLib.Source.REMOVE;
```
Just to make it more clear :)

tintou (changes requested):
Good thing! Just a tiny bit and it is ready to ship!

tintou (approved):

PR #423, opened 2020-08-31
Demo disable options 

paledega:
We can disable demo subdir building.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,3 +50,7 @@ The additional requirements for building the documentation are:
 To generate gtk-doc and valadoc documentation for this project, pass the
 additional `-Ddocumentation=true` flag to meson, and run `ninja` as before.
 
+## Demo
+
+To disable generate demo for this project, pass the additional `-Ddemo=false` 
+flag to meson,and run `ninja` as before.
--- a/meson.build
+++ b/meson.build
@@ -57,9 +57,12 @@ i18n = import('i18n')
 
 subdir('lib')
 subdir('data')
-subdir('demo')
 subdir('icons')
 subdir('po')
+
+if get_option('demo')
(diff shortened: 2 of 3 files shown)


Philip-Scott:
Hey, first of all thanks for your contribution! Is there a specific need or a problem that you are trying to solve by disabling the demo? It's not like it takes multiple minutes to compile the demo app, so I'm curious as to why this extra option is needed

paledega:
demo cannot compile with non english locale 
https://github.com/elementary/granite/issues/424
we have 2 way to fix this

1. change LANG variable 

2. disable demo

also developers may not need demo part

paledega:
```shell
[74/75] Generating Granite-1.0.typelib with a custom command lib/Granite-1.0.

Just/in/g-ir-compiler .-shared-library 1ib granite.so.5.5.0 --output lib/Granite-1.0.typelib /home/efel.cache/yay/granite-git/ ** (g-ir-compiler: 98480): ERROR *: 16:23:00.443: Invalid typelib for module "Granite': In directory (Context: Serviceslogleve [75/75] Linking target demo/granite-demo

d/lib/Granite-1.0.gir /home/efe/.cache/yay/granite-git/src/build/lib/Granite-1.0.gir:1895:1: warning: element field from state 7 is unknown, ignoring

value contains invalid characters: ınfo'

ninja: build stopped subcommand failed. HATA: build() içinde bir hata oluştu.

Çıkılıyor..: error making: X!s(func) string=0x55de058aeedo)


```

danrabbit:
This was fixed upstream in Vala: https://gitlab.gnome.org/GNOME/vala/-/issues/1067

Closing as "Out of scope"

PR #425, opened 2020-08-31
Utils: accel_to_string add Ctrl_L and Ctrl_R

danrabbit:
Currently these report as "Control L" and "Control R".

I guess we should decide whether it should be "Left Control" or "Left Ctrl"

Some context:

Screenshot from 2020-08-31 14 16 58@2x
Screenshot from 2020-08-31 14 16 44@2x


cassidyjames:
It looks like Control on its own is returned as `Ctrl`, so I think `Left Ctrl` and `Right Ctrl` would make sense.

PR #430, opened 2020-09-30
Constants: Add transition durations

### LINKED ISSUE #429: Transition duration constants
Something we've started doing in the stylesheet is to use varying animation durations for different kinds of movement. This is largely based on the Google Material guidelines: https://material.io/design/motion/speed.html

So far we use the following 5 durations:

`TRANSITION_DURATION_OPEN` = 250ms
`TRANSITION_DURATION_CLOSE` = 200ms

`TRANSITION_DURATION_EXPAND` = 300ms
`TRANSITION_DURATION_COLLAPSE` = 250ms

`TRANSITION_DURATION_ELEMENT` = 100ms

It could be helpful when animating using revealers, stacks, etc to provide these durations as constants in Granite

danrabbit:
Fixes #429 

I opted for only these two constants because it seems the material guidelines only recommend the longer transition durations for things like changing pages, open dialogs, etc. Dialog animation duration is not controlled by apps and the default transition duration for stack is already 300ms. I also left off the state/element transition since I can't think of any examples where we would use this currently outside the stylesheet. If it turns out we do want to use these other transition values I figure we can add them in a follow-up branch

### DIFF
--- a/data/granite.appdata.xml.in
+++ b/data/granite.appdata.xml.in
@@ -12,10 +12,10 @@
   <releases>
     <release version="5.5.1" date="2020-08-25" urgency="medium">
       <description>
-        <p>New Style Constants:</p>
+        <p>New Constants:</p>
         <ul>
-          <li>STYLE_CLASS_WARMTH for scales</li>
-          <li>STYLE_CLASS_TEMPERATURE for scales</li>
+          <li>STYLE_CLASS_WARMTH and STYLE_CLASS_TEMPERATURE for scales</li>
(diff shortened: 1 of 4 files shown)


cassidyjames (approved):
🎉

cassidyjames:
@danrabbit I guess we should mention these in the AppData?

tintou:
The issue with using a constant here is that changing the value in Granite won't be reflected until every consumer application will be rebuilt

cassidyjames (approved):

PR #441, opened 2020-11-14
Widgets: add validated entry

danrabbit:
This is something that I'm using in:
* Initial Setup
* Installer
* Switchboard User Accounts

and soon in Switchboard Online Accounts. We should also probably use this in AppCenter's payments dialog and/or Wallet. Basically anywhere that we do validation of forms we should be providing feedback about which inputs are valid

I didn't include the revealer with messages here because it's nice to be able to access Gtk.Entry API, but I'm open to discussion about if it makes more sense to do that here too.

Another thing not accounted for here was a way to set validation as a warning. For example in Password fields, we only care that the password is not blank and we provide a warning if it's not secure. I'm not sure if that means we should have an enumeration for validation state instead maybe?

Something else I considered was having something like `set_validation_func` instead of connecting to the changed signal yourself. Might still be a better idea, let me know your thoughts.

Another thing that might be cool, maybe for a future revision, is being able to add more validated entries (like Gtk.RadioButton) so that there would be an easy way to set an action widget as sensitive/sensitive when the validation of entries changes. Currently the way I'm doing that is after validating I call another function that says "if all of these are valid, make the button sensitive, otherwise insensitive". But it would be super convenient to do like:

```
var entry1 = new Granite.ValidatedEntry ();
var entry2 = new Granite.ValidatedEntry.from_widget (entry1);

var button = new Gtk.Button ();

entry1.bind_property ("group-is-valid", button, "sensitive");
```

Anyways, open to thoughts!


jeremypw:
Only had a quick look, but wouldn'y it be better to supply the validation regex to the ValidatedEntry constructor so that the validation can be done internally rather than connecting to the `changed` signal?

danrabbit:
@jeremypw Yeah that would probably be convenient to have a `from_regex` constructor or a regex property. In some cases though we're not using regex for validation but we're using a library like libpwquality or we're doing something like checking if a username is already in use on the system etc. We probably are using regex in enough places that this would be really convenient though, that's a great idea

demo/Views/FormView.vala
```diff
@@ -0,0 +1,52 @@
...
+        var button = new Gtk.Button.with_label ("Submit");
+
+        margin = 12;
+        orientation = Gtk.Orientation.VERTICAL;
+        row_spacing = 3;
+        halign = Gtk.Align.CENTER;
+        valign = Gtk.Align.CENTER;
+        vexpand = true;
+        add (username_label);
+        add (username_entry);
```
jeremypw, line 43: You haven't added the button.

lib/Widgets/ValidatedEntry.vala
```diff
@@ -0,0 +1,85 @@
...
+ */
+public class Granite.ValidatedEntry : Gtk.Entry {
+    /**
+     * Whether or not text is considered valid input
+     */
+    public bool is_valid { get; set; default = false; }
+
+    public ValidatedEntry.from_regex (Regex regex) {
+        if (regex == null) {
+            critical ("Null regex passed to ValidatedEntry.from_regex.  All input will be considered invalid.");
```
tintou, line 59: This can't happen as the method takes a `Regex` and not a `Regex?`
Also, this should be a debug, not a critical

lib/Widgets/ValidatedEntry.vala
```diff
@@ -0,0 +1,85 @@
...
+            });
+        }
+    }
+
+    construct {
+        activates_default = true;
+
+        unowned Gtk.StyleContext style_context = get_style_context ();
+
+        changed.connect_after (() => {
```
tintou, line 72: Better put `unowned Gtk.StyleContext style_context = get_style_context ();` inside the closure as Vala would increase a reference here.
Also, we're technically not sure the the StyleContext object remains always valid here.

lib/Widgets/ValidatedEntry.vala
jeremypw: Ah, OK - after inspecting the generated C I see that the constructor returns `null` if passed a null Regex so this extra test is not reached. I'll remove it.

lib/Widgets/ValidatedEntry.vala
jeremypw: Good point - will do.

jeremypw:
@danrabbit Are you happy with the changes I made?  Can this be approved now?

danrabbit:
@jeremypw I can't approve my own branch, but it looks good to me! Happy to merge if you're happy :)

jeremypw (approved):
Lets go!

PR #444, opened 2020-12-08
From regex constructor

jeremypw:
And also `.for_purpose (InputPurpose)` constructor with associated enum

### DIFF
--- a/demo/Views/FormView.vala
+++ b/demo/Views/FormView.vala
@@ -18,9 +18,8 @@
 */
 
 public class FormView : Gtk.Grid {
-    private Regex username_regex;
-
     construct {
+        Regex? username_regex = null;
         try {
             username_regex = new Regex ("^[a-z]+[a-z0-9]*$");
         } catch (Error e) {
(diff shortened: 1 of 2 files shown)


jeremypw:
The `.for_purpose()` constructor might be considered too much for this PR - it can be reverted if so.  Need advice re suitable regexes to supply as standard purposes.

danrabbit:
Hm yeah I'm not sure about the for purpose bit. This specific username regex is probably not necessarily the right regex for usernames on all services. Might be too specific and not generally useful

jeremypw:
OK, its gone.

jeremypw:
Do we need to cater for a null regex being passed to the `.from_regex ()` constructor?

danrabbit:
Oh yeah hm. I guess it would make it a lot simpler on the developer end to pass in the regex string and not have to pass in a regex object. If the regex is null, I guess we'd get a segfault trying to call `match` on it right? So maybe that's not good

jeremypw:
The problem with passing in a string is that if it is not a valid regex string then we cannot complete the construction as requested. It is easier for a developer to check an error thrown when constructing the Regex I think especially if constructing from a user input regex string. It is better for an obvious error (and a crash is pretty obvious) to be thrown during development so that it can be fixed. However, I guess we could fallback to a ValidatedEntry which does not actually work (because the changed signal will not be connected) and output a critical error to the terminal just to be on the safe side.

jeremypw:
Part of the thinking behind the `.for_purpose ()` constructor was, if a number of commonly used regexes can be identified, they can be made easily used for developers who (like me) are less comfortable with regex strings. However, that can be left for another PR if desired.

danrabbit (approved):
Okay yeah makes sense :)

PR #446, opened 2020-12-11
new markup tooltip with context description

andirsun:
Is neccesary for new tooltip indicators project, instead use traditional gtk markup with pango. 
Example : https://github.com/elementary/wingpanel-indicator-nightlight/pull/75#issuecomment-742953423

### DIFF
--- a/lib/Widgets/Utils.vala
+++ b/lib/Widgets/Utils.vala
@@ -213,9 +213,9 @@ public static string markup_accel_tooltip (string[]? accels, string? description
         if (unique_accels.length > 0) {
             ///TRANSLATORS: This is a delimiter that separates two keyboard shortcut labels like "⌘ + →, Control + A"
             var accel_label = string.joinv (_(", "), unique_accels);
-            
+
             var accel_markup = """<span weight="600" size="smaller" alpha="75%">%s</span>""".printf (accel_label);
-            arts += accel_markup;
+            parts += accel_markup;
         }
     }
 
--- a/lib/Widgets/Utils.vala
+++ b/lib/Widgets/Utils.vala
@@ -224,27 +224,27 @@ public static string markup_accel_tooltip (string[]? accels, string? description
 }
 
 /**
- * Takes a description and Subdescription
+ * Takes a context string and description and returns {@link Pango} markup for use in a {@link Gtk.Tooltip}.
  *
  * Example:
  *
  * Description
- * Sub-Description
+ * Context
  *
- * @param subdescription a standart tooltip text string
+ * @param context a standard tooltip text string
  *
  * @param description a standard tooltip text string
  *
- * @return markup with the description label on one line and a the subdescription on a new line
+ * @return markup with the description label on one line and a the context on a new line
  */
- public static string markup_tooltip_with_string (string? subdescription, string? description = null) {
+ public static string markup_contextual_tooltip (string? context, string? description = null) {
     string[] parts = {};
     if (description != null && description != "") {
         parts += description;
     }
 
-    if (subdescription != null && subdescription != "") {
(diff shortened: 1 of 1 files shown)


cassidyjames:
Would it make sense to have `markup_accel_tooltip` use this method as well so we're not repeating the Pango formatting?

andirsun:
> Would it make sense to have `markup_accel_tooltip` use this method as well so we're not repeating the Pango formatting?

Something like ```markup_accel_tooltip``` would be a 3rth parameter as ```string subdescription`` ?`

cassidyjames:
@andirsun no, I mean like `markup_accel_tooltip` could now use `markup_tooltip_with_string` to build its markup instead of manually building the same Pango markup string in two places.

andirsun:
> @andirsun no, I mean like `markup_accel_tooltip` could now use `markup_tooltip_with_string` to build its markup instead of manually building the same Pango markup string in two places.

Ohh, okay then, let me work in this change.

lib/Widgets/Utils.vala
```diff
@@ -223,6 +223,34 @@ public static string markup_accel_tooltip (string[]? accels, string? description
...
+    }
+
+    if (subdescription != null && subdescription != "") {
+        var secondary = "<small>%s</small>".printf (subdescription);
+        parts += secondary;
+    }
+
+    return string.joinv ("\n", parts);
+}
+
```
cassidyjames, line 253: Trying to think of better terminology here, plus a bit more fleshed out docs:

```suggestion
/**
 * Takes a context string and description and returns {@link Pango} markup for use in a {@link Gtk.Tooltip}.
 *
 * Example:
 *
 * Description
 * Context
 *
 * @param context a standard tooltip text string
 *
 * @param description a standard tooltip text string
 *
 * @return markup with the description label on one line and a the context on a new line
 */
 public static string markup_contextual_tooltip (string? context, string? description = null) {
    string[] parts = {};
    if (description != null && description != "") {
        parts += description;
    }

    if (context != null && context != "") {
        var secondary = "<small>%s</small>".printf (context);
        parts += secondary;
    }

    return string.joinv ("\n", parts);
}

```

danrabbit (changes requested):
This seems like overkill for this purpose tbh. For the purposes of `markup_accel_tooltip`, it accepts an array of strings and has to transform them. But I don't really see what advantage adding a whole method here has over just adding:

`public const string TOOLTIP_CONTEXT_MARKUP  """<span weight="600" size="smaller" alpha="75%">%s</span>""";`

you've also lost a pretty significant portion of the formatting here

andirsun:
> This seems like overkill for this purpose tbh. For the purposes of `markup_accel_tooltip`, it accepts an array of strings and has to transform them. But I don't really see what advantage adding a whole method here has over just adding:
> 
> `public const string TOOLTIP_CONTEXT_MARKUP """<span weight="600" size="smaller" alpha="75%">%s</span>""";`
> 
> you've also lost a pretty significant portion of the formatting here

Sure!, Im going to return as before.

cassidyjames:
@danrabbit derp, you're smart. Yeah, just exposing a public constant to use with `printf` syntax makes sense to me since there's no extra transforming or anything happening.

danrabbit:
Hey @andirsun I went ahead and made a separate PR since it's a little difficult to explain what I meant here: https://github.com/elementary/granite/pull/447

andirsun:
> Hey @andirsun I went ahead and made a separate PR since it's a little difficult to explain what I meant here: #447

What about this PR then ?

cassidyjames:
@andirsun sorry for all the back-and-forth on this. Dan's PR makes sense and we should use that constant in the indicators for the tooltip text. Since that PR supersedes this one, I'm closing this. Thanks again for your work on this!

PR #447, opened 2020-12-15
Utils: make tooltip secondary text markup public

danrabbit:
There's probably a better name for this, but this does the least to make this markup re-usable in other contexts. For example in indicators for middle-click functions and in the Files sidebar for storage devices

Supersedes #446


andirsun:
@danrabbit this is not equivalente to
```
<small>%s</small>
```

danrabbit:
@andirsun that's correct

cassidyjames (approved):
Yeah this is smart. Thanks!

PR #450, opened 2020-12-17
Remove deprecated About dialogs

cassidyjames:
Also removes AppMenu since it relied on AboutDialogs (and was deprecated at the same time, and only appears to be in use by abandoned projects across all of GitHub).


danrabbit (changes requested):
I think we should leave all the strings in Granite.Application for another PR. While it's easy to say that there are no apps using Granite.Widgets.About, there are over 300 results for Granite.Application so this is a lot more complicated to verify

lib/Application.vala
```diff
@@ -206,65 +206,5 @@ namespace Granite {
...
-
-        protected Granite.Widgets.AboutDialog about_dlg;
-
-        /**
-         * This method shows the about dialog of this app.
-         *
-         * @param parent This widget is the window that is calling the about page being created.
-         */
-        [Version (deprecated = true, deprecated_since = "0.4.2", replacement = "")]
-        public virtual void show_about (Gtk.Widget parent) {
```
danrabbit, line 234: I'm a little worried about removing this because it seems to be used as recently as 2 months ago: https://github.com/mirkobrombin/ElementaryPython

I think it would be better to leave it in but make it either do nothing or throw a https://valadoc.org/gtk+-3.0/Gtk.AboutDialog.html

danrabbit (changes requested):
afaict, AppMenu and AboutDialog haven't been used by anything in the last 4 years or more. But it does look like in recent history that at least one person has used `Granite.Application.show_about`.

I think probably the best solution there is to make it throw `Gtk.AboutDialog`. So we don't break anything but we can remove the thing we really care about

danrabbit (approved):
Alright looking into it more, this is probably not production code. So I'm less worried about it. Yolo

PR #459, opened 2021-01-11
Add SwitchModelButton

### LINKED ISSUE #458: Granite.SwitchModelButton
## Problem

We're using the pattern of a switch in a modelbutton quite a lot and we have to do it manually every time

## Proposal

Move this widget into Granite and save ourselves some time and ensure consistency

## Prior Art

https://github.com/elementary/terminal/blob/1eb6348193b5b572394bdf7c3a059d7c57ddb777/src/MainWindow.vala#L426

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/95946142-granite-switchmodelbutton?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #458 

This doesn't actually subclass ModelButton, so I wonder if it's better to call it `SwitchMenuItem`? Though it doesn't actually subclass menuitem either since it's more meant to be used with popovers

This supersedes [`Wingpanel.Widgets.Switch`](https://github.com/elementary/wingpanel/blob/master/lib/Widgets/Switch.vala) and provides the additional description text functionality as used in Nightlight indicator and Terminal. The reason for moving this to Granite is because non-indicators are using this pattern such as Code and Terminal

The additional padding below the header styled item is a Greenfield bug. I included that style here because we're using it in indicators


JoseExposito (changes requested):
The code looks nice, however, I'm not seeing the nice toggle animation I see in the indicators. Is it because I'm missing a dependency?

https://user-images.githubusercontent.com/1335948/107739479-3ff3af00-6d09-11eb-9f2a-5fc215897d04.mp4

danrabbit:
@JoseExposito nice catch! Should be fixed

JoseExposito (approved):
Ready! :rocket:

PR #460, opened 2021-01-11
Add Granite.Dialog

### LINKED ISSUE #443: Granite dialog should inherit Gtk.Window instead of Gtk.Dialog
If we want to stop setting the thing that gets GTK dialog buttons out of headerbars, we need to not subclass Gtk.Dialog.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/95031060-granite-dialog-should-inherit-gtk-window-instead-of-gtk-dialog?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #443 

The primary motivation for creating Granite.Dialog is the problems we're facing with action area in Gtk.Dialog. Specifically, we're currently setting `GTK_CSD=1` in the session to work around Gtk.Dialog using SSD when the action area is set to the bottom of the window. This is causing [issues](https://github.com/elementary/gala/issues/244) with some 3rd party apps.

It sounds like Gtk upstream is not interested in making Gtk.Dialog more complicated and that it is somewhat unmaintained. In fact, in a recent call its unclear that they plan to keep Gtk.Dialog at all.

The benefit for us to have a Dialog implementation is to set default styles, enforce the action area location, and make it easy to create dialog layouts.

This is separate from MessageDialog because we have other kinds of Dialogs like the Properties dialog in Files, settings dialogs in various apps, and other kinds of more complicated linear workflows that are not simple messages


lib/Widgets/Dialog.vala
```diff
@@ -0,0 +1,67 @@
...
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+* General Public License for more details.
+*
+* You should have received a copy of the GNU General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*/
+
+public class Granite.Dialog : Hdy.Window {
```
danrabbit, line 20: This is probably controversial and adds a dependency that we probably only want until Gtk4. Since we control the contents at the bottom of the window we don't strictly need this for rounded corners. What this does give us is a window with no titlebar area messing with our layout and lets us drag from anywhere in the window.

lib/Widgets/Dialog.vala
```diff
@@ -0,0 +1,67 @@
...
+    public signal void response (int response_id);
```
danrabbit, line 21: It's unclear that the whole response system is actually very beneficial. `add_button` returns a widget that is pretty easy to connect to signals ourselves. Is it worth duplicating this API?

lib/Widgets/Dialog.vala
```diff
@@ -0,0 +1,67 @@
...
+        };
+
+        action_area = new Gtk.ButtonBox (Gtk.Orientation.HORIZONTAL) {
+            layout_style = Gtk.ButtonBoxStyle.END
+        };
+        action_area.get_style_context ().add_class ("dialog-action-box");
+
+        var layout = new Gtk.Grid () {
+            margin = 12,
+            row_spacing = 24
```
danrabbit, line 44: I feel like it would probably be a lot better to actually set these values in the stylesheet so that we can scale them with text size

lib/Widgets/Dialog.vala
```diff
@@ -0,0 +1,67 @@
...
+        };
+        layout.attach (content_area, 0, 0);
+        layout.attach (action_area, 0, 1);
+
+        var window_handle = new Hdy.WindowHandle ();
+        window_handle.add (layout);
+
+        type_hint = Gdk.WindowTypeHint.DIALOG;
+        window_position = Gtk.WindowPosition.CENTER_ON_PARENT;
```
danrabbit, line 53: Are there other default properties that dialogs should have?

lib/Widgets/Dialog.vala
davidmhewitt: I guess my only opinion on that is that if we keep the API as similar as possible to the `Gtk.Dialog` API, then you just do find and replace `Gtk.Dialog` with `Granite.Dialog` when you want to port stuff.

lib/Widgets/Dialog.vala
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    construct {
+        content_area = new Gtk.Box (Gtk.Orientation.VERTICAL, 0) {
+            expand = true
+        };
+
+        action_area = new Gtk.ButtonBox (Gtk.Orientation.HORIZONTAL) {
+            layout_style = Gtk.ButtonBoxStyle.END,
+            spacing = 6
```
danrabbit, line 39: Making another note here that this could be done in CSS with something like:
```
.dialog-action-box button + button {
    margin-left: rem(6px)
}
```

lib/Widgets/Dialog.vala
```diff
@@ -0,0 +1,96 @@
...
+
+            return Gdk.EVENT_STOP;
+        });
+
+        action_area.add (widget);
+
+        return widget;
+    }
+
+    public int run () {
```
danrabbit, line 87: This is probably going to be the most complicated part to replicate. Docs here: https://valadoc.org/gtk+-3.0/Gtk.Dialog.run.html

lib/Widgets/Dialog.vala
```diff
@@ -0,0 +1,127 @@
...
+                return Gdk.EVENT_STOP;
+            }
+
+            return Gdk.EVENT_PROPAGATE;
+        });
+    }
+
+    public override bool key_press_event (Gdk.EventKey event) {
+        if (event.keyval == Gdk.Key.Escape) {
+            response (Gtk.ResponseType.CLOSE);
```
danrabbit, line 76: Is this the correct response type or should it be cancel?

lib/Widgets/Dialog.vala
```diff
@@ -0,0 +1,127 @@
...
+        // It's possible that this is called before action_area is constructed such as in Granite.MessageDialog
+        if (action_area == null) {
+            if (action_widgets == null) {
+                action_widgets = new List<Gtk.Widget> ();
+            }
+
+            action_widgets.append (widget);
+        } else {
+            action_area.add (widget);
+        }
```
danrabbit, line 113: Is there a better way to do this?

davidmhewitt:
@danrabbit I've implemented the `run` method. It'll probably need some testing to check it emits all the right response types, but seems to do the thing on the face of it.

lib/Widgets/Dialog.vala
```diff
@@ -0,0 +1,202 @@
...
+        return widget;
+    }
+
+    /**
+    * Blocks in a recursive main loop until {@link Granite.Dialog} either emits the {@link Granite.Dialog.response} signal, or is destroyed.
+    * If the dialog is destroyed during the call to run, run returns {@link Gtk.ResponseType.NONE}. Otherwise, it returns the response ID from the {@link Granite.Dialog.response} signal emission.
+    * Before entering the recursive main loop, run calls {@link Gtk.Widget.show_all} on the dialog for you.
+    * After run returns, you are responsible for hiding or destroying the dialog if you wish to do so.
+    */
+    public int run () {
```
danrabbit, line 173: It sounds like this is gone in Gtk4, so we should mark it as deprecated right away to help with the transition

lib/Widgets/Dialog.vala
davidmhewitt: Just doing some testing with `Gtk.Dialog` here and it seems the correct response would be `DELETE_EVENT`. However, it should be preceded by a `close` signal.

lib/Widgets/Dialog.vala
davidmhewitt: Agree :+1:

danrabbit (commented):
It looks like :
- [x] Files is using `get_widget_for_response` so that will need to be implemented.
- [x] And also there is an issue with the ownership of `add_button`. `Invalid assignment from owned expression to unowned variable`
- [x] It looks like Files is only calling `show` which causes dialogs to be empty. We may need to override this to show_all. Reproduceable with the delete permanently dialog for example

hanaral:
I'm not sure about what's alreaady planned but if this is seperate from MessageDialog you might want to add a demo to granite-demo

danrabbit:
@hanaral that's already covered in this branch

tintou:
@danrabbit you might want to check https://github.com/elementary/granite/tree/tintou/granite-dialog

davidmhewitt:
@tintou I thought one of the reasons for making our own dialog was because `get_action_area` is deprecated (and removed in GTK 4). Subclassing Gtk.Dialog doesn't fix that?

tintou:
Making our own dialog isn't related to GTK4, for GTK3 we can use `get_action_area` as long as `use_header_bar` is false

danrabbit:
@tintou that seems like it just means we have to do this twice and break granite.dialog API again when we move to Gtk4 no?

tintou:
@danrabbit I've looked at Gtk4 code and it would just continue to work, the content area still exists there, we just have to use another way to get it (which isn't difficult)

tintou:
Switching to Gtk4 is an API break anyway 😜 but that's not the point here

davidmhewitt:
Well, I just installed this branch, and the first thing I notice is that the polkit dialog no longer has a password field. So that's fun :laughing:

danrabbit:
@davidmhewitt I experienced that as well but then I rebuilt the polkit dialog and it was fine. Can you confirm that fixes it for you as well?

davidmhewitt (approved):
Code looks good and initial testing is going well. Seems like some projects already using Granite.MessageDialog will need a rebuild for things to work properly. So probably a good time for someone with admin rights to go round and version bump everything to granite-6 once this is built in the PPA.

PR #461, opened 2021-01-13
SettingsPage Header Icon: Standardize Sizing

lainsce:
They are now following the 16, 32, 64, 128 icon size progression.

Looks like:
v


lainsce:
48px is needed, closing.

PR #474, opened 2021-03-05
Bind active property bidirectionally

### LINKED ISSUE #473: SwitchModelButton does not allow active property to be binded
## What Happened

Honestly I'm not really sure about the title. If someone could suggest a better one please do.

Using `GLib.Settings.bind (key, Granite.SwitchModelButton, "active", GLib.SettingsBindFlags.DEFAULT)` does not actually change the key when the `active` property is changed.

## Expected Behavior

`key` should be binded to `Granite.SwitchModelButton.active`.

## Steps to Reproduce

1. Bind Granite.SwitchModelButton to a key.
2. Flip the switch.
3. Observe if key's value has changed.

## Platform Information
Granite master

pongloongyeat:
Fixes #473

### DIFF
--- a/lib/Widgets/SwitchModelButton.vala
+++ b/lib/Widgets/SwitchModelButton.vala
@@ -83,7 +83,7 @@ public class Granite.SwitchModelButton : Gtk.ToggleButton {
 
         bind_property ("text", label, "label");
         bind_property ("description", description_label, "label");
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
Big ooops. Thanks for fixing!

PR #479, opened 2021-03-12
Demo: Replace Granite.Avatar with Gtk.Spinner

danrabbit:
Get rid of deprecated Granite.Avatar but without adding a Hdy dep to the demo


hanaral:
You better round those corners while ur at it then

cassidyjames:
@hanaral 

>**without** adding a Hdy dep

cassidyjames:
Derp wrong button

hanaral:
> @hanaral 
> 
> 
> 
> >**without** adding a Hdy dep

It is too late for me to be on gh rn 😑

cassidyjames (approved):

PR #483, opened 2021-04-10
Get prefers accent color from AccountsService

meisenzahl:
Could be handy for

- https://github.com/elementary/gala/pull/1104
- https://github.com/elementary/switchboard-plug-pantheon-shell/pull/263

### DIFF
--- a/lib/Widgets/Settings.vala
+++ b/lib/Widgets/Settings.vala
@@ -2,6 +2,7 @@
 namespace Granite {
     [DBus (name = "io.elementary.pantheon.AccountsService")]
     private interface Pantheon.AccountsService : Object {
+        public abstract int prefers_accent_color { owned get; set; }
         public abstract int prefers_color_scheme { owned get; set; }
     }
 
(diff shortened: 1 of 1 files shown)


lib/Widgets/Settings.vala
```diff
@@ -88,6 +155,28 @@ namespace Granite {
...
+                prefers_accent_color = (AccentColor) pantheon_act.prefers_accent_color;
+
+                ((GLib.DBusProxy) pantheon_act).g_properties_changed.connect ((changed, invalid) => {
+                    var accent_color = changed.lookup_value ("PrefersAccentColor", new VariantType ("i"));
+                    if (accent_color != null) {
+                        prefers_accent_color = (AccentColor) accent_color.get_int32 ();
+                    }
+                });
+            } catch (Error e) {
+                critical (e.message);
```
JoseExposito, line 176: It is very unlikely, but if `GLib.Bus.get_proxy_sync` fails, `prefers_accent_color` will return `null`.

I'm not sure which accent color is going to be the default one, but maybe we should fallback to it on error when `_prefers_accent_color == null`.

lib/Widgets/Settings.vala
```diff
@@ -14,6 +15,56 @@ namespace Granite {
      * Granite.Settings provides a way to share Pantheon desktop settings with applications.
      */
     public class Settings : Object {
+        /**
+         * Possible accent color preferences expressed by the user
+         */
+         public enum AccentColor {
```
JoseExposito, line 21: I'm fine with this names, they are easy to remember so, this is more of a question. Do we want to use `BLUE` instead of `BLUEBERRY`? The same applies for lime, banana, bubblegum, etc.

Let's see what the others think about it

JoseExposito (changes requested):
This looks great :tada: Just a question about the name of the colors in the API and a paranoid check.

lib/Widgets/Settings.vala
meisenzahl: @cassidyjames has already thought about this 😀️

> @meisenzahl io.elementary.AccentColor seems fine to me. Could we ensure we're matching this mapping, though? I think we're also generally moving to only use the cutesy food names in user strings and just use more standard color names anywhere for developers.
> 
> [elementary/default-settings@17d737e](https://github.com/elementary/default-settings/commit/17d737ecc1fdd083127cbc290c49e3ad0563030b)

lib/Widgets/Settings.vala
meisenzahl: I'm not quite sure. I simply copied and modified the code for `ColorScheme` 🤓️

lib/Widgets/Settings.vala
JoseExposito: Cool, makes sense :+1:

cassidyjames (approved):

PR #486, opened 2021-04-20
Utils: add force_elementary_style ()

cassidyjames:
Turns https://github.com/elementary/calculator/pull/177/ into a one-liner.

### DIFF
--- a/lib/Widgets/Utils.vala
+++ b/lib/Widgets/Utils.vala
@@ -291,22 +291,18 @@ public static Gdk.RGBA contrasting_foreground_color (Gdk.RGBA bg_color) {
 }
 
 /**
- * Sets the app's icon theme and stylesheet to elementary defaults. Optionally takes a specific stylesheet accent variant as a string.
- *
- * @param accent optional stylesheet accent variant to use, e.g. "blueberry"
+ * Sets the app's icon theme and stylesheet to elementary defaults.
  */
-public void force_elementary_style (string accent = "blueberry") {
+public void force_elementary_style () {
     const string STYLESHEET_PREFIX = "io.elementary.stylesheet.";
+
     var gtk_settings = Gtk.Settings.get_default ();
     var current_stylesheet = gtk_settings.gtk_theme_name;
 
     gtk_settings.gtk_icon_theme_name = "elementary";
 
-    if (
-        !current_stylesheet.has_prefix (STYLESHEET_PREFIX) ||
-        current_stylesheet.replace (STYLESHEET_PREFIX, "") != accent
-    ) {
-        gtk_settings.gtk_theme_name = STYLESHEET_PREFIX + accent;
+    if (!current_stylesheet.has_prefix (STYLESHEET_PREFIX)) {
+        gtk_settings.gtk_theme_name = STYLESHEET_PREFIX + "blueberry";
     }
 }
 
--- a/lib/Widgets/Utils.vala
+++ b/lib/Widgets/Utils.vala
@@ -296,7 +296,7 @@ public static Gdk.RGBA contrasting_foreground_color (Gdk.RGBA bg_color) {
 public void force_elementary_style () {
     const string STYLESHEET_PREFIX = "io.elementary.stylesheet";
 
-    var gtk_settings = Gtk.Settings.get_default ();
+    unowned var gtk_settings = Gtk.Settings.get_default ();
     var current_stylesheet = gtk_settings.gtk_theme_name;
 
     gtk_settings.gtk_cursor_theme_name = "elementary";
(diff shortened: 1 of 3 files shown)


danrabbit:
I suppose maybe this should be "force" instead of just "use"?

cassidyjames:
@danrabbit yeah that's probably fair

cassidyjames:
I'm also unsure if I should handle the variant thing here or ignore if for now; I figured it _might_ be nice to consider forcing a specific variant, but if that's too much scope creep, I'm fine dropping it.

Marukesu:
Maybe we want to manage the notify signal here? with elementary/calculator#177 if i start the program with the elementary stylesheet and change to other theme they will fallback to Adwaita. If i start with Adwaita, the theme is frozen on io.elementary.blueberry ever if i set the theme to any stylesheet theme.

lib/Widgets/Utils.vala
```diff
@@ -290,6 +290,26 @@ public static Gdk.RGBA contrasting_foreground_color (Gdk.RGBA bg_color) {
     return fg_color;
 }
 
+/**
+ * Sets the app's icon theme and stylesheet to elementary defaults. Optionally takes a specific stylesheet accent variant as a string.
+ *
+ * @param accent optional stylesheet accent variant to use, e.g. "blueberry"
+ */
+public void force_elementary_style (string? accent = "blueberry") {
```
Marukesu, line 298: No need to be nullable here since we are defaulting to blueberry if empty and if someone set it to `null`, we will set the theme to `io.elementary.stylesheet.`.

cassidyjames:
> Maybe we want to manage the notify signal here? with [elementary/calculator#177](https://github.com/elementary/calculator/pull/177) if i start the program with the elementary stylesheet and change to other theme they will fallback to Adwaita. If i start with Adwaita, the theme is frozen on io.elementary.blueberry ever if i set the theme to any stylesheet theme.

I thought about that but honestly supporting changing to/from/between non-elementary stylesheets on the fly is much less of a concern since that's already pretty janky in Flatpak apps to begin with; I mainly want to support the case where you are on elementary OS or at least using the elementary stylesheet (so we don't change anything and just use the stylesheet as-is) or are not on elementary stylesheet so we just force a specific one. I think this does the least amount of work to get that working satisfactorily.

lib/Widgets/Utils.vala
cassidyjames: D'oh, you're right. Thanks!

cassidyjames:
I do wonder if it makes sense to check this against a list of valid variants instead of just allowing any arbitrary string; maybe we accept a `Granite.Settings.AccentColor` instead of a string?

Marukesu:
> I do wonder if it makes sense to check this against a list of valid variants instead of just allowing any arbitrary string; maybe we accept a `Granite.Settings.AccentColor` instead of a string? Again, I'm also happy to split that off into its own branch if it is creeping too much scope here. Maybe we just start without any params.

I don't think setting a variant here is necessary now, apps that use fixed accent are already doing it with `Gtk.StyleContext` and a custom css file.

cassidyjames:
@Marukesu yep that was my conclusion as well; I've removed that from this branch.

Exalm:
If it's "force", I'd expect it to also track changes and not be a one-time thing. :)

cassidyjames:
@Exalm I mean, see the previous comments in that vein. I was going to prefer `use` but realized that on elementary OS, that's already the assumption, so `force` or something does make sense. I'm open to suggestions or improvements to track changes here but honestly I don't want to turn this into a bikeshed. Further improvements can follow in future PRs.

lib/Widgets/Utils.vala
```diff
@@ -290,6 +290,23 @@ public static Gdk.RGBA contrasting_foreground_color (Gdk.RGBA bg_color) {
     return fg_color;
 }
 
+/**
+ * Sets the app's icons, cursors, and stylesheet to elementary defaults.
+ */
+public void force_elementary_style () {
+    const string STYLESHEET_PREFIX = "io.elementary.stylesheet";
+
+    var gtk_settings = Gtk.Settings.get_default ();
```
danrabbit, line 299: Should this be an unowned ref?
```suggestion
    unowned var gtk_settings = Gtk.Settings.get_default ();
```

lib/Widgets/Utils.vala
```diff
@@ -290,6 +290,23 @@ public static Gdk.RGBA contrasting_foreground_color (Gdk.RGBA bg_color) {
...
+    var current_stylesheet = gtk_settings.gtk_theme_name;
```
danrabbit, line 300: Do we need to cache this since we only use it once?

Exalm:
Not bikeshed at all, what I mean is it can track changes as well instead of being a one-time thing. :)

lib/Widgets/Utils.vala
cassidyjames: @danrabbit ah yeah I think I used this when I was planning to do more but now it's a one-off. :+1:

danrabbit (approved):
Code looks good and works as intended afaict. Nice job :)

PR #487, opened 2021-04-20
Release 6.1.0

cassidyjames:
Additions:

- TRANSITION_DURATION_IN_PLACE for consistent in-place widget transformations

Other Changes:

- `accel_to_string` handles accel markup without modifiers or that are only modifiers
- Updated translations


danrabbit:
We need to wait on this a bit. There's an alternative proposal for the runtime which would make this unnecessary

danrabbit (approved):
Let's fuckin' go :rocket:

PR #490, opened 2021-04-29
Build on windows

albfan:
Add support for windows.

Working on mingw to run Akira:

https://github.com/msys2/MINGW-packages/pull/8505/files#diff-18b1226b74e8b4adf04b5e3401ac6886ead2da9a1dc26b6c61be29141ca0f641R23

### DIFF
--- a/lib/Application.vala
+++ b/lib/Application.vala
@@ -154,8 +154,10 @@ namespace Granite {
             Granite.Services.Logger.initialize (program_name);
             Granite.Services.Logger.DisplayLevel = Granite.Services.LogLevel.INFO;
             message ("%s version: %s", program_name, build_version);
+#if !WINDOWS
             var un = Posix.utsname ();
             message ("Kernel version: %s", (string) un.release);
+#endif
             Granite.Services.Logger.DisplayLevel = Granite.Services.LogLevel.WARN;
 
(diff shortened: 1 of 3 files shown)


lib/Application.vala
```diff
@@ -154,8 +154,10 @@ namespace Granite {
             Granite.Services.Logger.initialize (program_name);
             Granite.Services.Logger.DisplayLevel = Granite.Services.LogLevel.INFO;
             message ("%s version: %s", program_name, build_version);
+#if ! WINDOWS
```
tintou, line 157:
```suggestion
#if !WINDOWS
```

lib/Services/System.vala
```diff
@@ -144,9 +144,12 @@ namespace Granite.Services {
                 return;
 
             AppInfo info;
+
+#if ! WINDOWS
```
tintou, line 148:
```suggestion
#if !WINDOWS
```

lib/Application.vala
albfan: 😉

lib/Services/System.vala
albfan: Done

tintou:
It looks good to me, Although I haven't tested it. I don't see why we couldn't merge this as it's painless for us

tintou (approved):

Issue #492, opened 2021-05-01
Increase required valac version?


peteruithoven:
## What Happened

Can't compile fresh clone of master branch of Granite on elementary OS 5. I get these errors:
```
../lib/Widgets/SwitchModelButton.vala:65.9-65.19: error: The type name `var' could not be found
        unowned var description_style_context = description_label.get_style_context ();
        ^^^^^^^^^^^
../lib/Widgets/Utils.vala:298.5-298.15: error: The type name `var' could not be found
    unowned var gtk_settings = Gtk.Settings.get_default ();
    ^^^^^^^^^^^
```

While I got no warnings from meson's setup:
```
meson build --prefix=/usr
The Meson build system
Version: 0.49.2
Source dir: /home/peteruithoven/Projects/elementary/granite
Build dir: /home/peteruithoven/Projects/elementary/granite/build
Build type: native build
Project name: granite
Project version: 6.1.0
Native C compiler: cc (gcc 7.5.0 "cc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0")
Native Vala compiler: valac (valac 0.40.25)
Build machine cpu family: x86_64
Build machine cpu: x86_64
Found pkg-config: /usr/bin/pkg-config (0.29.1)
Dependency gio-unix-2.0 found: YES 2.56.4
Dependency gee-0.8 found: YES 0.20.2
Dependency gio-2.0 found: YES 2.56.4
Dependency glib-2.0 found: YES 2.56.4
Dependency gobject-2.0 found: YES 2.56.4
Dependency gtk+-3.0 found: YES 3.22.30
Configuring Config.vala using configuration
Library m found: YES
Library posix found: YES
Program g-ir-compiler found: YES (/usr/bin/g-ir-compiler)
Program /home/peteruithoven/Projects/elementary/granite/meson/post_install.py found: YES (/home/peteruithoven/Projects/elementary/granite/meson/post_install.py)
Build targets in project: 10
Found ninja-1.8.2 at /usr/bin/ninja
```

## Expected Behavior

I'd get a clearer warning that my current setup isn't supported. 

According to @jeremypw the meson.build needs to be updated:
https://github.com/elementary/granite/pull/236#issuecomment-830609727
Probably specifically the valac version: 
https://github.com/elementary/granite/blob/master/meson.build#L8

[...]

PR #493, opened 2021-05-01
Increased required vala compiler version

peteruithoven:
Fixes: #492

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -5,8 +5,8 @@ project(
     version: '6.1.0'
 )
 
-if meson.get_compiler('vala').version().version_compare('<0.40.0')
-    error('vala compiler version 0.40.0 or newer is required.')
+if meson.get_compiler('vala').version().version_compare('<0.48.0')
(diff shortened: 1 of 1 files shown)


jeremypw (approved):
Did it fix your problem building on Hera? 

Anyway it needs to be done.

peteruithoven:
As far as I know I can't easily install a newer version of the Vala compiler, but now at least it gives a clear warning:
```
meson.build:9:4: ERROR:  Problem encountered: vala compiler version 0.48.0 or newer is required.
```

jeremypw:
You have to install a ppa to get the later versions (at least, that is what I did). See https://launchpad.net/~vala-team/+archive/ubuntu/next

peteruithoven:
Ah good point! (I've been mostly doing web front-end development, so I'm a bit rusty)

I'm not able to install the latest version of the vala compiler from that ppa, but github isn't a good place for such questions so I opened another stack exchange question: 
https://elementaryos.stackexchange.com/questions/27859/how-to-install-newer-vala-compiler

PR #496, opened 2021-05-03
Revert "Get prefers accent color from AccountsService"

danrabbit:
Reverts elementary/granite#483

This seems only interesting to two very specific applications and isn't a good long-term solution for flatpak'd apps, as we don't really want them to be adding accountsservice to their sandbox.

### DIFF
--- a/data/granite.appdata.xml.in
+++ b/data/granite.appdata.xml.in
@@ -14,7 +14,6 @@
       <description>
         <p>Additions:</p>
         <ul>
-          <li>Settings.prefers_accent_color to get a user's accent color preference</li>
           <li>force_elementary_style to set the app's icons, cursors, and stylesheet to elementary defaults</li>
         </ul>
         <p>Other Changes:</p>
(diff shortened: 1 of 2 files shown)


cassidyjames (approved):
Yeah we talked about this on the standup call today. The original intent was in case we needed this for Flatpak apps to request the stylesheet themselves, but the stylesheet is correctly forwarded to the apps in Flatpak so there's no reason for this.

I also removed it from the AppData.

Those implementations (Gala, Desktop plug, and Onboarding) should talk to AccountsService directly, right? And do we want to update them before reverting this?

meisenzahl:
This code is used in Onboarding and Desktop plug. I can refactor this tomorrow if you like.

meisenzahl:
Done

- https://github.com/elementary/switchboard-plug-pantheon-shell/pull/263
  - https://github.com/elementary/switchboard-plug-pantheon-shell/pull/263/commits/781868811f0939bf861661472f2aa554befdddf4
- https://github.com/elementary/onboarding/pull/130

PR #501, opened 2021-05-14
Revert adding force_elementary_style

cassidyjames:
@tintou requested we revert this and instead handle it in the Flatpak platform, which makes sense.

### DIFF
--- a/data/granite.appdata.xml.in
+++ b/data/granite.appdata.xml.in
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Copyright 2019–2020 elementary, Inc. <contact@elementary.io> -->
+<!-- Copyright 2019–2021 elementary, Inc. <contact@elementary.io> -->
 <component>
   <id>io.elementary.granite</id>
   <metadata_license>CC0-1.0</metadata_license>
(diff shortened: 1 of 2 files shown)


danrabbit (approved):

avojak:
@cassidyjames What's the recommended way to accomplish this currently?

"handle it in the Flatpak platform" makes me think it should happen automatically for apps that are built against the elementary Flatpak platform?

cassidyjames:
I'll have to double check, but at the moment you can just manually set the stylesheet, and it will be available in the platform. I'd recommend something like this:

```vala
var gtk_settings = Gtk.Settings.get_default ();
gtk_settings.gtk_icon_theme_name = "elementary";
gtk_settings.gtk_theme_name = "io.elementary.stylesheet.blueberry";
```

You could also wrap that in a check if the current stylesheet starts with `io.elementary.stylesheet.` and only manually set it if it's false; that way you retain the accent colors on elementary OS.

avojak:
Awesome, yea I tested that last night by just copying the code used for force_elementary_style into my app and it worked great, but since there was a reason for it being removed from Granite I didn't necessarily want to go that route if there was a better way. Thanks!

PR #502, opened 2021-05-20
Add API for custom AccelLabels in the tab context menu

andreasfelix:
Related to https://github.com/elementary/terminal/pull/580

This PR makes it possible to pass custom `AccelLabel`s to the `Granite.DynamicNotebook` and `Granite.Tab`. I never did such an API change before, so I am not sure if this is the most elegant way to achieve this.

I also refactored the variable names of instances of `Gtk.MenuItem` to more explicit names so they are consistent with the rest of the codebase:

`new_window_m` ---> `new_window_menuitem`

#### Example accellabels for terminal and files:

image

image


jeremypw:
One way of making it compatible with existing apps would be to use new constructors, not change the signature of the old one.
i.e. new DynamicNotebook.with_accels (...) for example.

andreasfelix:
* I didn't implement this because, I thought it wouldn't be relevant and would make the demo code more complex than necessary. But I can implement them if you want or remove the accellabels from the demo.
*
* Yes, I thought it would keep the code simpler if the accellabels can't change. Another option would to just expose `Gtk.Menu` publicly and add an option which removes the default `MenuItem`s, so that the application can create its own menuitems instead.
* I provided default arguments in the constructor so the elementary apps shouldn't need to be changed to be compatible. If you recompile the elementary apps with the this version of Granite installed, they shouldn't segfault? Right? If I am not mistaken this is API vs ABI? But if it is required that the apps work without recompiling, I will add an additional constructor.

jeremypw:
* To keep the PR small maybe it would be better not to change the demo. That can be changed in another PR if someone feels moved to do so.
* I would be happier if the PR did not require recompilation of client apps.

@peteruithoven Would you like to join the discussion as to the relative merits of your approaches?

andreasfelix:
@jeremypw The last commit adds two new named constructors `DynamicNotebook.with_accellabels` and `Tab.with_accellabels`. On my machine, the existing elementary apps now work without recompiling. But this change makes the code somewhat redundant, because the parameters are optional anyway. So these new constructors don't add any new functionality and only make sure to not break existing binaries.

andreasfelix:
I also updated elementary/terminal#582 to use the new named constructors.

andreasfelix:
Also rename `.._menuitem` back to `..._m` to keep the diff smaller.

jeremypw:
> But this change makes the code somewhat redundant, because the parameters are optional anyway. So these new constructors don't add any new functionality and only make sure to not break existing binaries.

As this is a public library used outside elementary I think it best to be cautious.  You could always mark the old constructor as "deprecated" I guess.

jeremypw:
@tintou Is it OK to change the ABI of the original constructor like this?

andreasfelix:
Should I also add options for the `close_others`, `close_other_right` and `pin/unpin` menu items?

tintou:
@jeremypw There is technically no ABI/API change here, so it's alright for me :)
Also, the feature looks good!

tintou (approved):

jeremypw:
@tintou Thanks! Maybe I was over-cautious :stuck_out_tongue:

PR #503, opened 2021-06-03
Fix accel_to_string to not automatically add a '+' character unless needed

zeebok:
This fixes the accel_to_string function so that it does not always add " + " to any accel passed to it. This is created to address a comment on [this pull request](https://github.com/elementary/applications-menu/pull/480) and includes an update to the demo to include both an accel without any mod and an accel that is only a mod.


danrabbit (approved):
Very nice. Thanks for updating the demo as well!

danrabbit (approved):

PR #505, opened 2021-06-08
ModeButton: allow keyboard focus

### LINKED ISSUE #504: ModeButtons are not operable with keyboard
Most details copied from elementary/switchboard-plug-pantheon-shell#278.

## What Happened

I am unable to select ModeButton sets with the keyboard. For example, in the Appearance panel, I cannot operate the style schedule or text size selectors. I can click on them or tab around the window, but in both cases the focus skips over them. The only selectable items seem to be the style modes, accent color, dyslexia font switch, and back button.

## Expected Behavior

Both button sets should be selectable by keyboard.

## Steps to Reproduce

In the Desktop settings pane:
1. Select the currently active theme radio button (you can use the left/right arrow keys to demonstrate this)
2. Press tab to select the next widget
3. Focus will move to the accent color (you can use the left/right arrow keys to demonstrate this)
4. Continue tabbing around: the only widgets that select are the ones I stated. Eventually, you will cycle back to the style selector without ever selecting the ModeButtons.

### Logs

n/a (I think)

## Platform Information

Odin daily, up to date as of 6/4/21

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/99013224-modebuttons-are-not-operable-with-keyboard?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60830613&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #504 

No idea why this was set to disallow focus. I can't seem to find any regressions allowing the widgets to be focused

### DIFF
--- a/lib/Widgets/ModeButton.vala
+++ b/lib/Widgets/ModeButton.vala
@@ -34,7 +34,6 @@ namespace Granite.Widgets {
             public int index { get; construct; }
             public Item (int index) {
                 Object (index: index);
-                can_focus = false;
                 add_events (Gdk.EventMask.SCROLL_MASK);
             }
         }
@@ -71,7 +70,6 @@ namespace Granite.Widgets {
         construct {
             homogeneous = true;
             spacing = 0;
-            can_focus = false;
 
             item_map = new Gee.HashMap<int, Item> ();
 


JoseExposito (approved):
The only "problem" this generates is that the focus glow now is visible until the element looses the focus:

Screenshot from 2021-06-09 07 39 52

But this is really an a11y feature, so I'm 100% in, nice job!

jeremypw (approved):
Yes, can't see any reason not to have the mode-button focusable.

PR #507, opened 2021-06-24
HyperTextView

marbetschar:
This PR adds  `Granite.Widgets.HyperTextView` , a drop in replacement for `Gtk.TextView` with added support for navigatable URI's. The code is mainly based on the [example of Alberto Fanjul](https://gitlab.gnome.org/albfan/textviewhyperlinks):

Granite HyperTextView

Although it is not heavily tested yet, it _should_ work for the majority of cases. Also I guess there is room for performance improvements (see discussion in https://github.com/elementary/calendar/pull/656).

This feature request came up in Calendar (https://github.com/elementary/calendar/issues/571), but would also be useful in descriptions for Tasks and potentially further cases.

Last but not least, this PR is a response to @cassidyjames request on Twitter: https://twitter.com/CassidyJames/status/1404848863395991558


mcclurgm:
I like this idea, although I haven't had the chance to test it yet. This definitely seems like a good thing to implement in Granite imo, since it seems like a pain to write but useful for a wide variety of apps.

I think doing something like pulling up a long blog post's source code on weaker hardware like a Pi or Pinebook would be a good stress test for performance, but I don't have that hardware. I'll try pulling up a long editor sometime today if I have the time, but my computer's pretty fast, so I expect it can handle this pretty well.

A few thoughts in general:

- How generic do we want the URI regex to be? See my comment in elementary/calendar#656, you can get very in the weeds with trying to detect a generic URL: https://mathiasbynens.be/demo/url-regex
- One random thought about performance: is there a way to restrict the regex search to only a part of the buffer? If so, we could restrict it to only an area that we know has been modified, since that's the only place that a previously undetected link could appear. AFAIK it's impossible to have a line break in a URI, so we could search only in the paragraph that was edited, for example. This would require some special casing for pastes and such that can edit more than 1 paragraph at a time. I'm not familiar with Gtk text widgets to say how we could handle that. But if this is possible to implement cleanly, it should let us parse arbitrarily long text buffers, as long as people don't go crazy with making long paragraphs. There may be other ways to divide up the buffer too, but that's the first that came to mind.
- Like I said I haven't tested this, but does it handle deleting part of a link, to the point where it becomes invalid? That should make the link markup disappear. (This is more a reminder to myself to test.)
- Do we want to make the markup style configurable? IE, specify color and maybe bold/italic/underline? That would definitely increase the scope though.

marbetschar:
@mcclurgm I do agree with the ideas on how to improve performance, but I'm wondering if the effort is really worth it. So having someone doing some real world testing on restricted hardware would be great.

Regarding the link detection; The regex in place right now is rather simple. Leading and trailing whitespace mark the start and end of the URI. In between there must be at least four non-whitespace characters with a dot in the middle (2 characters, 1 dot, 2 characters). The protocol information at the beginning (xyz://) is optional and HyperTextView will use http:// if there isn't any protocol.

I'm aware this kind of regex will lead to some false positives (ironically "Gtk.TextView" is a false positive)  - but its simple and I can't think of anything it does not work for. IMHO this outweights the false positives ;)

marbetschar:
Here's the result of running the [examples](https://mathiasbynens.be/demo/url-regex) against the regex currently in place: https://regex101.com/r/bmX4TC/1

marbetschar:
@mcclurgm adressed the performance concerns: Rescanning of the buffer is now limited from start to end of line of the cursor position. Also buffer change events are debounced. That said, scanning of said substring happens once each 300ms at most. The only exception is when text was pasted - in this case a rescan of the whole buffer is forced after 300ms.

This strategy seems to work quite well on my end - even for large texts. As test case I copied the source of https://blog.elementary.io/elementary-os-6-odin-beta-2/ (around 1309 lines of code) and editing feels quite well on my machine.

mcclurgm:
I'm seeing some strange behaviors here. First, it seems like there's some issue with offset. Depending on what I paste, the hypertext underlines start after the URI does, and seem to extend past the end. Here's an example:
```
/*
 *  Copyright (C) 2019–2020 elementary, Inc. (https://elementary.io)
 *                2011–2013 Maxwell Barvian <maxwell@elementaryos.org>
 *
 *  This program or library is free software; you can redistribute it
 *  and/or modify it under the terms of the GNU Lesser General Public
 *  License as published by the Free Software Foundation; either
 *  version 3 of the License, or (at your option) any later version.
```
On my machine, I see that the link underlines start after the `ht` and `maxw` text. Depending what I select, the offset varies. For example, just pasting 
```
Copyright (C) 2019–2020 elementary, Inc. (https://elementary.io)
 *                2011–2013 Maxwell Barvian <maxwell@elementaryos.org>
```
the link underlines start after `https:` and `maxwell@`. They also extend past the end of the URI: in the second example, the elementary.io link extends a few characters into the next line (it ends a few spaces after the `*`). As a counterexample, this text behaves perfectly normally when pasted:
```
Someone is inviting you to a scheduled Zoom meeting.

Join Zoom Meeting
https://abcd.zoom.us/j/0123456

Meeting ID: 904 938 668

One tap mobile
+10302023418# US (San Jose)
+12343210,,# US (San Jose)
(https://elementary.io)

Dial by your location
        +1 832 120 9417 US (San Jose)
Find your local number: https://abcd.zoom.us/u/jfdklsajfds
```

I see some strange behavior with links surrounding the insertion point blinking in and out when editing by hand, when I was looking at the offset issue. I've attached a video. Sorry about the artifacts. Focus on the `elementary.io` links on either side of the cursor.
https://user-images.githubusercontent.com/29022469/123734185-1b344a00-d85a-11eb-9a35-53470518ce96.mp4


Then another (maybe) unrelated bug. The regex (sometimes?) doesn't match at the very end of a buffer. For example, if I paste the second example string (the shortened copyright header) at the end of the file, neither URI matches. If I add a line break after that same text, the link will sometimes match. But sometimes I paste a long buffer with a link at the end and it works ok? This is a weird one too. I haven't found any indication of when it works or doesn't.

Also, some things that aren't bugs but I wanted to bring up:
- When you type at the end of a link, it extends as you type. This looks a little odd until the debouncing catches up and the underline disappears. Maybe we should keep the link constant until the next parse? I can see either way on this.
- The link is followed on mouse down, but it should probably be mouse up to follow how other buttons work in the system. (For example, I can't select any text in a link with the mouse because the link just opens.)
  - Should we do something like `ctrl+click` to activate, at least as an option that the client can set? Some editors do that.

I can try to look into the cause of the bugs I mentioned, but I'm not really sure when I'll have time, sorry.

marbetschar:
@mcclurgm good catch - that was a tricky one! The wrong behaviour was because of multibyte string encoding. Or as the documentation puts it:

> Character counts are usually referred to as offsets, while byte counts are called indexes. If you confuse these two ... bad things will happen - [Text Widget Conceptual Overview](https://developer.gnome.org/gtk3/stable/TextWidget.html)

I fixed the error and detection works fine now even for more complex multi-byte-strings with emojis or similar. Feel free to re-test. Regarding the `Ctrl + Click`: I like the idea, but unfortunately I had no luck showing a tooltip indicating this behaviour. Maybe someone else can take over here?

mcclurgm:
Aha--good catch. The joys of text rendering! 😬

Interesting point about the tooltips. I found a mailing list thread with some cursory duckduckgo, but it looks complicated, and the thread never comes up with a complete solution. So you're right, that may be too hard to pull off. As a potential compromise, what if we only followed the link if the mouse releases close to the position where it was pressed? That would allow you to select text without following the link. (For prior art, this is how Mac's TextEdit does it. I know Google Docs shows a popup with a clickable link when you click on link text, and I think Word uses ctrl+click.) I'm not sure how easy a "click" vs a "selection drag" would be to define though: we'd probably have to make some thresholds.

marbetschar:
I'm up for anything which improves the experience. Maybe someone from @elementary/ux has got some additional thoughts on this?

marbetschar:
I _think_ we have something working now: The tooltip is shown and the link is only followed upon `Control + Click`. However, there is one caveat: This only works if the HyperTextView widget is focused - I did not figure out how to figure out if the `Control` key is pressed.

There are some minor things we can improve upon for sure, but overall I'm quite happy with the results. This should be ready now for review.

hanaral:
@marbetschar 
> This only works if the HyperTextView widget is focused - I did not figure out how to determine if the Control key is pressed while the widget is not focused.

This sounds more like a Gala and GTK cooperating thing, which I don't currently think is going to happen. Luckily it's less serious of a problem than not working at all, but it would be cool to see more low level operation for things like this (in the future)

marbetschar:
Found a fix: Binding to `Gtk.Window.list_toplevels` seems to do the trick! Now `Control` presses are recognized, even if HyperTextView is not focused.

hanaral:
That's pretty cool, but does it function if another_app_ is focused (which is what I thought was being talked about)

marbetschar:
> That's pretty cool, but does it function if another _app_ is focused (which is what I thought was being talked about)

Good catch! Of course it does ... not 😅 - but thats pretty much out of scope of what a Gtk Widget can do now 😄

hanaral:
@marbetschar I was going to ask this on twitter but it's probably better off here; is there an 'open in browser'/'send mail to'/'add to calendar'/etc. entry when a context menu is opened on it?

marbetschar:
@hanaral no, there is no context menu handling implemented atm. Currently only a tooltip appears with the "Follow Link" text - and once the user `Control+Click`'s on the link, the full uri is handed over to [`GLib.AppInfo.launch_default_for_uri`](https://valadoc.org/gio-2.0/GLib.AppInfo.launch_default_for_uri.html) - and this function decides whatever happens next.

lib/Widgets/HyperTextView.vala
```diff
@@ -0,0 +1,320 @@
...
+
+            while (match_info.matches ()) {
+                string match_text = match_info.fetch (0);
+
+                /**
+                 * Character counts are usually referred to as offsets, while byte counts are called indexes.
+                 * If you confuse these two, things will work fine with ASCII, but as soon as your
+                 * buffer contains multibyte characters, bad things will happen.
+                 * https://developer.gnome.org/gtk3/stable/TextWidget.html
+                 */
```
mcclurgm, line 168:
```suggestion
```

I think we only need one of these comments. Also I think it should be a single star comment instead of a valadoc comment, since it's not really documentation.

mcclurgm:
Seems to work mostly as intended for me! I'd like to leave approval for someone who's more familiar with Gtk though, I'm pretty new. 

Only thing I can see is that the regex is still a little troublesome. It doesn't detect URIs that don't have a dot in them, like `file:///usr/share` which it sounds like should match? I haven't done a lot of other checking though.

I think we should keep some things in mind for iteration, but IMO it'd probably be better to get a first version working and in use and return to them if needed.
- Should we use the Granite keyboard shortcut styling for the tooltip (like in context menus)? It's a little strange since one of the actions is a click though.
- Should we use ctrl+click to follow the link, or just click but reject click+drag?
- Should we require protocol prefixes (`https:`, `mailto:`) to lower the number of false positives? (I think this is probably fine as is, but again it's hard to tell until it's been in use for a bit.)

lib/Widgets/HyperTextView.vala
marbetschar: Changed as suggested.

hanaral:
> Seems to work mostly as intended for me! I'd like to leave approval for someone who's more familiar with Gtk though, I'm pretty new.
> 
> Only thing I can see is that the regex is still a little troublesome. It doesn't detect URIs that don't have a dot in them, like `file:///usr/share` which it sounds like should match? I haven't done a lot of other checking though.
> 
> I think we should keep some things in mind for iteration, but IMO it'd probably be better to get a first version working and in use and return to them if needed.
> 
>     * Should we use the Granite keyboard shortcut styling for the tooltip (like in context menus)? It's a little strange since one of the actions is a click though.


>     * Should we use ctrl+click to follow the link, or just click but reject click+drag?
> 
>     * Should we require protocol prefixes (`https:`, `mailto:`) to lower the number of false positives? (I think this is probably fine as is, but again it's hard to tell until it's been in use for a bit.)

marbetschar:
As you both pointed out, ultimately it all boils down to a few decisions on how the UX should be designed:

Personally I like the `Control + Click` approach, because it does not interfere with text editing. However, if the TextView is read only it might be ok to activate a link without the need for pressing `Control`. In the read only case, we might not even need a popover.

Also, what about touch screens? Does this even work there right now? If yes, does it feel right to press `Control + Tap`? Unfortunately I can't test this because I don't own such hardware.

cassidyjames:
I'm in favor of making this a bit more strict, at least to start, and requiring a protocol in URLs. Without it, I feel like the false positives are going to be way too annoying. Perhaps it could be an optional property for the developer using the widget to set how broad/strict the matching is, but I also don't want to over-design it from the start.

Matching URLs with protocols would be a good start and solve immediate needs in apps like Tasks and Calendar as far as I know, though.

marbetschar:
@cassidyjames sure, we can start small and take it from there. Should we explicitly require the `mailto:` protocol as well - or do we want to start as small as it gets and strictly require either `http://` or `https://`?

mcclurgm:
The new regex seems good to me, although it doesn't match Unicode.
FWIW, there's an MIT licensed gist that seems to tick almost all the (HTTP) boxes that we could use: https://gist.github.com/dperini/729294. It's a _lot_ more complicated than your version though which is very nice and simple, and I haven't tested compatibility with GRegex or PCRE.

About click behavior: good point about the touchscreen. I haven't thought about that, and I don't have a touchscreen either to check. On iOS Notes.app they disable links when they keyboard is active, but that's not a state change we can use. I agree that ctrl+tap doesn't really make sense. Perhaps we could place a "Follow link…" item in the context menu, which should be accessible by touch? I'm not sure about this though.

I'd be happy to commit the current version and iterate on both of those in future PRs if we want, this discussion is going pretty long.

lib/Widgets/HyperTextView.vala
```diff
@@ -0,0 +1,317 @@
...
+
+ namespace Granite.Widgets {
+
+    /**
+     * This class enables navigatable URLs in Gtk.TextView
+     */
+    public class HyperTextView : Gtk.TextView {
+
+        private uint buffer_changed_debounce_timeout_id = 0;
+        private int buffer_cursor_position_when_change_started = 0;
```
jeremypw, line 28: Shouldnt this be initialized to `-1` (it is set to this after a paste)

lib/Widgets/HyperTextView.vala
```diff
@@ -0,0 +1,317 @@
...
+            } else {
+                warning ("Unable to bind key press events to toplevel window, Control + Click may not behave correct under all circumstances.");
+                // bind to this as a fallback
+                key_press_event.connect (on_key_press_event);
+                key_release_event.connect (on_key_release_event);
+            }
+        }
+
+        private void on_buffer_cursor_position_changed () {
+            if (buffer_cursor_position_when_change_started == 0) {
```
jeremypw, line 90: Shouldnt this be `if (buffer_cursor_position_when_change_started == -1` (when this variable is initialized to `-1`)?

Otherwise the value of this variable cannot remain `0`

lib/Widgets/HyperTextView.vala
```diff
@@ -0,0 +1,317 @@
...
+                buffer_changed_debounce_timeout_id = 0;
+            }
+
+            buffer_changed_debounce_timeout_id = GLib.Timeout.add (300, () => {
+                buffer_changed_debounce_timeout_id = 0;
+
+                var change_start_offset = buffer_cursor_position_when_change_started;
+                var change_end_offset = buffer.cursor_position;
+
+                buffer_cursor_position_when_change_started = 0;
```
jeremypw, line 112: Reset to `-1`? See above

jeremypw:
Despite my comments above, this does seem to work as expected (in general) even when the hyperlink starts at the first character in the document. 

The text "https://https//elementary.io" is wrongly recognized as a valid hyperlink (not necessarily a blocker).

If you do not have Mail set up (e.g. if you just use webmail) then launching `mailto` links fails, but there is probably not a lot that can be done about that here.  Not blocking for this PR.

demo/Views/HyperTextView.vala
```diff
@@ -0,0 +1,42 @@
...
+*
+* You should have received a copy of the GNU Lesser General Public
+* License along with this library; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA.
+*/
+
+public class HyperTextView : Gtk.Grid {
+    construct {
+        var hypertext_label = new Granite.HeaderLabel ("Clickable URLs in a TextView");
```
jeremypw, line 22: Notwithstanding the tooltip, it might be better to add expand on this with e.g. "Hold Ctrl and click to follow the link"

lib/Widgets/HyperTextView.vala
```diff
@@ -0,0 +1,317 @@
...
+                    toplevel_window = window;
+                    break;
+                }
+            }
+
+            if (toplevel_window != null) {
+                toplevel_window.key_press_event.connect (on_key_press_event);
+                toplevel_window.key_release_event.connect (on_key_release_event);
+            } else {
+                warning ("Unable to bind key press events to toplevel window, Control + Click may not behave correct under all circumstances.");
```
comradekingu, line 82:
```suggestion
                warning ("Unable to bind key-press events to top-level window, Control + Click may not always behave correctly.");
```

comradekingu:
Is there a reason for "Unable to …" instead of the usual "Could not …" here?

marbetschar:
@comradekingu no specific reason for "Unable" instead of "Could not" - and no strong feelings either. So I changed it to "Could not".

@jeremypw setting the value offset value to `-1` is by purpose, since it would be set to `0` in case the user edits the first line. If it was set to `0`, the logic would only rescan the edited line - but in case of pasted text we want to behave the logic to behave differently. I now introduced the magic constant `FORCE_FULL_BUFFER_RESCAN_CHANGE_START_OFFSET` in https://github.com/elementary/granite/pull/507/commits/ae41c78b97f0280b59555468daad74fd6a8e1d82 which hopefully makes this distinction more obvious in the code.

All other concerns have been addressed as well.

lib/Widgets/HyperTextView.vala
jeremypw: Seems to work as is.

jeremypw (approved):
I think this works well enough to be useful now.  No doubt tweaks can be made to the regex etc if required in another PR.

marbetschar:
final thoughts @elementary/ux ?

comradekingu:
@marbetschar I am not the UI team, but 
`🔓/🔒 elementary.io` for HTTP/HTTPS
Less silly idea, remove the trailing `\` ?

marbetschar:
@comradekingu I don't understand. Can you please elaborate?

comradekingu:
@marbetschar If it was visually presented as just the domain name/TLD, then a lot could be omitted.
`https://domain.org/`
vs
`domain.org/`
If a lock replaced the HTTPS (like I think Chrome is intending to do). Then visually it is a more simple representation.
or
`🔒 domain.org/`
The padlock would have to be clickable for it to be actually useful (presenting about the SSL chain like in a web browser).

Finally, which is the only feasible idea (?), the ending / after the domain is in similar fashion something that can be done without.
`domain.org/`
vs
`domain.org`

marbetschar:
@comradekingu got it. But changing the appearance of the text is out of scope for this PR.

comradekingu:
@marbetschar That is what I thought :)
"Could not" I use when something didn't work. "Unable" is "impossible" for some reason, without getting into "cannot/can not".
I'd rather err on the side of too many "could not" because translations for "unable" might be even stricter.

tintou:
I think that the best way forward would be to copy/paste it into Calendar (and renamespace it) so that we are sure that it's the good implementation and API

demo/meson.build
```diff
@@ -12,6 +12,7 @@ executable(
     'Views/DialogsView.vala',
     'Views/DynamicNotebookView.vala',
     'Views/FormView.vala',
+    'Views/HyperTextView.vala',
```
jeremypw, line 15: To avoid a name clash (albeit in different namespaces) it may be better to call this `HyperTextViewDemo.vala` Not important I guess.

marbetschar: Adressed the concerns by naming the view `HyperTextViewGrid`.

marbetschar:
My testings with Calendar and Tasks were quite successful. The only thing I noticed is, the popover does not seem to be always shown in Tasks - but from what I found this is due to something closely related to the Tasks app.

So from my point of view, we are good to go here! Anyone from @elementary/desktop-developers want to give this an additional review?

mcclurgm:
I'm checking out the Tasks PR to see. From my perspective from what I've seen in Calendar, this seems good to me though.

I don't understand what you mean by the popover in Tasks. Is that a part of `HyperTextView`?

One question though: is it possible to have links without a domain name? I get that something like `https://e` is highlighted as a link. That may be something to fix if that's an invalid hyperlink address. I'll post a suggestion that would fix this is you want.

lib/Widgets/HyperTextView.vala
```diff
@@ -0,0 +1,324 @@
...
+
+    private bool is_control_key_pressed = false;
+
+    construct {
+        var http_charset = "[a-zA-Z0-9_\\/\\-\\+\\.:@\\?&%=#]";
+        var email_charset = "[a-zA-Z0-9_\\-\\.]";
+        var email_tld_charset = "[a-zA-Z0-9_\\-]";
+
+        var http_match_str = "https?:\\/\\/" + http_charset + "{1,}";
+        var email_match_str = "(mailto:)?" + email_charset + "{1,}@" + email_charset + "{1,}\\." + email_tld_charset + "{1,}";
```
mcclurgm, line 41:
```suggestion
        var http_charset = "[\\w\\/\\-\\+\\.:@\\?&%=#]";
        var email_charset = "[\\w_\\-\\.]";
        var email_tld_charset = "[\\w_\\-]";

        var http_match_str = @"https?:\\/\\/$(http_charset)+?\\.$(http_charset)+";
        var email_match_str = @"(mailto:)?$(email_charset)+@$(email_charset)+\\.$(email_tld_charset)+";
```

You could simplify the regex a little bit. I also added a check that the URL has at least one `.`. If you don't want that, `http_match_string` should be
```vala
@"https?:\\/\\/$(http_charset)+"
```

Here's how it currently matches up against the list of URLs to test against:
regex-negative
regex-positive

marbetschar:
@mcclurgm thanks for your feedback!

> is it possible to have links without a domain name? I get that something like https://e is highlighted as a link. That may be something to fix if that's an invalid hyperlink address. I'll post a suggestion that would fix this is you want.

I think we can safely assume there needs to be at least one `.` in an url for now. I also like the idea of getting rid of the string concatenation, so I adjusted the regex accordingly!

However, I wasn't sure what the `?` before the dot is supposed to do - so I omitted it and the regex now reads: `@"https?:\\/\\/$(http_charset)+\\.$(http_charset)+"` - which does work fine on my end.

> I don't understand what you mean by the popover in Tasks. Is that a part of HyperTextView?

Yes, the popover is part of HyperTextView and appears if you hover over a hyperlink and wait for a bit. It should also appear in Tasks, but does not right now - no clue why. But I think that's a minor issue and related to the Tasks app, so we should be good to merge HyperTextView regardless.

Here's how it looks like in the `granite-demo` app:

Screenshot from 2021-09-19 11-45-19

mcclurgm:
> However, I wasn't sure what the ? before the dot is supposed to do - so I omitted it and the regex now reads: @"https?:\\/\\/$(http_charset)+\\.$(http_charset)+" - which does work fine on my end.

That should work just as well, I think. The `?` is a "non-greedy quantifier," so it matches as few instances of the pattern as possible. So if you have multiple dots, it would capture only up to the first one with the first quantifier, then the rest would be captured by the next quantifier. In our case, it should be indistinguishable except for subpatterns, which we aren't using. It also doesn't behave quite how I initially though with regard to subpatterns, so let's keep it how you implemented it for now.

I am seeing the tooltip in Tasks, but strangely it's intermittent and only appears on certain links some of the time. I may try to track it down, but if you think it's Tasks-related, that doesn't have to be a blocker for this PR.

marbetschar:
Thanks to the engineering work of @mcclurgm and @jeremypw we even fixed the popup issue in the Tasks app now - thank you both for this missing piece!

We are now at a point with HyperTextView where I'm fully confident that HyperTextView provides a smooth experience as I'm not aware of any issues. So IMHO we are here finally able to merge - thanks for this team effort! :partying_face:

mcclurgm:
For what it’s worth I’d say LGTM. Makes sense to get a final review from Danielle though. 

I have some thoughts for enhancement to discuss, but let’s wait on those until we commit a functional version. (Mostly just colors and branding could be nice to customize.)

lib/Widgets/HyperTextView.vala
```diff
@@ -0,0 +1,324 @@
...
+ *
+ *  You should have received a copy of the GNU Lesser General
+ *  Public License along with this library; if not, write to the
+ *  Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ *  Boston, MA 02110-1301 USA.
+ */
+
+/**
+* This class enables navigatable URLs in Gtk.TextView
+*/
```
mcclurgm, line 22: Sorry, one more minor thing. Since this is a library, we should probably include "since" data. I added a suggestion you can just commit, assuming I guessed the actual release it'll into correctly.

```suggestion
 * @since 6.1.2
 */
```

danrabbit (approved):
Let's :shipit:  :)

PR #510, opened 2021-07-18
Make sure to work accel_to_string with multiple modifiers

ryonakano:
## Before
If you pass `accel_to_string` to a accel that only contains modifiers e.g. "\<Shift\>\<Control\>" or "\<Shift\>\<Alt\>", it only return either of them, as seen in Shortcut Overlay:

Screenshot from 2021-07-18 23-36-09

## After
Screenshot from 2021-07-18 23-33-31

### DIFF
--- a/lib/Widgets/Utils.vala
+++ b/lib/Widgets/Utils.vala
@@ -170,7 +170,7 @@ public static string accel_to_string (string? accel) {
             break;
         default:
             // If a specified accelarator contains only modifiers e.g. "<Control><Shift>",
-            // we get anything from accelerator_get_label method, so skip that case
+            // we don't get anything from accelerator_get_label method, so skip that case
             string accel_label = Gtk.accelerator_get_label (accel_key, 0);
             if (accel_label != "") {
                 arr += accel_label;
--- a/lib/Widgets/Utils.vala
+++ b/lib/Widgets/Utils.vala
@@ -169,11 +169,16 @@ public static string accel_to_string (string? accel) {
             arr += _("Right Shift");
             break;
         default:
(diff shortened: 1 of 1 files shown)


lib/Widgets/Utils.vala
```diff
@@ -169,11 +169,16 @@ public static string accel_to_string (string? accel) {
             arr += _("Right Shift");
             break;
         default:
-            arr += Gtk.accelerator_get_label (accel_key, 0);
+            // If a specified accelarator contains only modifiers e.g. "<Control><Shift>",
+            // we get anything from accelerator_get_label method, so skip that case
```
jeremypw, line 173: Do you mean "we don't get anything ..."?

lib/Widgets/Utils.vala
```diff
@@ -169,11 +169,16 @@ public static string accel_to_string (string? accel) {
...
+            string accel_label = Gtk.accelerator_get_label (accel_key, 0);
+            if (accel_label != "") {
+                arr += accel_label;
+            }
             break;
     }
 
-    if (accel_key != 0 && accel_mods != 0) {
+    if (accel_mods != 0) {
```
jeremypw, line 181: I am not sure we can assume all accelerators have a modifier.  E.g. `F2` is used as an accelerator for "Rename" in Files.

lib/Widgets/Utils.vala
ryonakano: Yeah you're right. Sorry for my poor English 😅

jeremypw (approved):
I can confirm this works - code looks good.  Thanks for fixing this!

PR #518, opened 2021-08-16
l10n: Correctly setup the locales

bobby285271:
Provides the directory where the locales are actually installed.

We are [packaging this](https://github.com/NixOS/nixpkgs/pull/130380#issuecomment-895720580) in NixOS, and due to NixOS 's special `localedir`, we cannot apply the translations without this patch.

### DIFF
--- a/lib/Config.vala.in
+++ b/lib/Config.vala.in
@@ -19,4 +19,5 @@
 
 namespace Granite {
     private const string GETTEXT_PACKAGE = "@GETTEXT_PACKAGE@";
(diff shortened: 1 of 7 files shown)


tintou:
That's indeed a good point but that's not going to be the solution here, I'll try to provide a patch to make it work correctly.

bobby285271:
Feel free to close this PR in favor of yours :-)

tintou:
@bobby285271 If you can try #519

PR #519, opened 2021-08-16
l10n: Correctly setup the locales

tintou:
Provides the directory where the locales are actually installed.
Localization is only setup on-demand.

Supersedes #518


bobby285271:
I can confirm this works on NixOS, thanks for your work!

davidmhewitt (approved):

PR #520, opened 2021-08-16
Release 6.1.1

cassidyjames:
Requested for #519 but it'd also be nice to ship #510.


marbetschar:
Any chance we can get #507 shipped too?

PR #527, opened 2021-09-13
Use Portal For clock-format

jhaygood86:
Use portal instead of GSettings for clock-format

### DIFF
--- a/lib/DateTime.vala
+++ b/lib/DateTime.vala
@@ -8,6 +8,12 @@
  * getting the default translated format for either date and time.
  */
 namespace Granite.DateTime {
+
+    [DBus (name = "org.freedesktop.portal.Settings")]
+    private interface FDO.Portal.Settings : Object {
+        public abstract Variant read (string @namespace, string key);
+    }
+
     /**
      * Gets a default translated time format.
      * The function constructs a new string interpreting the //is_12h// and //with_second// parameters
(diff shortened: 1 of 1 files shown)


jhaygood86:
Honestly, #528 is a more complete change. We can probably close this out if we decide to go with #528 instead.

lib/DateTime.vala
```diff
@@ -96,8 +102,16 @@ namespace Granite.DateTime {
...
-        var h24_settings = new GLib.Settings ("org.gnome.desktop.interface");
-        var format = h24_settings.get_string ("clock-format");
+        FDO.Portal.Settings settings_bus = GLib.Bus.get_proxy_sync (
+                    GLib.BusType.SESSION,
+                   "org.freedesktop.portal.Desktop",
+                   "/org/freedesktop/portal/desktop"
+                );
+        
+        var clock_format_variant = settings_bus.read ("org.gnome.desktop.interface", "clock-format").get_variant ();
+        var format = clock_format_variant.get_string ();
```
tintou, line 112:
```suggestion
        string format;
        try {
            var settings_bus = GLib.Bus.get_proxy_sync<FDO.Portal.Settings> (
                        GLib.BusType.SESSION,
                       "org.freedesktop.portal.Desktop",
                       "/org/freedesktop/portal/desktop"
                    );

            var clock_format_variant = settings_bus.read ("org.gnome.desktop.interface", "clock-format").get_variant ();
            format = clock_format_variant.get_string ();
        } catch (GLib.Error e) {
            debug ("Unable to connect to desktop portal (%s), using GSettings", e.message);
            var h24_settings = new GLib.Settings ("org.gnome.desktop.interface");
            format = h24_settings.get_string ("clock-format");
        }
```

lib/DateTime.vala
```diff
@@ -96,8 +102,23 @@ namespace Granite.DateTime {
...
+
+            var clock_format_variant = settings_bus.read ("org.gnome.desktop.interface", "clock-format").get_variant ();
+            format = clock_format_variant.get_string ();
+        } catch (GLib.Error e) {
+            debug ("Unable to connect to desktop portal (%s), using GSettings", e.message);
+            var h24_settings = new GLib.Settings ("org.gnome.desktop.interface");
+            format = h24_settings.get_string ("clock-format");
+        }
+
+
```
cassidyjames, line 121: Lint

```suggestion
```

Issue #525, opened 2021-09-10
Make Granite use defaults from desktop settings whenever possible


marbetschar:
## Problem

There are quite a lot of places on the desktop where we'd like to display a time to the user. To use the correct format we can use [`Granite.DateTime.get_default_time_format`](https://github.com/elementary/granite/blob/e4a3e839b5513f7e364f1910505518fab1ffa149/lib/DateTime.vala#L24).

However, this function does not provide a reasonable default value for its `is_12h` argument because its hard coded. But in user facing cases we want to respect the time setting of the user so there are a bunch of cases where we need to retrieve the setting and pass the correct value to Granite. Namely Tasks, Calendar and Mail - potentially others:

https://github.com/search?q=org%3Aelementary+get_default_time_format&type=code

## Proposal

The example above is specific to one case, but maybe there are other similar things I'm missing here? If there is more to this than the 24h time format, then it would be beneficial to tell granite it should use the user settings instead of having them hard wired in the code. If the 24h time format is the only one, then it would be great to have its default set to whatever the user is expecting.

## Prior Art

- I've seen there is Granite.Settings - but I'm not sure if that can be used for this case: https://valadoc.org/granite/Granite.Settings.html
- Granite itself already implements a (private) helper function for detecting the right default value for the 24h format: https://github.com/elementary/granite/blob/7190ee81dcd66c2b28445edf2b7baabdb8524add/lib/DateTime.vala#L98-L102

marbetschar:
Related to https://github.com/elementary/granite/issues/522

PR #528, opened 2021-09-17
Expose clock format as granite setting

jhaygood86:
This builds on #527 , but also solves some needs in #525 

This:

- exposes the clock format as setting in Granite.Settings, including listening from updates, using the portal. This also allows applications to notify on the update even inside a flatpak'ed application
- updates internal APIs and widgets to use the new setting, including listening on updates. This magically makes applications work inside flatpak as well as makes them more dynamic (i.e., when the system setting is changed, reflect that change instantly)
- adds a new API to format a GLib.DateTime directly using the existing APIs and the system settings so that applications don't have to duplicate the format logic everywhere.

### DIFF
--- a/lib/Widgets/Settings.vala
+++ b/lib/Widgets/Settings.vala
@@ -13,11 +13,21 @@ namespace Granite {
     interface FDO.Accounts : Object {
         public abstract string find_user_by_name (string username) throws GLib.Error;
     }
+    
+    [DBus (name = "org.freedesktop.portal.Settings")]
+    private interface FDO.Portal.Settings : Object {
+        public abstract Variant read (string @namespace, string key);
+        public abstract signal void setting_changed (string @namespace, string key, Variant @value);
+    }
 
     /**
      * Granite.Settings provides a way to share Pantheon desktop settings with applications.
      */
     public class Settings : Object {
+    
+        private const string GNOME_DESKTOP_INTERFACE = "org.gnome.desktop.interface";
+        private const string CLOCK_FORMAT_KEY = "clock-format";
+        
         /**
          * Possible color scheme preferences expressed by the user
          */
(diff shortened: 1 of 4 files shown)


jhaygood86:
@danrabbit It'll still be useful. I've adjusted to account for the new Portal.Settings class being available.

lib/Widgets/Settings.vala
```diff
@@ -69,6 +73,36 @@ namespace Granite {
...
+        private ClockFormat? _clock_format = null;
+        /**
+         * Whether the user would prefer a 12 hour or 24 hour clock
+         */
+        public ClockFormat clock_format {
+            get {
+                if (_clock_format == null) {
+                    setup_clock_format ();
+                }
+                return _clock_format;
```
tintou, line 99:
```suggestion
                return _clock_format;
```
_clock_format might be null, so we need to explicitely do/
`return _clock_format ?? ClockFormat.24H`

lib/Widgets/Settings.vala
```diff
@@ -69,6 +73,36 @@ namespace Granite {
             }
         }
 
+        /**
+         * Possible clock format preferences expressed by the user
+         */
+        public enum ClockFormat {
```
tintou, line 79: Reverse the order of the fields, we want 24H = 0 and 12H = 1 :slightly_smiling_face:

lib/Widgets/Settings.vala
```diff
@@ -139,5 +179,47 @@ namespace Granite {
...
+                setup_portal ();
+
+                var clock_format_variant = portal.read (GNOME_DESKTOP_INTERFACE, CLOCK_FORMAT_KEY).get_variant ();
+                var format = clock_format_variant.get_string ();
+
+                set_clock_format_from_nick (format);
+
+                portal.setting_changed.connect ((@namespace, key, @value) => {
+                    if (@namespace == GNOME_DESKTOP_INTERFACE && key == CLOCK_FORMAT_KEY) {
+                        var updated_format = @value.get_string ();
```
tintou, line 194:
```suggestion
                        unowned var updated_format = @value.get_string ();
```

PR #529, opened 2021-09-21
Support the settings portal where is possible

Marukesu:
blocked by elementary/settings-daemon#38

This make the functions that read some setting provided by the portal to try it first, allowing we get the right values when sandboxed.

### DIFF
--- a/lib/DateTime.vala
+++ b/lib/DateTime.vala
@@ -96,8 +96,20 @@ namespace Granite.DateTime {
      * @return true if the clock format is 12h based, false otherwise.
      */
     private static bool is_clock_format_12h () {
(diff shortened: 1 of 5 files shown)


jhaygood86:
Can you look at merging this with https://github.com/elementary/granite/pull/528 ?

I don't think you need a new Portal Settings class, the existing Settings class is probably fine.

Marukesu:
> Can you look at merging this with #528 ?

I think #528 is out of scope for this PR, since this aren't adding a new API, but just making the current functions use the portal. However i can rebase it on top of #527.

> I don't think you need a new Portal Settings class, the existing Settings class is probably fine.

I made it this way so that we declare the interface (and any other interface we need) just in one place.

Issue #489, opened 2021-04-28
ValidatedEntry: Add "minimum_length" and "maximum_length" properties


jeremypw:
## Problem

In some cases it useful to set a minimum and/or maximum length properties for an input, in addition to other requirements.  Although this can be done with Regex and other ways, it would be nice to have an easier way.

## Proposal

Set these as properties on creation and leave it to the ValidatedEntry to handle their enforcement.

danrabbit:
Gtk.Entry has `max_length` and `max_width_chars`. I think one of these properties should restrict character input already. But yeah a minimum length before the input is considered valid would be cool. We should probably base the name on whichever of these gtk properties does the thing we want

cassidyjames:
You're looking for `max_length` which restricts the numbers of characters entered (whereas `max_width_chars` is the widget's maximum width in approximate characters, like when you set that on a label, I guess).

So `min_length` would be a great property to add.

jeremypw:
Yes, I forgot that "max_length" was already a property of ValidatedEntry :grimacing:

danrabbit:
Going to copy this relevant information from a recent fix in switchboard:

> I replaced the method `length` to check the number of characters with `char_count` because the original code assumed that every character was 1 byte, while in some languages some characters sizes could be more than 1 byte.
>
> For example this string is 8 bytes long, but only 4 characters long: áéóü.

PR #531, opened 2021-10-11
489: Add minimum character count check to ValidatedEntry

jpgleeson:
As per issue [489](https://github.com/elementary/granite/issues/489), this adds a check to the validated entry box that checks against a number of characters issued at construction. This works correctly with character of multiple byte sizes, as discussed in the issue linked.

### DIFF
--- a/data/granite.appdata.xml.in
+++ b/data/granite.appdata.xml.in
@@ -22,7 +22,7 @@
         </ul>
       </description>
     </release>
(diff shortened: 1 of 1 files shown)


jpgleeson:
Danielle, apologies for the issues with the first submission. I've refactored things to fit with the code style guide.

I've also refactored the existing `from_regex` constructor to utilise the same function as the generic constructor. This lets code bases keep using this as is while adding a more generic usage also.

PR #544, opened 2021-12-09
Enable ARM builds

isantop:
Should allow for arm64 builds for granite 6 (required by Pop_Shop and probably the installer on ARM).


isantop:
Whoops, wrong fork!

PR #545, opened 2021-12-17
DynamicNotebook: Do not use deprecated child property

jeremypw:
Small preparation for Gtk4.


tintou (approved):

davidmhewitt:
`Gtk.MenuItem` doesn't exist in GTK4, so there was probably very little point in making this change. All of the Menu code will need to be re-written anyway.

I have a GTK4 branch of Granite that I'm slowly hacking away at here: https://github.com/elementary/granite/pull/542

I would prefer that any GTK4 related changes were just pushed into there or proposed as changes into there to avoid needless conflicts.

PR #546, opened 2022-01-04
Dialogs: Update for GTK4 styles

danrabbit:
- [x] In conjunction with: https://github.com/elementary/stylesheet/pull/1145
- [x] Uses https://github.com/elementary/stylesheet/pull/1165 for terminal padding

* The selector for `MessageDialog` in GTK4 is now `window.dialog.message` so imitate that
* In GTK4, `MessageDialog` uses the `title` class for its primary label, so we can piggyback on that. 
* Add a style constant for `dialog-content-area` so we can handle dialog margins in CSS
* Pack dialog contents into a window handle so that dialogs are draggable without a titlebar
* Fix icon sizes
* In the Terminal/Details widget, we use margin properties of TextView instead of padding in CSS because of the different behavior with scroll overflow

## BEFORE

Screenshot from 2022-01-04 12 27 36
Screenshot from 2022-01-04 12 27 28

## AFTER

Screenshot from 2022-01-04 12 21 24
Screenshot from 2022-01-04 12 21 17


cassidyjames:
>Pack dialog contents into a window handle so that dialogs are draggable without a titlebar

It looks like this prevents highlighting text in the details, e.g. for copy+paste; do you know if we can exclude the details area from the window handle somehow?

danrabbit:
@cassidyjames okay should be fixed!

cassidyjames (approved):

PR #548, opened 2022-01-05
Remove old icons

danrabbit:
As mentioned on Slack. These are legacy from old widgets that have been removed like `Granite.AppMenu`


davidmhewitt:
Looks like `appointment` is used in the time picker?

danrabbit:
Yeah but it's the symbolic version, so it isn't covered by this anyways

davidmhewitt (approved):
Ah, indeed!

PR #551, opened 2022-01-05
Constants: Add destructive-action

danrabbit:
It was removed from GTK


davidmhewitt (approved):
Makes sense since we have suggested action in here too.

PR #553, opened 2022-01-05
Create HeaderLabel

danrabbit:
Kind of dumb, but it works :shrug:

### DIFF
--- a/lib/Widgets/HeaderLabel.vala
+++ b/lib/Widgets/HeaderLabel.vala
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2017-2022 elementary, Inc. (https://elementary.io)
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ */
+
+
+/**
+ * HeaderLabel is a start-aligned {@link Gtk.Label} with the Granite H4 style class
+ */
+public class Granite.HeaderLabel : Gtk.Box {
+    public string label { get; construct set; }
+
+    /**
+     * Create a new HeaderLabel
+     */
+    public HeaderLabel (string label) {
+        Object (label: label);
+    }
+
+    construct {
+        var label_widget = new Gtk.Label (label) {
+            xalign = 0
+        };
+        label_widget.add_css_class (Granite.STYLE_CLASS_H4_LABEL);
+
+        append (label_widget);
+        halign = Gtk.Align.START;
+
+        bind_property ("label", label_widget, "label");
+    }
+}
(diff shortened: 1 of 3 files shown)


cassidyjames:
@danrabbit I think you forgot to add `HeaderLabel.vala`?

danrabbit:
Oh yup that's an important step

lib/Widgets/HeaderLabel.vala
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright 2017-2022 elementary, Inc. (https://elementary.io)
+ * SPDX-License-Identifier: LGPL-2.1-or-later
+ */
+
+
+/**
+ * HeaderLabel is a start-aligned {@link Gtk.Label} with the Granite H4 style class
+ */
+public class Granite.HeaderLabel : Gtk.Box {
```
tintou, line 10:
```suggestion
public class Granite.HeaderLabel : Gtk.Widget {
```
We shouldn't subclass anything else than Gtk.Widget in GTK4
You then need to add:
```vala
static construct {
    set_layout_manager_type (typeof (Gtk.BoxLayout));
}

~HeaderLabel () {
    while (unowned var child = get_first_child () != null) {
        child.unparent ();
    }
}
```

danrabbit: Hm okay, so I tried this, but I'm getting an error about `append` not being available in Gtk.Widget. Is there some other way I have to tell the layout manager to make this widget a child here?

```
    public string label { get; construct set; }

    /**
     * Create a new HeaderLabel
     */
    public HeaderLabel (string label) {
        Object (label: label);
    }

    static construct {
        set_layout_manager_type (typeof (Gtk.BinLayout));
    }

    construct {
        var label_widget = new Gtk.Label (label) {
            xalign = 0
        };
        label_widget.add_css_class (Granite.STYLE_CLASS_H4_LABEL);

        child = label_widget;
        halign = Gtk.Align.START;

        bind_property ("label", label_widget, "label");
    }

    ~HeaderLabel () {
        while (get_first_child () != null) {
            get_first_child ().unparent ();
        }
    }
}
```

Also, if I use `BinLayout` instead, I probably don't need the `while` loop right?

ZenitsuDev: Hi! @danrabbit , I think we can `label_widget.set_parent (this);` or 'insert_after (label_widget, null);` instead of child property

danrabbit: That worked, Thanks!

PR #554, opened 2022-01-06
Constants: use header class names from Adw

danrabbit:
- [ ] Goes with https://github.com/elementary/stylesheet/pull/1116

Makes us more compatible with Adwaita at no real cost


cassidyjames (approved):
lfg

PR #555, opened 2022-01-06
Create Granite.Placeholder

danrabbit:
This is a widget to replace `Welcome` and `AlertView` with a single placeholder. Requires https://github.com/elementary/stylesheet/pull/1154

* It supports actions, icons, and adding style classes like `accent` or `warning`.
* The `append_button ()` method returns a button so you can do things like assign an action or add style classes
* All margins, padding, and icon sizes are set in CSS so it's scalable

Screenshot from 2022-01-06 14 49 01
Screenshot from 2022-01-06 14 49 04
Screenshot from 2022-01-06 14 49 06


cassidyjames (commented):
Wanna get some @elementary/desktop-developers feedback here on code style and GTK4 conventions and whatnot, but UI-wise I'm happy with this.

A reminder that we talked about the left-alignment of button icons on a call and @danrabbit mentioned looking into moving the padding or margin from the icon to the label if that would give us a few more px on the left edge; I also want to make sure that doesn't make focus styles look too janky though.

danrabbit:
@cassidyjames alright, how do you feel about this?

Screenshot from 2022-01-06 14 49 01

Screenshot from 2022-01-06 14 51 01

cassidyjames:
@danrabbit I think I can live with that. :shipit:

lib/Widgets/Placeholder.vala
```diff
@@ -0,0 +1,159 @@
...
+     * @param title_text main title for new Welcome Page
+     * @param subtitle_text subtitle text for new Welcome Page
+     */
+
+    /**
+     * Makes new AlertView
+     *
+     * @param title the first line of text
+     * @param description the second line of text
+     * @param icon_name the icon to be shown
```
Exalm: The last 2 params don't seem to exist.

lib/Widgets/Placeholder.vala
```diff
@@ -0,0 +1,167 @@
...
+        bind_property ("description", description_label, "label");
+        bind_property (
+            "description", description_label, "visible", BindingFlags.SYNC_CREATE | BindingFlags.DEFAULT,
+            (binding, srcval, ref targetval) => {
+                var label = (string) srcval;
+                if (label != null) {
+                    targetval.set_boolean (true);
+                } else {
+                    targetval.set_boolean (false);
+                }
```
Exalm, line 101:
```
targetval.set_boolean (label != null);
```
?

Though probably want to check if it's not empty as well.

lib/Widgets/Placeholder.vala
```diff
@@ -0,0 +1,167 @@
...
+
+                return true;
+            },
+            null
+        );
+    }
+
+    ~Placeholder () {
+        get_first_child ().unparent ();
```
Exalm, line 126: AFAIK this is finalize, not dispose, so might be too late. You probably want to override dispose specifically.

lib/Widgets/Placeholder.vala
```diff
@@ -0,0 +1,167 @@
...
+            wrap = true,
+            xalign = 0
+        };
+
+        var grid = new Gtk.Grid ();
+        grid.attach (image, 0, 0, 1, 2);
+        grid.attach (label_widget, 1, 0);
+        grid.attach (description_widget, 1, 1);
+
+        var button = new Gtk.Button () {
```
Exalm, line 158: Intentional tjhat there's no `.flat`?

lib/Widgets/Placeholder.vala
```diff
@@ -0,0 +1,167 @@
...
+     * Appends new {@link Gtk.Button} to the placeholder's action area
+     *
+     * @param icon the {@link GLib.Icon} that describes this action
+     * @param label a {@link string} to use as the title for this action. It should use Title Case.
+     * @param description a {@link string} to use as a description for this action. It should use sentence case.
+     * @return a {@link Gtk.Button} representing this action
+     */
+    public Gtk.Button append_button (Icon icon, string label, string description) {
+        var image = new Gtk.Image.from_gicon (icon) {
+            icon_size = Gtk.IconSize.LARGE
```
Exalm, line 139: Any reason to set large on every image? The only thing it does is adding `.large` style class, as sizes are now defined in css.

lib/Widgets/Placeholder.vala
danrabbit: Yeah in SCSS I'm extending `.flat` here. Could just set it in the widget itself though! I don't feel super strongly about it

Exalm: That works too, though after libadwaita header bars I'm super conscious about not extending things if possible - the amount of generated css snowballs very very quickly with this.

lib/Widgets/Placeholder.vala
danrabbit: Yeah I could just not set it here and rely totally on setting it in our stylesheet, but this gets me 32px in `default` as well so it's slightly less broken even without actual support in Adwaita

Exalm: Ok.

lib/Widgets/Placeholder.vala
danrabbit: Good call! Fixed

lib/Widgets/Placeholder.vala
danrabbit: Good point. Switched to adding flat class here

lib/Widgets/Placeholder.vala
tintou: In this case we control the first child lifetime so it is alright

cassidyjames (approved):

PR #559, opened 2022-01-11
Fix popovers for DatePicker & TimePicker

danrabbit:
Review with hide whitespace changes

I ended up opting not to use arrows and position the popover on the icon because it just looked cleaner and more native to do it this way as a menu style aligned to the end of the entry:

Screenshot from 2022-01-10 16 26 08
Screenshot from 2022-01-10 16 26 01

While we're here:

* Remove the `widgets` namespace
* Put icons describing the entry in the primary position, and dropdown arrows in the secondary position

- [x] Needs https://github.com/elementary/stylesheet/pull/1159 for proper popover padding


tintou:
It at least is working better than current main. The date picket seem to not work correctly here and only respond from time to time. Also the widget isn't usable with the keyboard but that's probably something we can iterate on.

danrabbit:
@tintou I made a separate branch here to address the date picker's responsiveness: https://github.com/elementary/granite/pull/571

It looks like unfortunately the calendar widget not being keyboard accessible is an issue with GTK itself, so we should probably pursue that upstream

cassidyjames (approved):

PR #568, opened 2022-01-13
Utils: Remove Granite.CloseButtonPosition

danrabbit:
This appears to be used only once in Gala


JoseExposito:
Removed from Gala:
https://github.com/elementary/gala/pull/1364

PR #573, opened 2022-01-19
GTK4/Remove SourceList

### LINKED ISSUE #563: Port or Remove Granite.SourceList
It appears that only Code and Mail are using SourceList.

Photos, Files, and Tasks have their own custom treeview implementation or subclass ListBox.

We should decide if it's worth porting SourceList to GTK4, we want a new sidebar implementation, or apps should implement their own sidebars as needed

danrabbit:
Fixes #563 

Removes SourceList as well as CellRenderedBadge and CellRendererExpander which are only used by SourceList.

If we want to have a shared sidebar widget, it probably should be based on something else and with a different API than treeview, so remove this for now and we can consider adding in a new Granite.Sidebar widget later


danrabbit:
No objections, and it sounds like if we want a Sidebar widget we'd remove this anyways and do a new API, so :rocket:

PR #575, opened 2022-01-19
Remove SeekBar

### LINKED ISSUE #564: Port or Remove Granite.SeekBar
It looks like SeekBar is only used in Videos now. The new Music has its own seekbar implementation. It seems there are some third party apps using SeekBar.

The old seekbar was written to be used with the clutter gstreamer APIs, so if we want to port it it'll need a rewrite anyways.

Need to decide if it's worth carrying this widget or apps should just do their own custom implementation. Especially now that Gtk.Video element is a thing

danrabbit:
Fixes #564 

I'm inclined to remove this for now since it's only used in Videos and it's built for Clutter.Gst which means a new implementation will have to break API anyways. We can always add in a new widget later if it's necessary


danrabbit:
No Objections so :rocket:

PR #578, opened 2022-01-24
Remove DynamicNotebook

### LINKED ISSUE #572: Fix or Remove Granite.DynamicNotebook
These days there's a good TabBar implementation in LibAdwaita: https://valadoc.org/libadwaita-1/Adw.TabBar.html

It's used in Epiphany and is much more modern than Gtk.Notebook with nice animations, improved Drag and Drop, etc.

Is it worth fixing/maintaining dynamic notebook or should we drop it for GTK4 and use Adw.TabBar instead?

danrabbit:
Fixes #572 

Sounds like the plan is to try using [Adw.TabBar](https://valadoc.org/libadwaita-1/Adw.TabBar.html) instead


danrabbit:
I've got a confirmation on the issue report and no objections so :rocket:

PR #579, opened 2022-01-26
Remove Granite.Services.Logger

danrabbit:
It looks like this is only used in Files and Terminal: https://github.com/search?q=org:elementary+Granite.Services.Logger&type=code

But it kinda looks like Files doesn't actually use it since it doesn't seem to ever initialize the logger?


danrabbit:
From @tintou in Slack:
> I'm 100% fine with this being removed :wink:

PR #580, opened 2022-01-26
Toast: subclass Widget, redo styles

danrabbit:
Modernize this a bit. Review with "Hide whitespace changes"

Uses: https://github.com/elementary/stylesheet/pull/1183

* Subclass Gtk.Widget instead of Gtk.Box
* Remove the `Widgets` namespace
* Move all padding and margins to the stylesheet
* set css name
* We support `circular` so use this here instead of `close-button`. Better compat with Adwaita/GTK4 Default
* We don't yet support `osd` in the stylesheet, but adding it here improves compat wth Adwaita/GTK4 Default so why not


danrabbit:
Hey @jeremypw everything in the main branch is GTK4 now and can built on OS 7 daily yes :)

danrabbit:
Gonna merge this since the Stylesheet branches got merged

PR #599, opened 2022-04-25
Release 7.0.0

danrabbit:
Needs Review:
- [x] #600
- [x] #602
- [x] #603
- [x] #605


davidmhewitt:
I think we need to think about how this is going to work with the release action before merging. For starters, the release action still targets the odin branch.

danrabbit:
@davidmhewitt oops you're right. Fixed that!

If you mean about releasing versions 6 and 7, on Launchpad we can build for jammy from the `odin` branch if we need to do another release of 6.x

davidmhewitt (approved):

PR #600, opened 2022-04-25
Remove AsyncImage

### LINKED ISSUE #574: Port or Remove Granite.AsyncImage
Isn't this case covered by `loadable_icon`? https://valadoc.org/gio-2.0/GLib.LoadableIcon.load_async.html

It looks like we currently use this in System Settings for wallpaper images and in the Applications Menu

danrabbit:
Fixes #574 

Didn't get any feedback on the linked issue so :shrug:. We can always port it later if it's necessary, but breaking API after release would be a no no


davidmhewitt (changes requested):
I believe there's some `AsyncImage` related code in the `demo` folder that should be removed too.

JoseExposito (approved):

JoseExposito:
I think that everything is removed now, Leto see if David agrees.

PR #603, opened 2022-04-25
Remove Drawing

### LINKED ISSUE #567: Remove BufferSurface
This is used once in Photos: https://github.com/elementary/photos/blob/77a6b2eb0d847b6f6388e2e9924f3feb4f410195/src/Resources.vala#L264

But it's primarily used in Gala. Is this something we need in a GTK4 world with Gtk.CSS etc?

danrabbit:
Fixes #567

* BufferSurface is only used in Gala
* RoundedRectangle is only being used in Gala and to mask images (which we can do natively in Gtk now with `overflow`)
* average_color () is already copied in the places it's used
* Granite.Color is only used in Gala as well


JoseExposito:
This functions would need to be migrated to Gala, right?

I think that Gala should avoid GTK (in reference to your comment about `overflow`) and stick to its drawing API.

davidmhewitt:
Agreed, this shouldn't be merged until Gala has replacements or else the daily images will break as these methods will go missing from the shared library.

And yes, when adding things to Gala, GTK should be avoided. There's already issues with the GTK code in Gala that's preventing the Wayland transition.

davidmhewitt (approved):

PR #610, opened 2022-05-01
Symbols: Fix namespace for Toast and OverlayBar

danrabbit:
https://launchpadlibrarian.net/599357691/buildlog_ubuntu-jammy-arm64.granite7_7.0.0+r2306+pkg127~ubuntu7.1_BUILDING.txt.gz


davidmhewitt:
I already added these versions here:
https://github.com/elementary/granite/pull/608/files

Seems that `dpkg-gensymbols` didn't remove the old ones when I ran it. So I think you just need to remove these lines rather than change them.

danrabbit:
@davidmhewitt makes sense. Fixed!

davidmhewitt (approved):

PR #613, opened 2022-05-26
Make Toast Wrap

ZenitsuDev:
When the 'Grantie.Toast' title gets too long, for example when used to toast long filenames, the 'Granite.Toast' continues to expand horizontally regardless of the parent window. This aims to fix it.

**Before** 
Screenshot from 2022-05-26 08 28 44

**After**
Screenshot from 2022-05-26 08 42 24

### DIFF
--- a/demo/Views/ToastView.vala
+++ b/demo/Views/ToastView.vala
@@ -26,6 +26,7 @@ public class ToastView : Gtk.Box {
 
         overlay.add_overlay (box);
         overlay.add_overlay (toast);
+        overlay.set_measure_overlay (toast, true);
 
         button.clicked.connect (() => {
             toast.send_notification ();
--- a/lib/Widgets/Toast.vala
+++ b/lib/Widgets/Toast.vala
@@ -71,7 +71,9 @@ public class Granite.Toast : Gtk.Widget {
             visible = false
         };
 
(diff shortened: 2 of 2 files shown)


jeremypw (changes requested):
This causes a problem when the toast is not a simple label
Screenshot from 2022-06-08 18 14 34
In master, the demo toast does not wrap.

jeremypw:
The solution seems to be to add `natural_wrap_mode = Gtk.NaturalWrapMode.NONE` to the properties of the `notification_label`.

ZenitsuDev:
@jeremypw would it be bad to add something like:
`if (parent is Gtk.Overlay) {
    ((Gtk.Overlay) parent).set_measure_overlay (this, true);
}`

jeremypw:
@ZenitsuDev I have tried it both ways with the `main` branch of the `io.elementary.music` which is the only elementary project currently using a Toast and Granite-7.   If you do not `set_measure_overlay` then the toast extends outside the left-hand pane (the queue) but otherwise nothing bad happens.  On balance, I think it better to leave each toast user to decide whether or not `set_measure_overlay` is needed in their particular case rather than try to be too clever about it.

ZenitsuDev:
It seems fixed to me now. Thank you @jeremypw . Is it for you too?

jeremypw (approved):
Looks good now - thanks!

PR #614, opened 2022-06-02
Constants: Add STYLE_CLASS_RICH_LIST

danrabbit:
This class is currently used in Gtk.Inspector, Gtk.WidgetFactory, and Gtk.Demo. It sounds like it may change upstream in future releases, such as to `boxed-list` or something else, but in general having a style class that sets some standard padding/sizes/etc on Gtk.ListBox is useful and we support it in the stylesheet

Todo:
- [x] Add to Demo
- [x] AppData


jeremypw (approved):
Makes sense and the demo works.

I note that the demo is not translated but that is pre-existing.

danrabbit:
Yeah we previously decided not to translate Demo since it's only really used by developers

PR #616, opened 2022-06-16
MessageDialog: Fix large height bug

danrabbit:
Fix code style while we're here

## BEFORE
Screenshot from 2022-06-16 11 15 33

## AFTER
Screenshot from 2022-06-16 11 13 18


marbetschar (approved):
Fixes the issue - and improved the code. Nice, thank you!

Issue #617, opened 2022-06-18
Missing style classes in Granite 7 to port from Gtk3 to Gtk4


marbetschar:
I started working on the [Gtk4 port of Mail](https://github.com/elementary/mail/pull/790) and noticed that the following Gtk3 style classes are not available in Granite. Not sure if this is a bug or intentional - in any case, please advise how to migrate them to Gtk4 / Granite 7:

- [`Gtk.STYLE_CLASS_BACKGROUND`](https://valadoc.org/gtk+-3.0/Gtk.STYLE_CLASS_BACKGROUND.html)
- [`Gtk.STYLE_CLASS_SIDEBAR`](https://valadoc.org/gtk+-3.0/Gtk.STYLE_CLASS_SIDEBAR.html)
- [`Gtk.STYLE_CLASS_TOOLBAR`](https://valadoc.org/gtk+-3.0/Gtk.STYLE_CLASS_TOOLBAR.html)
- [`Gtk.STYLE_CLASS_INLINE_TOOLBAR`](https://valadoc.org/gtk+-3.0/Gtk.STYLE_CLASS_INLINE_TOOLBAR.html)

danrabbit:
Oops yeah `SIDEBAR` and `BACKGROUND` are definitely an oversight and need to be added

action bars previously using `INLINE_TOOLBAR` should use `FLAT` instead

I don’t think the stylesheet has a style for `.toolbar` and Gtk.Toolbar was removed in Gtk4, so I’d be curious to know how that’s being used so we can figure out how to migrate it

marbetschar:
Regarding `Gtk.STYLE_CLASS_TOOLBAR`: Mail is only using it in the [AttachmentBar.vala](https://github.com/elementary/mail/blob/master/src/MessageList/AttachmentBar.vala#L32) (used to list attachments below a message).

I will test if it makes any visual differences if the class is there - if so, I will hardcode the string value and otherwise simply remove the class alltogether. Would you agree to this course of action? Or do I miss something here?

danrabbit:
@marbetschar ah so for the attachment one, probably either pack that into an actionbar?or set the css name of the flowbox to “actionbar”, whichever one preserves the styles better there. I’m pretty sure the “toolbar” style class is removed completely in the gtk4 stylesheet 

I think the entry one is a workaround for not being able to pack children into an entry, but I think that’s changed in Gtk. That’s also used for cssname as well right? So you could manually set the cssname to “entry” for the same effect

[...]

PR #618, opened 2022-06-20
Added missing style constants (sidebar + background)

marbetschar:
Adds the missing style classes discussed in https://github.com/elementary/granite/issues/617 as constants


danrabbit (approved):
Thanks for adding these!

PR #621, opened 2022-08-28
Constants: add large-icons

danirabbit:
This style class is used by Gtk.Image, so we support it too. See also https://github.com/elementary/stylesheet/pull/1222


tintou:
Apparently this should be set by Gtk when doing `image.icon_size = Gtk.IconSize.LARGE`

danrabbit:
Yeah, but I think it should also be available as a constant so we can set it on image containers like buttons or you might want to set all the images in an actionbar at once or something like that

JoseExposito (approved):
LGTM, thanks for all the hard work @danirabbit ❤️

PR #622, opened 2022-09-07
SimpleSettingsPage: layout text smarter at small widths

danrabbit:
* Wrap titles instead of ellipsize
* Valign switches so they always sit in the top right corner, even with wrapped titles
* Let description text go under the switch

## BEFORE

Screenshot from 2022-09-07 07 39 42

## AFTER

Screenshot from 2022-09-07 07 33 54

Screenshot from 2022-09-07 07 40 34


jeremypw (approved):
Makes sense.

PR #624, opened 2022-09-10
HeaderLabel: add optional body text

danirabbit:
As I'm working with more settings panes in particular I'm finding that I want to add description text below certain headers, but because of the built in extra spacing in headers it makes the layout kind of awkard. I'd like to be able to have spacing and sizing kind of specific for this case so we always have nice typography in this kinds of headers:

## BEFORE
Screenshot from 2022-09-10 16 01 51
Screenshot from 2022-09-10 16 01 44

## AFTER
With a little CSS: https://github.com/elementary/stylesheet/pull/1228

Screenshot from 2022-09-10 16 14 28

### DIFF
--- a/lib/Widgets/HeaderLabel.vala
+++ b/lib/Widgets/HeaderLabel.vala
@@ -17,7 +17,7 @@ public class Granite.HeaderLabel : Gtk.Widget {
      * Optional secondary label string displayed below the header
      */
     public string? secondary_text {
-        get; construct set; default = null; 
+        get; construct set; default = null;
     }
 
     /**
--- a/lib/Widgets/HeaderLabel.vala
+++ b/lib/Widgets/HeaderLabel.vala
@@ -42,7 +42,7 @@ public class Granite.HeaderLabel : Gtk.Widget {
         var label_widget = new Gtk.Label (label) {
             xalign = 0
         };
-        label_widget.add_css_class (Granite.STYLE_CLASS_H4_LABEL);
+        label_widget.add_css_class ("heading");
         label_widget.set_parent (this);
 
         var secondary_label = new Gtk.Label (secondary_text) {
--- a/lib/Widgets/HeaderLabel.vala
+++ b/lib/Widgets/HeaderLabel.vala
@@ -8,17 +8,29 @@
  * HeaderLabel is a start-aligned {@link Gtk.Label} with the Granite H4 style class
  */
 public class Granite.HeaderLabel : Gtk.Widget {
+
     public string label { get; construct set; }
+
+    public string? body {
+        get; construct set; default = null; 
+    }
 
     /**
      * Create a new HeaderLabel
      */
-    public HeaderLabel (string label) {
-        Object (label: label);
+    public HeaderLabel (string label, string? body = null) {
+        Object (
+            body: body,
+            label: label
+        );
+    }
+
+    class construct {
+        set_css_name ("header");
     }
 
     static construct {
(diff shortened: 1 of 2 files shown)


lib/Widgets/HeaderLabel.vala
```diff
@@ -8,32 +8,69 @@
...
+    public string? secondary_text {
+        get; construct set; default = null;
+    }
+
     /**
      * Create a new HeaderLabel
      */
-    public HeaderLabel (string label) {
-        Object (label: label);
+    public HeaderLabel (string label, string? secondary_text = null) {
```
tintou, line 26: that's an API break, a solution would be to have:
```vala
public HeaderLabel.with_secondary (string label, string secondary_text) {
```

lib/Widgets/HeaderLabel.vala
```diff
@@ -8,32 +8,69 @@
...
 
     construct {
         var label_widget = new Gtk.Label (label) {
             xalign = 0
         };
-        label_widget.add_css_class (Granite.STYLE_CLASS_H4_LABEL);
+        label_widget.add_css_class ("heading");
         label_widget.set_parent (this);
 
+        var secondary_label = new Gtk.Label (secondary_text) {
```
tintou, line 48: why not create it on-demand within the property setter?
```vala
public string? secondary_text {
        get;
        construct set {
            if (secondary_label != null) {
                if (value == null) {
                    secondary_label.unparent ();
                    secondary_label = null;
                } else {
                    secondary_label.label = value;
                }
            } else if (value != null) {
                secondary_label = new Gtk.Label (secondary_text) {
                    wrap = true,
                    xalign = 0
                };
                secondary_label.add_css_class (Granite.STYLE_CLASS_DIM_LABEL);
                secondary_label.add_css_class (Granite.STYLE_CLASS_SMALL_LABEL);
            }
        }
        default = null;
    }
```

tintou (changes requested):

lib/Widgets/HeaderLabel.vala
danirabbit: Ah yeah good point. In this case, another constructor probably isn't any better than just initializing with property:

```vala
new HeaderLabel.with_secondary (
    "A Header String",
    "A secondary description string"
);
```
vs
```vala
new HeaderLabel ("A Header String") {
    secondary_text = "A secondary description string"
};
```

lib/Widgets/HeaderLabel.vala
```diff
@@ -8,32 +8,73 @@
  * HeaderLabel is a start-aligned {@link Gtk.Label} with the Granite H4 style class
  */
 public class Granite.HeaderLabel : Gtk.Widget {
+    /**
+     * The primary header label string
+     */
     public string label { get; construct set; }
 
+    private Gtk.Label secondary_label;
```
tintou, line 16:
```suggestion
    private Gtk.Label? secondary_label = null;
```

lib/Widgets/HeaderLabel.vala
```diff
@@ -8,32 +8,73 @@
...
+    private string _secondary_text;
+    /**
+     * Optional secondary label string displayed below the header
+     */
+    public string? secondary_text {
+        get {
+            return _secondary_text;
```
tintou, line 23:
```suggestion
            return secondary_label != null ? secondary_label.label : null;
```
that way you don't even need `private string _secondary_text;`

tintou (changes requested):

tintou (approved):

PR #632, opened 2023-02-21
Init: add ensure type

danirabbit:
Does the least to give a way for Adwaita to detect that Granite has been initialized so that it can disable styles i.e. for Epiphany

### DIFF
--- a/lib/Init.vala
+++ b/lib/Init.vala
@@ -10,6 +10,7 @@ namespace Granite {
      * Initializes Granite.
      * If Granite has already been initialized, the function will return.
      * Makes sure translations and types for Granite are set up properly.
+     * @since 7.2.0
      */
     public void init () {
         if (initialized) {
--- a/lib/Init.vala
+++ b/lib/Init.vala
@@ -4,12 +4,19 @@
  */
 
 namespace Granite {
-    static bool initialized = false;
+    private static bool initialized = false;
 
-    internal void init () {
+    /*
+     * Initializes Granite.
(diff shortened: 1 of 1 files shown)


lib/Init.vala
```diff
@@ -4,13 +4,20 @@
...
 namespace Granite {
-    static bool initialized = false;
+    private static bool initialized = false;
 
-    internal void init () {
+    /*
+     * Initializes Granite.
+     * If Granite has already been initialized, the function will return.
+     * Makes sure translations and types for Granite are set up properly.
+     */
```
Exalm, line 13: Missing `Since:`?

lib/Init.vala
```diff
@@ -4,13 +4,20 @@
...
+    public void init () {
         if (initialized) {
             return;
         }
 
+        typeof (Granite.Settings).ensure ();
```
Exalm, line 19: You're already inside Granite namespace, so jut
```suggestion
        typeof (Settings).ensure ();
```
would have worked. Might be a matter of code style tho, not sure.

Exalm (approved):
LGTM otherwise

lib/Init.vala
danirabbit: Yeah I just want to make sure we disambiguate between GLib.Settings just in case!

Exalm:
Works fine on libadwaita side.

tintou (approved):
One might revisit this init function to use a GLib.Once but that's just a nit

PR #634, opened 2023-02-22
Release 7.2.0

danirabbit:
Adwaita is waiting on this change to implement Granite detection so they can skip loading styles in apps like Ephy


lenemter (approved):

davidmhewitt:
Would be nice to get #636 in to this release as well so that the public API can be annotated.

PR #639, opened 2023-03-15
Add granite styles as a gresource

danirabbit:
Start moving Granite-specific styles into the library


Marukesu (approved):
Code looks good, but it's seems that there's some missing styles in the granite demo when using Adwaita as the theme.

danirabbit:
Yeah for sure, it's not fully complete, just better than it was haha

PR #644, opened 2023-04-07
Init: load gresourced styles if present

### LINKED ISSUE #643: Adw apps don't automatically load styles with Granite.init ()
### What Happened?

Adwaita has a feature that will automatically load stylesheet gresources named `style.css` or `style-dark.css`, but this is disabled when Granite is initialized, meaning apps like font-viewer lose their custom CSS that the app expects.

https://gitlab.gnome.org/GNOME/libadwaita/-/commit/d671a1f4ef76968a5b62b20557ea28628905a58b

### Steps to Reproduce

Build a LibAdwaita app using Granite.init such as https://github.com/elementary/font-viewer/pull/6

See that the app's custom styles are missing

### Expected Behavior

We should probably just provide this same feature 

### OS Version

7.x (Horus)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
Fixes #643

Try to find a `style.css` file and if it exists, add it as an `APPLICATION` priority

### DIFF
--- a/lib/Init.vala
+++ b/lib/Init.vala
@@ -37,8 +37,22 @@ namespace Granite {
         if (css_provider == null) {
             css_provider = new Gtk.CssProvider ();
             css_provider.load_from_resource ("/io/elementary/granite/Granite.css");
+
+            var base_path = Application.get_default ().resource_base_path;
+            if (base_path != null) {
+                var base_uri = "resource://" + base_path;
+                var base_file = File.new_for_uri (base_uri);
+
+                init_provider_from_file (css_provider, base_file.get_child ("style.css"));
+            }
         }
 
         Gtk.StyleContext.add_provider_for_display (display, css_provider, Gtk.STYLE_PROVIDER_PRIORITY_THEME);
+    }
+
+    private static void init_provider_from_file (Gtk.CssProvider provider, File file) {
+        if (file.query_exists ()) {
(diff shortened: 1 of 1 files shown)


Marukesu:
I believe this would make epiphany load they `style.css`, maybe `style-elementary.css` like epiphany does, or the `Application.css` that's already used by elementary apps should be a better name.

Exalm:
It would, yeah. Not nice that there is a style clash, but not a lot we can do here. :(

Any reason not to have a dark variant as well? libadwaita does have that too

danirabbit:
@Marukesu @Exalm Yeah maybe we can check for `Application.css` or `style-elementary.css` first and then try `style.css` if it isn't available? The problem is that for apps like font viewer, we need the padding etc that they provide in their `style.css`

@Exalm I just wanted to get the minimum in here first and then come back in a separate branch for dark style loading since it introduces quite a bit more complexity

Marukesu:
In the font viewer case, could we patch the gresource file to alias the `style.css` to the expected name by granite?

danirabbit:
Ah yeah that's true. In that case, I'll change this to look for `Application.css` which is the convention we've been using so far in our apps

Marukesu (approved):
LGTM.

PR #649, opened 2023-05-10
Settings Urls

Antolius:
Implements the Granite part of [Settings Urls #190](https://github.com/elementary/docs/issues/190) issue from the docs repo.

Currently covers:
- Location Services
- Online Accounts
- Network
- App Permissions
- Notifications
- Sound Input
- Custom Shortcuts

Links are showcased in a new demo app view as links:

Screenshot from 2023-05-11 21 56 29

### DIFF
--- a/lib/Constants.vala
+++ b/lib/Constants.vala
@@ -184,78 +184,44 @@ namespace Granite {
     public const int TRANSITION_DURATION_OPEN = 250;
 
     /**
-    * Link to open Applications → Defaults settings page
+    * Deep links to specific Settings pages.
     */
-    public const string LINK_SETTINGS_APPLICATIONS_DEFAULTS = "settings://applications/defaults";
+    namespace SettingsUri {
 
-    /**
-    * Link to open Applications → Startup settings page
-    */
-    public const string LINK_SETTINGS_APPLICATIONS_STARTUP = "settings://applications/startup";
+         /**
+         * Link to open Security & Privacy → Location Services settings page
+         */
+         public const string LOCATION = "settings://privacy/location";
 
-    /**
-    * Link to open Applications → Permissions settings page
-    */
-    public const string LINK_SETTINGS_APPLICATIONS_PERMISSIONS = "settings://applications/permissions";
+         /**
+         * Link to open Online Accounts settings page
+         */
+         public const string ONLINE_ACCOUNTS = "settings://accounts/online";
 
-    /**
-    * Link to open Desktop → Wallpaper settings page
-    */
-    public const string LINK_SETTINGS_DESKTOP_APPEARANCE_WALLPAPER = "settings://desktop/appearance/wallpaper";
+         /**
+         * Link to Network settings page
+         */
+         public const string NETWORK = "settings://network";
 
-    /**
-    * Link to open Desktop → Appearance settings page
-    */
-    public const string LINK_SETTINGS_DESKTOP_APPEARANCE = "settings://desktop/appearance";
+         /**
+         * Link to open Applications → Permissions settings page
+         */
+         public const string PERMISSIONS = "settings://applications/permissions";
 
-    /**
-    * Link to open Desktop → Dock & Panel settings page
-    */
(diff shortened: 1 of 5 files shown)


lib/Constants.vala
```diff
@@ -182,4 +182,80 @@ namespace Granite {
      * Transition duration when a widget opens, reveals more content, or enters the screen
      */
     public const int TRANSITION_DURATION_OPEN = 250;
+
+    /**
+    * Link to open Applications → Defaults settings page
+    */
+    public const string LINK_SETTINGS_APPLICATIONS_DEFAULTS = "settings://applications/defaults";
```
Antolius, line 189: Would it be better to put these constants under a dedicated namespace? Something like `Granite.Links`? (`Granite.Settings` is already "taken" by the settings class.)

lib/Constants.vala
```diff
@@ -182,4 +182,80 @@ namespace Granite {
...
+
+    /**
+    * Link to open Security & Privacy → Location Services settings page
+    */
+    public const string LINK_SETTINGS_PRIVACY_LOCATION = "settings://privacy/location";
+
+    /**
+    * Link to open Security & Privacy → Housekeeping settings page
+    */
+    public const string LINK_SETTINGS_PRIVACY_TRASH = "settings://privacy/trash";
```
Antolius, line 249: Some times URLs have diverged from settings page titles shown in UI. In these cases should the constant name align with URL (like here) or UI title (in which case this should be renamed to `LINK_SETTINGS_PRIVACY_HOUSEKEEPING`)?

Personally I like URL-like naming better, since those are less likely to change over time, while we may want to tweak UI titles more often.

lib/Constants.vala
danirabbit: Yeah I think that would be better. My suggestion would be `Granite.SettingsUri` so that using it in practice would be like `new Gtk.UriLauncher (Granite.SettingsUri.LOCATION)` which seems consistent and succinct

lib/Constants.vala
danirabbit: I think we should just do the final page name since the way pages are organized could possibly change in the future. So as in my previous example `Granite.SettingsUri.LOCATION` or `Granite.SettingsUri.ONLINE_ACCOUNTS`

danirabbit:
Hey thanks for starting in on this! I left a couple responses to the comments you left previously.

Additionally I think we should probably keep this pretty tightly scoped to settings that we're sure are useful to developers so we don't end up with a bunch of extra un-useful constants to dig through. From the top of my head the things that seem useful are:
- Location
- Online accounts
- Network
- App permissions
- Notifications?
- Sound Input?

I'm not sure why, for example, an app other than Files would be linking to Housekeeping settings.

Antolius:
> Additionally I think we should probably keep this pretty tightly scoped to settings that we're sure are useful to developers so we don't end up with a bunch of extra un-useful constants to dig through.

I must admit I went into this one with a bit of a completionist mindset. 😅

But yeah, we can definitely start with a narrow set of links that seem more useful, and potentially add new ones down the line if there's interest.

Antolius:
Ok, I checked the 3rd party apps in the AppCenter, and in addition to Nimbus requiring network, we have:

1. [Eddy linking to `settings://privacy`](https://github.com/donadigo/eddy/blob/master/src/PackageListView.vala#L268)
2. [Clips linking to `settings://input/keyboard/shortcuts`](https://github.com/hezral/clips/blob/master/src/settings_view.py#L284)
3. [Iridium linking to `settings://network`](https://github.com/avojak/iridium/blob/master/src/Widgets/NetworkInfoBar.vala#L47)
4. [Stashed linking to `settings://input/keyboard/shortcuts`](https://github.com/hezral/stashed/blob/main/src/window.py#L321)
5. [Home linking to `settings://network`](https://github.com/manexim/home/blob/master/src/views/Overview.vala#L138)

So yes, @lenemter, your pull request on keyboard settings is timed perfectly! I suggest adding `settings://input/keyboard/shortcuts/custom` to the list of constants.

danirabbit (approved):
Nice job! Reasonable list, I think. Works as expected :)

PR #651, opened 2023-05-12
Constants: Add since versions to new URI values

davidmhewitt:
Tidies up the indentation while we're here


danirabbit (approved):
Looks good thanks for fixing that!

PR #665, opened 2023-11-09
Toast: Add withdraw method

danirabbit:
It looks like in some cases we want to be able to withdraw the currently sent toast and we were relying on accessing the revealer directly for that. This adds a proper withdraw function that also stops the timeout.

https://github.com/search?q=org%3Aelementary%20toast.reveal_child%20%3D%20false%3B&type=code


lib/Widgets/Toast.vala
```diff
@@ -171,4 +171,14 @@ public class Granite.Toast : Gtk.Widget {
         stop_timeout ();
         start_timeout ();
     }
+
+    /**
+     * Withdraws the currently displayed Toast
+     * @since 7.4.0
+     */
+    [Version (since = "7.4.0")]
+    public void withdraw_notification () {
```
lenemter, line 180: Why not call it simply `withdraw`?

danirabbit: I think the idea was to model the API after Gtk.Application? But yeah I'm not opposed to just calling it "withdraw". I'll update it

lenemter (approved):

PR #676, opened 2024-01-07
Make granite-demo optional

ryonakano:
We don't need demo in the Flatpak Platform for example, so disabling demo when building for Flatpak should lessen its size and make build time faster a little


danirabbit (approved):
Makes sense to me!

PR #677, opened 2024-01-09
Toast: add dismiss signal with dismissal reason

danirabbit:
Always send the `closed` signal when Toasts are closed manually or automatically, except when the default action is triggered. Also, add a close reason in case we want to respond differently based on whether we closed manually, expired, or withdrawn.

Not sure if we need the close reason actually. afaict nothing is connecting to this signal anywhere. The motivation for this PR is that Desktop settings was watching for the toast revealer to be closed https://github.com/elementary/switchboard-plug-pantheon-shell/pull/377/commits/c031ada33df0d64adda8ec6a90e2518dc13160e5. Happy to exclude that if it seems unnecessary


lib/Widgets/Toast.vala
```diff
@@ -16,11 +16,21 @@
...
+        DISMISSED = 2,
+        WITHDRAWN = 3
+    }
 
     /**
-     * Emitted when the Toast is closed by activating the close button
+     * Emitted when the Toast is closed
      */
-    public signal void closed ();
+    public signal void closed (int reason_id);
```
jeremypw, line 33: Any reason to use `int` rather than `uint` as this is an `enum`?

jeremypw (commented):
Just a minor query otherwise looks good.

lib/Widgets/Toast.vala
davidmhewitt: Better still, the type should probably be `CloseReason`.

```suggestion
    public signal void closed (CloseReason reason);
```

davidmhewitt:
> Note that adding a parameter to the "close" signal is an API break (meaning that all the code using it will break without being recompiled)

Yeah, this this case, it may be better to add a new signal (`vanished` maybe), and mark `closed` as deprecated. Because not only does it break the API contract, it's also a change in behaviour with the `closed` signal now being fired for other reasons when it wasn't previously.

danirabbit:
Added a new signal `dismissed`, after https://valadoc.org/libadwaita-1/Adw.Toast.dismissed.html and deprecated the old signal

davidmhewitt (approved):

PR #678, opened 2024-01-10
Placeholder: resize based on parent

### LINKED ISSUE #664: Responsive placeholder
### Problem

Granite.Placeholder is great as a welcome screen or as a placeholder in large lists, but for smaller lists in dialogs and popovers it is way too big

### Proposal

An alternate, more compact style. Maybe we can do this automatically if the placeholder is inside a popover or dialog and/or with a style class?

### Prior Art (Optional)

Screenshot from 2023-10-24 16 27 17

In this popover in Mail we make our own custom placeholder because Granite's is too big

danirabbit:
Fixes #664 

Screenshot from 2024-01-10 14 08 19

### DIFF
--- a/demo/Views/WelcomeView.vala
+++ b/demo/Views/WelcomeView.vala
@@ -38,12 +38,53 @@ public class WelcomeView : Gtk.Box {
             icon = new ThemedIcon ("edit-find-symbolic")
         };
 
+        var listbox = new Gtk.ListBox () {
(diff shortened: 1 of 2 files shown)


jeremypw (approved):
Works nicely!

Not sure whether the entry should be active - or if so whether something should happen if you type in it, but that can be changed another PR if desired.  This PR fulfils its aim.

PR #692, opened 2024-03-01
Add BackButton

leolost2605:
Introduce a BackButton that automatically lables itself correctly with the preceding page and navigates back when clicked. Useful for elementary/switchboard#300 as well as in videos, appcenter, etc.

This will introduce an Adw dependency with minimum required version 1.4 which will break building granite on OS 7


leolost2605:
It now won't hide when there is no preceding page since that caused problem with it not being mapped again because it's not visible so that it won't recheck whether anything changed. It could still be implemented if needed but since there's usually always exactly one page that's the first and we know which one it is (Home on appcenter, category on settings, welcome on videos) I don't think it's worth the effort.

lib/Widgets/BackButton.vala
```diff
@@ -0,0 +1,62 @@
...
+        var navigation_page = (Adw.NavigationPage) get_ancestor (typeof (Adw.NavigationPage));
+
+        if (navigation_view == null) {
+            return;
+        }
+
+        var previous_page = navigation_view.get_previous_page (navigation_page);
+
+        if (previous_page != null) {
+            label = previous_page.title;
```
alice-mkh, line 51: Note that page label can and will change.

The logic in libadwaita's own back buttons is also a lot more complicated because navigation views can be nested, and can also be structured like:

navigation split view
  sidebar: navigation view
  content: navigation view

and that situation _is_ handled correctly

alice-mkh: Ah, tho the latter involves private api - forgot about that

leolost2605: > Note that page title can and will change.

Fixed although I think that would be kinda questionable UX while the page itself is not visible?

> The logic in libadwaita's own back buttons is also a lot more complicated because navigation views can be nested, and can also be structured like:
> 
> navigation split view
> sidebar: navigation view
> content: navigation view
> 
> and that situation is handled correctly

I'm guessing that if the button is on the deepest view and there is no page left it still appears and instead navigates one above? 
I think for now and for our use case it might be enough to leave that out (since we also don't hide when there is no preceding page)

alice-mkh: > Should be fixed although I think that would be kinda questionable UX while the page itself is not visible?

Questionable yes, but I mean the property is writable, so...

> I'm guessing that if the button is on the deepest view and there is no page left it still appears and instead navigates one above?

Correct.

> I think for now and for our use case it might be enough to leave that out (since we also don't hide when there is no preceding page)

As you prefer

leolost2605: Regarding the nested part cc @danirabbit ?

danirabbit:
I think @tintou has some thoughts about adding Adw as a dep here in Granite

lib/Widgets/BackButton.vala
```diff
@@ -0,0 +1,68 @@
...
+    construct {
+        var image = new Gtk.Image.from_icon_name ("go-previous-symbolic");
+
+        var label_widget = new Gtk.Label ("");
+
+        var box = new Gtk.Box (HORIZONTAL, 0);
+        box.append (image);
+        box.append (label_widget);
+
+        child = box;
```
danirabbit, line 29:
```suggestion
        action_name = "navigation.pop";
        child = box;
        tooltip_markup = Granite.markup_accel_tooltip ({"<alt>Left"});
```

lib/Widgets/BackButton.vala
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright 2024 elementary, Inc. (https://elementary.io)
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+/**
+ * BackButton is meant to be used in headers to navigate in
+ * {@link Adw.NavigationView}s. It will automatically detect when under a
+ * {@link Adw.NavigationView} and label itself with the title of the preceding page.
```
danirabbit, line 9:
```suggestion
 * {@link Adw.NavigationView}.
 *
 * By default `action_name` is set to `navigation.pop`
```

lib/Widgets/BackButton.vala
```diff
@@ -0,0 +1,68 @@
...
+ */
+[Version (since = "7.5.0")]
```
danirabbit, line 11:
```suggestion
[Version (since = "7.7.0")]
```

lib/Widgets/BackButton.vala
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+
+    private void on_click () {
+        var navigation_view = (Adw.NavigationView) get_ancestor (typeof (Adw.NavigationView));
+
+        if (navigation_view != null) {
+            navigation_view.pop ();
+        }
+    }
```
danirabbit, line 67:
```suggestion
    }
```

meson.build
```diff
@@ -50,6 +50,7 @@ libgranite_deps = [
     dependency('glib-2.0', version: '>=' + glib_min_version),
     dependency('gobject-2.0', version: '>=' + glib_min_version),
     dependency('gtk4', version: '>=4.4'),
+    dependency('libadwaita-1', version: '>=1.4'),
```
danirabbit, line 53:
```suggestion
```

lib/Widgets/BackButton.vala
```diff
@@ -0,0 +1,68 @@
...
+ * {@link Adw.NavigationView} and label itself with the title of the preceding page.
+ */
+[Version (since = "7.5.0")]
+public class Granite.BackButton : Gtk.Button {
+    /**
+     * A manually set label when used outside of {@link Adw.NavigationView}
+     */
+    public new string label { get; set; }
+
+    private Binding? title_binding = null;
```
danirabbit, line 18:
```suggestion
```

lib/Widgets/BackButton.vala
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * BackButton is meant to be used in headers to navigate in
+ * {@link Adw.NavigationView}s. It will automatically detect when under a
+ * {@link Adw.NavigationView} and label itself with the title of the preceding page.
+ */
+[Version (since = "7.5.0")]
+public class Granite.BackButton : Gtk.Button {
+    /**
+     * A manually set label when used outside of {@link Adw.NavigationView}
```
danirabbit, line 14:
```suggestion
     * Text of the label inside of #this
```

lib/Constants.vala
```diff
@@ -11,6 +11,7 @@ namespace Granite {
     /**
      * Style class for shaping a {@link Gtk.Button}
      */
+    [Version (deprecated = true, deprecated_since = "7.5.0", replacement = "Granite.BackButton")]
```
danirabbit, line 14:
```suggestion
    [Version (deprecated = true, deprecated_since = "7.7.0", replacement = "Granite.BackButton")]
```

README.md
```diff
@@ -15,6 +15,7 @@ You'll need the following dependencies:
 * libgee-0.8-dev
 * libgirepository1.0-dev
 * libgtk-4-dev >= 4.4.0
+* libadwaita-1-dev >= 1.4
```
danirabbit, line 18:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Install Dependencies
         run: |
           apt update
-          apt install -y meson gobject-introspection libgee-0.8-dev libgirepository1.0-dev libgtk-4-dev sassc valac
+          apt install -y meson gobject-introspection libgee-0.8-dev libgirepository1.0-dev libgtk-4-dev libadwaita-1-dev sassc valac
```
danirabbit, line 26:
```suggestion
          apt install -y meson gobject-introspection libgee-0.8-dev libgirepository1.0-dev libgtk-4-dev sassc valac
```

danirabbit (changes requested):
It's slightly less magical, but I think it's still worth it to merge this without the adw dep. NavigationView adds a `navigation.pop` action we can use to at least make the button function and then all that needs to be done in-app is to set the label.

leolost2605:
@danirabbit feel free to take over and merge, I'm currently really limited on time :/

danirabbit (approved):

danirabbit:
@wpkelso does this sound good to you?

wpkelso:
I'm in favor, but it does feel a little weird to have a technically half-implemented widget sitting around in the library. I think this would ideally be joined by a first-party Navigation View, but I don't know if we have the bandwidth to build and maintain that right now.

danirabbit:
It's slightly more implemented than the style class was though :sweat_smile:

wpkelso:
> It's slightly more implemented than the style class was though 😅

Agreed; I put it out more as a consideration for the future, rather than a blocker.

PR #693, opened 2024-03-01
MessageDialog: Set last button as focus widget

leolost2605:
Less hacky now


danirabbit:
Ah yeah this feels janky. Gtk.Dialog is deprecated anyways, so maybe we can solve this along the way when re-implementing button adding

danirabbit (approved):
Smart solution!

leolost2605:
Btw I'm not sure here but I remember having some symbol lookup errors with programs that weren't recompiled so this might be a full API break? I'll do some research later :)

PR #694, opened 2024-03-01
Use Gtk.Image, deprecate style class, demo

danirabbit:
* Replace the back button in Demo
* Deprecate the back-button style class
* Since we're providing a widget, use an image for the icon. Saves us some questionable CSS
* Allow setting the label manually in case there's some situations where we use this outside of Adw.NavigationView :shrug:


leolost2605 (approved):
Thank you!

PR #704, opened 2024-03-20
StyleManager: set accent color from portal

danirabbit:
Allows us to use system accent color without the multiple stylesheets hack

Screenshot from 2025-03-21 14 42 17
Screenshot from 2025-03-21 14 42 29


zeebok (approved):

zeebok (approved):

lib/Widgets/Settings.vala
```diff
@@ -96,6 +115,40 @@ namespace Granite {
...
+
+                portal.setting_changed.connect ((scheme, key, value) => {
+                    if (scheme == "org.freedesktop.appearance" && key == "accent-color") {
+                        update_color (value);
+                    }
+                });
+
+                return;
+            } catch (Error e) {
+                debug (e.message);
```
ryonakano, line 137: Maybe warning instead of debug?

lib/Widgets/Settings.vala
```diff
@@ -96,6 +115,40 @@ namespace Granite {
...
+            accent_color = "#3689e6";
+        }
+
+        private void update_color (GLib.Variant color) {
+            double red, green, blue;
+            color.get ("(ddd)", out red, out green, out blue);
+
+            Gdk.RGBA rgba = {(float) red, (float) green, (float) blue, 1};
+            accent_color = rgba.to_string ();
+        }
```
ryonakano, line 150: I think we should keep this method just parse Variant representation to string representation of color and substitute to `accent_color` in the caller side instead.

```vala
        private string parse_color (GLib.Variant color) {
            double red, green, blue;
            color.get ("(ddd)", out red, out green, out blue);

            Gdk.RGBA rgba = {(float) red, (float) green, (float) blue, 1};
            return rgba.to_string ();
        }
```

ryonakano:
@danirabbit Hmm, could you tell me how to test this? Maybe using an label with .accent style in a Flatpak app and then changing accent color in System Settings? :thinking:

danirabbit:
@ryonakano it should work with the demo. I can’t remember if this needs the new granite stylesheet or if it works with the current stylesheet though. You might need to set the stylesheet in inspector to Granite

danirabbit:
Will need to bump the “since” line here because we already released 7.5.0

lib/Styles/_label.scss
```diff
@@ -1,3 +1,11 @@
+.accent {
+    color: #{'mix(@accent_color, @fg_color, 0.27)'};
```
alice-mkh, line 2: I don't think this will produce very good results for e.g. yellow. In libadwaita I had _much_ better results with oklch colorspace (tho that would require you to update gtk once 4.16 is out)

alice-mkh: Well, that or calculate it programmatically ofc, then you don't need to wait.

lib/Widgets/Settings.vala
```diff
@@ -36,6 +36,25 @@ namespace Granite {
...
         }
 
+        private string? _accent_color = null;
+
+        /**
+         * The theme accent color chosen by the user
+         * @since 7.5.0
+         */
+        [Version (since = "7.5.0")]
+        public string accent_color {
```
alice-mkh, line 46: Any reason to do string and not Gdk.RGBA?

danirabbit: Probably not. RGBA would probably be the better/more flexible choice. Good call

alice-mkh:
> It currently isn't being updated while running because it looks like our portal isn't sending settings-changed signals?

I see what's happening. It stops working once you open inspector. Looks like while you do support multiple GdkDisplays for the most part, `accent_provider` is not per-display and basically you need separate providers for each.

What does completely fail to pick changes for me is color-scheme instead, at least in calculator.

lib/Init.vala
```diff
@@ -61,7 +61,7 @@ namespace Granite {
 
         Gtk.StyleContext.remove_provider_for_display (display, accent_provider);
         accent_provider.load_from_string ("@define-color accent_color %s;".printf (accent_color));
-        Gtk.StyleContext.add_provider_for_display (display, accent_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+        Gtk.StyleContext.add_provider_for_display (display, accent_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION - 1);
```
alice-mkh, line 64: I'd use THEME + 1 instead

lib/Widgets/Settings.vala
```diff
@@ -96,6 +115,39 @@ namespace Granite {
...
+                portal = Portal.Settings.get ();
+
+                var variant = portal.read (
+                    "org.freedesktop.appearance",
+                    "accent-color"
+                ).get_variant ();
+
+                accent_color = parse_color (variant);
+
+                portal.setting_changed.connect ((scheme, key, value) => {
```
danirabbit, line 129: @alice-mkh maybe it's because I'm on an older version of some library but I'm not getting anything at all here when I change the color in system settings so I can't really test further because it's broken for me all the way at the beginning

lib/Init.vala
danirabbit: I tried THEME + 1 but it didn't work for some reason

alice-mkh: Huh. THat is strange.

lib/Widgets/Settings.vala
alice-mkh: Huh okay. It did work for me, but only until I open inspector due to the global provider issue I mentioned.

danirabbit: Yeah I can't even test that :sweat_smile:

danirabbit:
@ryonakano this should be working with the new style manager if you want to give it a look again :)

lib/Widgets/Settings.vala
```diff
@@ -96,9 +115,46 @@ namespace Granite {
...
+                    if (scheme == "org.freedesktop.appearance" && key == "accent-color") {
+                        accent_color = parse_color (value);
+                    }
+                });
+            } catch (Error e) {
+                warning (e.message);
+
+                // Set a default in case we can't get from system
+                accent_color.parse ("#3689e6");
```
ryonakano, line 140: Doesn't we need `accent_color = Gdk.RGBA ();` here? I guess, otherwise, `accent_color` would remain null and this line could cause null access when `_accent_color` is null and `setup_accent_color ()` is called.

meson.build
```diff
@@ -57,7 +57,7 @@ libgranite_deps = [
     gio_os_dep,
     dependency('glib-2.0', version: '>=' + glib_min_version),
     dependency('gobject-2.0', version: '>=' + glib_min_version),
-    dependency('gtk4', version: '>=4.4'),
+    dependency('gtk4', version: '>=4.12'),
```
ryonakano, line 60: We should update the required gtk4 version in README.md.

ryonakano (commented):
Thanks, I haven't tested yet, but leaving some comments.

lib/Widgets/Settings.vala
```diff
@@ -96,9 +115,47 @@ namespace Granite {
...
+                    if (scheme == "org.freedesktop.appearance" && key == "accent-color") {
+                        accent_color = parse_color (value);
+                    }
+                });
+            } catch (Error e) {
+                warning (e.message);
+
+                // Set a default in case we can't get from system
+                accent_color = Gdk.RGBA ();
+                accent_color.parse ("#3689e6");
```
ryonakano, line 141: Sorry, I tested this fallback by throwing a new error in L128 and found this fallback does not work as expected; this sets anything to accent_color, which results it to be "rgb(0, 0, 0)".

I confirmed the following code works as expected instead:

```suggestion
                _accent_color = Gdk.RGBA ();
                _accent_color.parse ("#3689e6");
```

ryonakano (commented):
One tiny point in an error case, otherwise looks good and works for me.

ryonakano (approved):

PR #713, opened 2024-03-25
SwitchModelButton: make switch not focusable

danirabbit:
The button itself is already keyboard focusable so we don't need the switch to separately be keyboard focusable. This makes keyboard navigation faster


leolost2605 (approved):
Makes sense 
However one thing I noticed it that the menuitem doesn't get highlighted when navigating via keyboard, maybe because of #707?
But since that's not related to this PR approve :)

danirabbit:
@leolost2605 Granite's style sheet does have focus styles for model items, but io.elementary.stylesheet doesn't, so unless you switch to the Granite stylesheet in inspector you won't notice it

PR #715, opened 2024-04-01
HeaderLabel: wrap title, halign FILL

danirabbit:
Having this set to `halign = START` by default ends up with us overriding this in a bunch of places. All other widgets default to `FILL`

wrap the title while we're here


ryonakano:
Maybe we need to update the documentation comment here?

```vala
/**
 * HeaderLabel is a start-aligned {@link Gtk.Label} with the Granite H4 style class
 */
```

danirabbit:
@ryonakano updated! The label itself is still start aligned so I left that bit in

ryonakano (approved):
Can't check everywhere we use HeaderLabel, but at least it works as expected in my Reco app (I set `max_width_chars` to 50 for example here):

Screenshot from 2024-05-02 08-48-05

PR #722, opened 2024-06-16
Init: set accent color from portal feedback

ryonakano:
Review feedback for #704

I personally think this would even better for readability, but it's fine to close or modify :)

- Keep the `update_color` method just parses Variant into string representation of color
  - This also makes sure that `accent_color` is only set in the `setup_accent_color` method
- Use `warning` for error message
- Set the fallback color in the catch statement instead of returning the try statement


danirabbit (approved):
This makes and I agree is better. Thank you!

PR #725, opened 2024-06-19
HeaderLabel: set accessible description

danirabbit:
Properly set secondary text as a description


zeebok (approved):
This works as advertised though I wonder if it is better for the secondary text to become the description of the primary label, so it would be Name: Username and Description: Must be at least 8 characters long. Not sure how it all appears for the end user.

danirabbit:
@zeebok the labels themselves aren't actually available to the screen reader. So what happens is you have to set a mnemonic widget and then when that widget gets focus it reads out the accessible name, the type of widget, then the accessible description. So for example if the mnemonic widget was a text entry it would say:

"Username Text. Must be at least 8 characters long"

zeebok:
Gotcha. Looking via the Inspector it has the secondary text as a different label that is a sibling of the primary and so each one has its own name in the accessibility data tab, so I wasn't sure if that is how it will appear with the screen reader

danirabbit:
Oh hm yeah I'm not sure how the whole a11y tree stuff works, but I'll take a look at it!

danirabbit:
I'm gonna go ahead and merge this for now since after double checking with Florian it seems the label order is correct here

PR #730, opened 2024-09-18
Introduce Granite.Bin for subclassing

leolost2605:
We are now quite often doing an extension of widget and then setting the layout manager type to bin and doing all the other stuff with manually parenting and unparenting, so instead introduce Granite.Bin which holds only a single widget and is intended for subclassing. It's pretty much the same as [Adw.Bin](https://valadoc.org/libadwaita-1/Adw.Bin.html)


danirabbit (approved):
Seems reasonable to me!

PR #731, opened 2024-09-18
Introduce a style manager

leolost2605:
This moves the provider management etc. from the init function to a separate class StyleManager. 

This class now also automatically handles adjusting the theme to system preference and allowing the application to override this (which was my motivation since that's a bunch of boiler plate code that then could be removed from the applications startup functions).

Since more stuff is going to have to be done here (see e.g. https://gitlab.gnome.org/GNOME/gtk/-/issues/6821) this might be a good way to go now. 

~This currently is a global style manager for all displays but maybe we should follow Adw.StyleManager and make it per display?~ This is now per display since that makes more sense in general


leolost2605:
Regarding the demo: I opted to leave the switch that directly interfaces with gtk settings in the headerbar which can cause a mismatch between e.g. force dark in the stylemanager page and a light mode of the app. This is intended behavior because we shouldn't actually do anything with gtk settings manually but it might confuse people who use the demo.

lib/StyleManager.vala
```diff
@@ -0,0 +1,125 @@
...
+ * A class for managing the style of the application. This handles switching light and dark mode based
+ * based on system preference or application preference (see {@link color_scheme_override}), etc.
+ */
+public class Granite.StyleManager : Object {
+    private static Gtk.CssProvider? base_provider = null;
+    private static Gtk.CssProvider? dark_provider = null;
+    private static Gtk.CssProvider? app_provider = null;
+
+    private static GLib.Once<Granite.StyleManager> instance;
+    public static unowned Granite.StyleManager get_instance () {
```
danirabbit, line 16: Convention for glib/GTK/etc is `get_default`

lib/StyleManager.vala
```diff
@@ -0,0 +1,125 @@
...
+        return instance.once (() => {
+            return new Granite.StyleManager ();
+        });
+    }
+
+    /**
+     * If this is set to NO_PREFERENCE the systems preferred color scheme will be used.
+     * Otherwise the color scheme set here will be used.
+     */
+    public Settings.ColorScheme color_scheme_override { get; set; default = NO_PREFERENCE; }
```
danirabbit, line 26: Maybe just `color_scheme` same as Adw?

lib/StyleManager.vala
```diff
@@ -0,0 +1,125 @@
...
+
+    private static GLib.Once<Granite.StyleManager> instance;
+    public static unowned Granite.StyleManager get_default () {
+        return instance.once (() => {
+            return new Granite.StyleManager ();
+        });
+    }
+
+    /**
+     * If this is set to NO_PREFERENCE the systems preferred color scheme will be used.
```
danirabbit, line 23: I think we need to default to LIGHT if that's the case to preserve the current behavior. Otherwise I think it's technically a breaking change because developers might update and not expect that they need to support dark mode now by default

lib/StyleManager.vala
leolost2605: Tbh I found that a bit ambiguous. Is it the color scheme that currently, actually applies to the application? Is it the override? Is it the system one? It's not that bad but I kinda prefer it more explicitly idk :sweat_smile: 
But we can change it :shrug:

lib/StyleManager.vala
leolost2605: Yeah that probably makes sense will update

danirabbit: Actually thinking about it more, maybe we should go ahead and break and just bump everything to 8.0. I think we have to have a runtime break anyways because of the GNOME 47 release and we'll want to update the GNOME apps we ship

leolost2605:
Fyi I switched it to per display style managers because that just makes more sense and is a lot nicer in code (no more need for lots of foreachs)

danirabbit (changes requested):
Hm so testing this, it doesn't seem to work in the demo and I get:
`** (granite-7-demo:140614): CRITICAL **: 11:57:41.997: granite_style_manager_set_color_scheme_override: assertion 'self != NULL' failed`

danirabbit:
> we shouldn't actually do anything with gtk settings manually but it might confuse people who use the demo.

I think we should probably remove the switch to make sure the demo is working properly and contains only good example code and not stuff we don't want other people to copy

leolost2605:
> Hm so testing this, it doesn't seem to work in the demo and I get: `** (granite-7-demo:140614): CRITICAL **: 11:57:41.997: granite_style_manager_set_color_scheme_override: assertion 'self != NULL' failed`

This should be fixed now, for some reason static construction doesn't work?

demo/Views/StyleManagerView.vala
```diff
@@ -0,0 +1,51 @@
...
+
+        var dont_button = new Gtk.CheckButton.with_label (_("Use system theme")) {
+            active = true
+        };
+
+        var force_light = new Gtk.CheckButton.with_label (_("Force light theme")) {
+            group = dont_button
+        };
+
+        var force_dark = new Gtk.CheckButton.with_label (_("Force dark theme")) {
```
danirabbit, line 18: We actually don't translate the demo since it's only used by developers. Also probably better to be consistent with language here and use "style" which is used in Onboarding and Settings. We don't usually use the word "theme" because historically it means something quite different that we don't support

```suggestion
        var label = new Granite.HeaderLabel ("Visual Style");

        var dont_button = new Gtk.CheckButton.with_label ("Follow system setting") {
            active = true
        };

        var force_light = new Gtk.CheckButton.with_label ("Light") {
            group = dont_button
        };

        var force_dark = new Gtk.CheckButton.with_label ("Dark") {
```

danirabbit (changes requested):
Works as expected, looks good, just a few comments!

danirabbit:
@alice-mkh Any thoughts on how this will/won't interact with Adw in for example Epiphany?

lib/StyleManager.vala
```diff
@@ -0,0 +1,123 @@
...
+ * based on system preference or application preference (see {@link color_scheme_override}), etc.
+ */
+public class Granite.StyleManager : Object {
+    private static Gtk.CssProvider? base_provider = null;
+    private static Gtk.CssProvider? dark_provider = null;
+    private static Gtk.CssProvider? app_provider = null;
+
+    private static HashTable<Gdk.Display, StyleManager>? style_managers_by_displays;
+
+    public static unowned StyleManager get_default () {
```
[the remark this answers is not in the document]
leolost2605: Whoops forget about them :grimacing: 

Updated it :)

lib/StyleManager.vala
leolost2605: Fair enough I updated it :)

danirabbit:
I wanna get this Granite 7.6.0 release out and then we can go ahead and merge this for an 8.0.0 that we don't have to worry about breakage etc and have plenty of time to test

danirabbit (approved):
Let's fucking go

PR #734, opened 2024-10-14
Fix dead link

salim-b:


stsdc (approved):

ryonakano:
Maybe the target branch should have been `main`, not `master`?

salim-b:
> Maybe the target branch should have been `main`, not `master`?

Geez, sorry. I just clicked on the link prefixed with *Home* on this page: https://valadoc.org/granite/index.htm

So that one was outdated, too (I submitted https://github.com/vala-lang/valadoc-org/pull/421 to fix that).

And I resubmitted this PR as https://github.com/elementary/granite/pull/735.

Nevertheless, I'd like to ask: Why is there still a `master` branch when the default one was changed to be `main`? Wouldn't it make sense to delete `master`? (I'm in no way involved in Vala, so personally I don't care.)

danirabbit:
This actually makes sense to fix here too. There's both Granite 6 for GTK3 and Granite 7 for GTK4

PR #735, opened 2024-10-14
Fix dead link (#734)

salim-b:
Same as https://github.com/elementary/granite/pull/734, just applied to the right branch (`main`).


danirabbit (approved):
Thanks for fixing this for both versions of Granite!

PR #747, opened 2025-03-06
Blur API

lenemter:
Requires https://github.com/elementary/gala/pull/2293

Usage examples: https://github.com/elementary/dock/pull/390 and https://github.com/ellie-commons/Jorts/tree/lenemter/use-blur-api

### DIFF
--- a/blur-protocol/meson.build
+++ b/blur-protocol/meson.build
@@ -1,0 +1,41 @@
+dep_scanner = dependency('wayland-scanner', native: true)
+prog_scanner = find_program(dep_scanner.get_variable(pkgconfig: 'wayland_scanner'))
+
+blur_protocol_file = files('pantheon-blur-v1.xml')
+
+pantheon_blur_sources = []
+pantheon_blur_sources += custom_target(
+    'pantheon-blur-client-protocol.h',
+    command: [ prog_scanner, 'client-header', '@INPUT@', '@OUTPUT@' ],
+    input: blur_protocol_file,
+    output: 'pantheon-blur-client-protocol.h',
+)
+
+output_type = 'private-code'
+if dep_scanner.version().version_compare('< 1.14.91')
+    output_type = 'code'
+endif
+pantheon_blur_sources += custom_target(
+    'pantheon-blur-protocol.c',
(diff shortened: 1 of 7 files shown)


danirabbit:
I don't really know about making this a part of Granite. I think it makes sense to use transparent elements for the shell, but I'm not sure about for arbitrary applications

stsdc:
I might use this in Monitor to put labels over charts. Currently it's just semi-transparent.
obraz
The key is to not overuse that, so guidelines will be needed.

stsdc:
If the API is not meant to be used by apps then it shouldn't be here.
Though, I would appreciate to have a such API. I see `Granite.BlurSurface` possible use in Dock, so it might be used for label background, or am I wrong?

lenemter:
@danirabbit I don't think introducing a feature (blur in this case) for one shell element is a good idea, considering it requires a lot of new code. And what's wrong with giving access to background blur to any app? I know it can lead to problems with contrast for example, but since we review every app in AppCenter, we can add guidelines and requirement for it

danirabbit:
@lenemter I think it makes sense to also use it in the panel for busy wallpapers, maybe in future versions of quick settings and applications menu, and there's other shell elements that could use it possibly like alt + tab, but I'm personally 👎 on compositor blur in apps

I think this also introduces an app platform problem in our cross-distro story for AppCenter. If developers start writing apps expecting compositor blur, their apps will be illegible when used in other desktops. I'm also imagining this requires new sandbox holes

At the very least I think we should explore this ourselves in the shell first as part of pantheon-wayland and wait to expose it to app developers until maybe there's an upstreamed compositor blur protocol

danirabbit:
@stsdc this is a window manager API for blurring behind a transparent window. It can't create blurs in between widgets inside your app. It's a compositor effect. 

GTK.CSS does have blur filters for making blurs inside your app. I'm not sure how that works (or doesn't) with animated backgrounds but for example this is used in Warble on the game over screen https://appcenter.elementary.io/com.github.avojak.warble/

danirabbit:
@lenemter that sounds good to me and I think we'll have a better idea of the problems/limitations as well

Issue #749, opened 2025-03-09
Move stylesheets to Granite


wpkelso:
This is meant to be an overview/tracking/planning issue. Individual parts should be attached as sub-issues to this issue.

## Rationale
 
1. Pre-empting the move to platform-handled styling in GTK 5
2. We probably shouldn't be overriding app CSS with a system stylesheet [(Please don't theme our apps)](https://stopthemingmy.app/)
3. Apps built for elementary OS will likely look pretty broken on other operating systems, given that we currently don't have active support for GTK widgets in the Granite stylesheet

## Current & Relevant Work

#703 

https://github.com/elementary/stylesheet/pull/1278

## Current State
- `stylesheet:main` with accent color "blueberry"

![Image](https://github.com/user-attachments/assets/46ec5a2e-4193-4cbd-9a92-4fd40d8cdc9b)

![Image](https://github.com/user-attachments/assets/d8636846-0655-4040-a82b-277d0da0ef2f)

![Image](https://github.com/user-attachments/assets/92979190-49ea-49ae-b9a0-505562145b9b)

- `granite:main` as live (GTK stylesheet excluded)

![Image](https://github.com/user-attachments/assets/173ac2d3-9704-4a2a-9801-f4a01b302534)

- `granite:main` with GTK stylesheet included

![Image](https://github.com/user-attachments/assets/36b9cb5d-8f9e-483c-9dfb-8446559192ca)

![Image](https://github.com/user-attachments/assets/d17aec6c-fe54-4d29-9fac-ddedb8a62f68)

![Image](https://github.com/user-attachments/assets/b2eb3719-fb62-4aed-bb80-37df82bee68d)

danirabbit:
I started a wiki page to record some constraints, norms, etc https://github.com/elementary/granite/wiki/Stylesheet-Targets

PR #751, opened 2025-03-12
Rename files to follow consistent naming scheme

wpkelso:
Laying some groundwork for #749 by introducing consistency to file names


danirabbit (commented):
I'm -1 on this change. I've been trying to do make the scss file name match the Vala objects they belong to or use _ file name when it doesn't match a widget but instead a style class

So Granite/OverlayBar.scss styles Granite.OverlayBar. Gtk/Button.scss styles Gtk.Button

But _card.scss contains styles for Granite.CssClass.CARD

wpkelso:
> ... I've been trying to do make the scss file name match the Vala objects they belong to or use _ file name when it doesn't match a widget but instead a style class

@danirabbit Ah, ok, I've misunderstood the naming scheme then. Given this, a file like `stylesheets/_buttons.scss` would need to be split into `granite/Button.scss` and `granite/FontButton.scss`?

PR #755, opened 2025-03-18
Constants: add button styles

danirabbit:
Screenshot from 2025-03-17 19 35 42

Add new `CssClass` constants for buttons and add them to the Demo. Adding anything to the stylesheet itself doesn't have any effect yet since system CSS overwrites the stylesheet loaded in the library


lib/Styles/Index-dark.scss
```diff
@@ -47,7 +47,7 @@ $warning-icon-color: $BANANA_500;
 @import '_osd.scss';
 
 // Gtk Widgets
-// @import 'Gtk/Index.scss';
+@import 'Gtk/Index.scss';
```
wpkelso, line 50: Gtk stylesheet is uncommented here in the dark stylesheet, but not in the light version. Is that intentional?

lib/Styles/Gtk/Index.scss
wpkelso, line 1: Is this just prep for further refactors?

lib/Styles/Index-dark.scss
danirabbit: Oops no that was a mistake. Good catch

danirabbit:
@wpkelso thanks for catching those other changes. I was originally trying to add in the button styles as well, but since they can't overwrite the system stylesheet it didn't work and I didn't do a good job cleaning up after myself :sweat_smile:

wpkelso (approved):

PR #756, opened 2025-03-18
Use CSS properties defined in :root for exporting colors

wpkelso:
https://docs.gtk.org/gtk4/css-properties.html#non-css-colors

Also adds Latte into the mix, since it seems it never got added

### DIFF
--- a/lib/Styles/_palette.scss
+++ b/lib/Styles/_palette.scss
@@ -35,17 +35,23 @@ $BLUEBERRY_500: #3689e6;
 $BLUEBERRY_700: #0d52bf;
 $BLUEBERRY_900: #002e99;
 
+$GRAPE_100: #e4c6fa;
+$GRAPE_300: #cd9ef7;
+$GRAPE_500: #a56de2;
+$GRAPE_700: #7239b3;
+$GRAPE_900: #452981;
+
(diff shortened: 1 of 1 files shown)


lib/Styles/_palette.scss
```diff
@@ -72,74 +78,84 @@ $BLACK_700: #1a1a1a;
 $BLACK_900: #000;
 
 // Exported
-@define-color STRAWBERRY_100 #{$STRAWBERRY_100};
```
danirabbit, line 75: We can't remove these because that would be an API break. We can only add new API

danirabbit (changes requested):
This doesn't actually do anything in our current version of GTK and adds an error: `(granite-7-demo:23978): Gtk-WARNING **: 10:22:35.858: Theme parser error: Granite.css:1:1-6: Unknown name of pseudo-class` but it doesn't appear to exactly break anything.

Except we can't remove the old color variables because that is an API break and apps use those. We should basically just remove those when we're absolutely ready to break api and bump to Granite 8 (or 9 I guess)

wpkelso:
Looks like the docs I'm referencing are 4 minor versions newer than what we're using 🙃

danirabbit (approved):

PR #758, opened 2025-03-19
Styles: Add compile time option for including Gtk stylesheets

wpkelso:
- Splits Gtk and Granite stylesheets into different top level files
- Adds option to `meson_options` to enable loading the Gtk stylesheet at compile time
- I've probably unintentionally broken an API, but I don't know in what way. It seems to work fine with `granite-7-demo`

### DIFF
--- a/lib/Init.vala
+++ b/lib/Init.vala
@@ -41,6 +41,7 @@ namespace Granite {
 
     private static void register_display (Gdk.Display display) {
         var gtk_settings = Gtk.Settings.get_for_display (display);
+        gtk_settings.gtk_theme_name = "Granite";
         gtk_settings.notify["gtk-application-prefer-dark-theme"].connect (() => {
             set_provider_for_display (display, gtk_settings.gtk_application_prefer_dark_theme);
         });
--- a/lib/Init.vala
+++ b/lib/Init.vala
@@ -70,6 +70,9 @@ namespace Granite {
             if (dark_provider == null) {
                 dark_provider = new Gtk.CssProvider ();
                 dark_provider.load_from_resource ("/io/elementary/granite/Granite-dark.css");
+                #if INCLUDE_GTK
+                    dark_provider.load_from_resource ("/io/elementary/granite/Gtk-dark.css");
+                #endif
             }
 
             Gtk.StyleContext.add_provider_for_display (display, dark_provider, Gtk.STYLE_PROVIDER_PRIORITY_THEME);
@@ -81,6 +84,9 @@ namespace Granite {
             if (base_provider == null) {
                 base_provider = new Gtk.CssProvider ();
                 base_provider.load_from_resource ("/io/elementary/granite/Granite.css");
+                #if INCLUDE_GTK
+                    base_provider.load_from_resource ("/io/elementary/granite/Gtk.css");
+                #endif
             }
 
             Gtk.StyleContext.add_provider_for_display (display, base_provider, Gtk.STYLE_PROVIDER_PRIORITY_THEME);
--- a/lib/Init.vala
+++ b/lib/Init.vala
@@ -99,7 +99,7 @@ namespace Granite {
             }
 
 #if INCLUDE_GTK_STYLESHEETS
-            if (gtk_dark_provider != null) { 
+            if (gtk_dark_provider != null) {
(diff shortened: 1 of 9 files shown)


meson_options.txt
```diff
@@ -1,3 +1,4 @@
 option('documentation', type: 'boolean', value: false, description: 'generate documentation with gtk-doc and valadoc')
 option('introspection', type: 'boolean', value: true, description: 'Whether to build introspection files')
 option('demo', type: 'boolean', value: true, description: 'Whether to build granite-demo')
+option('gtk', type: 'boolean', value: false, description: 'Whether to include stylesheets for Gtk widgets when initializing stylesheets')
```
danirabbit, line 4: I think I'd like this to be a little more verbose like `gtk-stylesheet` maybe?

meson.build
```diff
@@ -43,6 +43,14 @@ add_project_arguments(
     language: ['vala']
 )
 
+if get_option('gtk')
+    add_project_arguments(
+        '-D',
+        'INCLUDE_GTK',
```
danirabbit, line 49: Same here. Just so someone doesn't look at this and say "What why wouldn't we compile Granite with GTK?" :sweat_smile:

lib/Styles/meson.build
```diff
@@ -30,6 +30,34 @@ stylesheet_dark_deps = custom_target(
...
+    input: 'Granite-dark.scss',
+    output: 'granite-dark.css',
+    command: [
+        sassc,
+        sassc_opts,
+        '@INPUT@',
+        '@OUTPUT@',
+    ],
+    install: true,
+    install_dir: get_option('datadir') / 'themes' / 'Granite' / 'gtk-4.0'
```
danirabbit, line 58: probably doesn't make sense to install these since they're always loaded by `Granite.init ()`

lib/Init.vala
```diff
@@ -81,6 +84,9 @@ namespace Granite {
             if (base_provider == null) {
                 base_provider = new Gtk.CssProvider ();
                 base_provider.load_from_resource ("/io/elementary/granite/Granite.css");
+                #if INCLUDE_GTK
+                    base_provider.load_from_resource ("/io/elementary/granite/Gtk.css");
```
danirabbit, line 88: According to docs the `load` functions clear any previously loaded resources, so you'd need to add a third and fourth provider for GTK stylesheets

danirabbit (changes requested):
Thanks for taking care of this! This makes sense to me. Just a few comments

lib/Init.vala
```diff
@@ -70,6 +70,9 @@ namespace Granite {
             if (dark_provider == null) {
                 dark_provider = new Gtk.CssProvider ();
                 dark_provider.load_from_resource ("/io/elementary/granite/Granite-dark.css");
+                #if INCLUDE_GTK
+                    dark_provider.load_from_resource ("/io/elementary/granite/Gtk-dark.css");
+                #endif
```
ryonakano, line 75:
```suggestion
#if INCLUDE_GTK
                dark_provider.load_from_resource ("/io/elementary/granite/Gtk-dark.css");
#endif
```

Preprocessor `#if`s are different from normal `if` statement, so you shouldn't make indentation.

lib/Init.vala
```diff
@@ -81,6 +84,9 @@ namespace Granite {
             if (base_provider == null) {
                 base_provider = new Gtk.CssProvider ();
                 base_provider.load_from_resource ("/io/elementary/granite/Granite.css");
+                #if INCLUDE_GTK
+                    base_provider.load_from_resource ("/io/elementary/granite/Gtk.css");
+                #endif
```
ryonakano, line 89:
```suggestion
#if INCLUDE_GTK
                base_provider.load_from_resource ("/io/elementary/granite/Gtk.css");
#endif
```

Same here

danirabbit (commented):
One more thing I would do is, in `register_display` where we have that `gtk_settings` object, we should also set the theme to `Granite`. Then you wouldn't have to mess around in inspector. We'd always load Granite styles and not system styles

ryonakano (commented):
Leaving coding style comments.

danirabbit (approved):
Let's go :rocket:

PR #761, opened 2025-03-20
Demo/UtilsView: fix deprecation warnings

danirabbit:
Use Gtk.ColorDialog, add provider for display

### DIFF
--- a/demo/Views/UtilsView.vala
+++ b/demo/Views/UtilsView.vala
@@ -13,8 +13,6 @@ public class UtilsView : Gtk.Grid {
     """;
     private const string DARK_BG = "#273445";
 
-    private Gtk.StyleContext demo_label_style_context;
-
     construct {
         var tooltip_markup_label = new Gtk.Label ("Markup Accel Tooltips:");
         tooltip_markup_label.halign = Gtk.Align.END;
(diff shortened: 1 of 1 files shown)


jeremypw (changes requested):
The color dialog appears at (0, 0) of the screen rather then centered.  Other than that looks good.

danirabbit:
Screenshot from 2025-03-27 10 45 17

@jeremypw can't confirm that. It appears centered on the window here. Are you still running OS 7? Maybe an OS 7 bug?

jeremypw:
@danirabbit It happens in Classic session but not in Secure session for me.  However, I am running OS8 in a VM on OS7.1

danirabbit:
@jeremypw I can confirm the window is spawning at 0,0 in X11, but we're dimming the correct window so the parent is being set correctly here. There's nothing I can do to set the window position other than setting the parent. So this must be a bug somewhere else

jeremypw (approved):
Sorry to have neglected this. I agree this is an upstream bug in the way the Gtk4 ColorDialog (hopefully not child windows more generally) are positioned under X11.  Hopefully we can make the Wayland session default soon and then it will be less of an issue. I'll approve and then you can decide whether to merge now or later.

PR #763, opened 2025-03-20
Styles: Move expander code from stylesheet

wpkelso:
A small, simple, easy win (for reading Inspector in particular).

image

### DIFF
--- a/lib/Styles/Gtk/Expander.scss
+++ b/lib/Styles/Gtk/Expander.scss
@@ -1,0 +1,18 @@
+expander {
+    min-height: 16px;
+    min-width: 16px;
+    transition: all duration("in-place") easing();
+    -gtk-icon-source: -gtk-icontheme("pan-end-symbolic");
+
+    &:dir(rtl) {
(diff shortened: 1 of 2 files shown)


lib/Styles/Gtk/Expander.scss
```diff
@@ -0,0 +1,18 @@
+expander {
+    min-height: 16px;
```
danirabbit, line 2: What do you think about using rempx for this? I did a poll a while back and it sounds like folks who use text scaling options would prefer symbolic icons be scaled even if it means they could be blurry

wpkelso: Just tested it out, and I do prefer it over the static size. It looks a bit silly to have the tiny icon next to big text, and I don't think the level of blurriness is too bad in this particular instance. For reference:

image
image

danirabbit (approved):

lib/Styles/Gtk/Expander.scss
wpkelso: I guess if blurriness were a big issue down the road, it wouldn't be too bad to add another version hinted at 24x

PR #764, opened 2025-03-21
Demo: use ToolbarView

danirabbit:
Wanna start a conversation about Adwaita widgets and their styles. I think we've conceded that we do use Adwaita for some things like NavigationPage and TabView. We could probably create our own ToolbarView and maintain that and it wouldn't be a huge burden. Should we? Or should we just go ahead and use Adw.ToolbarView?

Screenshot from 2025-03-20 17 29 51

### DIFF
--- a/demo/GraniteDemo.vala
+++ b/demo/GraniteDemo.vala
@@ -61,13 +61,13 @@ public class Granite.Demo : Gtk.Application {
         var end_header = new Gtk.HeaderBar () {
             show_title_buttons = false
         };
-        end_header.add_css_class (Granite.STYLE_CLASS_FLAT);
         end_header.pack_end (new Gtk.WindowControls (END));
         end_header.pack_end (mode_switch);
 
-        var end_box = new Gtk.Box (VERTICAL, 0);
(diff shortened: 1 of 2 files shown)


wpkelso:
If you don't think it'll be too much of a maintenance burden, we probably should vendor our own ToolbarView. I believe it's a common enough idea that a developer would expect to be able to reach for something like it without too much hassle, and giving them that access first party would decrease friction.

teamcons:
Still learning how to do UI in Vala, i discover a lot of cool options by typing a namespace and maybe a letter or two - and then seeing what VS Code spits out.
It has been an amazing shortcut from having to constantly switch out, fire a browser and look at docs.

Having to check out both Granite and Adw (+ Gtk) doubles the effort, and does not guarantee i get the same standardized look eOS apps have - which is what i am aiming for - particularly in the learning phase.
It is "on the shelf and ready for use"

I worry often about the maintenance burden the eOS team has, but this one could also prove useful should it need to diverge from the adwaita one in the future and be expanded on.

PR #765, opened 2025-03-21
Styles: make cards rounded, expand demo, new constants

danirabbit:
Screenshot from 2025-03-20 18 02 33

Turns out [everywhere we use `ROUNDED`](https://github.com/search?q=org%3Aelementary%20Granite.STYLE_CLASS_ROUNDED&type=code) is with `CARD` and the only places [we don't use rounded](https://github.com/search?q=org%3Aelementary%20Granite.STYLE_CLASS_CARD&type=code) with card are:
* in Files, Mail, and Photos because there's no easy way to crop overflow of images in GTK3, so this is not relevant and will be automatically fine in GTK4
* In AppCenter and Notifications indicator because we set a custom border radius

Basically there's no cases where we actually want square cards intentionally. So I think we can just make these round by default and deprecate the rounded class

### DIFF
--- a/demo/Views/CSSView.vala
+++ b/demo/Views/CSSView.vala
@@ -12,8 +12,8 @@ public class CSSView : Gtk.Box {
 
     construct {
         var header1 = new Gtk.Label ("\"title-1\" Style Class") {
-            margin_end = 24,
-            margin_start = 24,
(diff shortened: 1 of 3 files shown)


wpkelso (approved):
Makes sense to me

PR #766, opened 2025-03-21
HeaderLabel: add size property

danirabbit:
Makes title style classes internal to Granite.HeaderLabel as a size property. This also means we get the paired subtitle sizes yay

Screenshot from 2025-03-20 19 38 59

Screenshot from 2025-03-20 19 41 06


wpkelso (approved):

teamcons:
aw yea

PR #767, opened 2025-03-21
Add Granite.Box

danirabbit:
Screenshot from 2025-03-21 10 27 25

* Adds a box subclass with a built-in property for child spacing. Enforces box spacing and makes it respond to text scaling. No more custom CSS or putting in the wrong magic numbers when doing layouts
* Adds some test cases for linked children.


lib/Widgets/Box.vala
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+/**
+ * A {@link Gtk.Box} subclass that has built-in properties for child spacing
+ *
+ * @since 7.7.0
+ */
+[Version (since = "7.7.0")]
+public class Granite.Box : Gtk.Box {
+    public enum ChildSpacing {
```
danirabbit, line 13: Should this be just `SPACING`?

lib/Widgets/Box.vala
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private void update_child_spacing () {
+        string[] css_classes = {
+            ChildSpacing.SINGLE.to_string (),
+            ChildSpacing.DOUBLE.to_string (),
+            ChildSpacing.LINKED.to_string (),
+        };
+
+        foreach (unowned var css_class in css_classes) {
```
danirabbit, line 60: Is there a better big brain way to iterate over enum values in Vala?

lib/Widgets/Box.vala
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * How far apart to space children of #this
+     */
+    public ChildSpacing child_spacing { get; construct set; }
+
+    /**
+     * Constructs a new {@link Granite.Box}
+     */
+    public Box (Gtk.Orientation orientation, ChildSpacing child_spacing = NONE) {
```
danirabbit, line 41: Gtk.Box defaults to 0, but is NONE a good default? Should we default to SINGLE?

lib/Widgets/Box.vala
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @since 7.7.0
+ */
+[Version (since = "7.7.0")]
+public class Granite.Box : Gtk.Box {
+    public enum ChildSpacing {
+        NONE,
+        SINGLE,
+        DOUBLE,
+        LINKED;
```
danirabbit, line 17: Probably needs docs

lib/Widgets/Box.vala
wpkelso: Off-hand I would suspect SINGLE would be the most commonly used value. That is primarily based on my gut though.

lib/Widgets/Box.vala
leolost2605: You can get the enumclass via 
```vala
unowned var enum_class = (EnumClass) typeof (ChildSpacing).class_peek ();
```
with that you can iterate over the enum values like this:
```vala
foreach (unowned var val in enum_class.values) {
    var spacing = (ChildSpacing) val.value; //cast is important otherwise to_string () wont work
    ...
}
```

leolost2605 (changes requested):
A few comments but I like it makes it easier to do the right thing. (e.g. I recently was looking for exactly the linked style but didn't know it existed as a css class)

lib/Widgets/Box.vala
```diff
@@ -0,0 +1,68 @@
...
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+/**
+ * A {@link Gtk.Box} subclass that has built-in properties for child spacing
+ *
+ * @since 7.7.0
+ */
+[Version (since = "7.7.0")]
+public class Granite.Box : Gtk.Box {
```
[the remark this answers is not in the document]
danirabbit: That's a good call. I was wondering if I should just go ahead and make it a box layout and not a Gtk.Box subclass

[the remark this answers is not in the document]
danirabbit: @leolost2605 I opened https://github.com/elementary/granite/pull/770 but I'm kind of less sure about it. It seems like it won't be terribly hard to basically support all the same Gtk.Box API since it mostly inherits from other interfaces, but I wonder if we should wait until there's something more definitive about not being able to subclass Gtk.Box just in case?

wpkelso (approved):

lib/Widgets/Box.vala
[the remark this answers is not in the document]
leolost2605: #770 looks cool but we can wait. We'll have API breaks anyways once GTK5 drops so it doesn't really matter ig.

leolost2605 (approved):
Can't comment on CSS but otherwise looks good

PR #769, opened 2025-03-22
Styles: Propose mixin directory structure

wpkelso:
Proposes an expansion to the naming conventions and the repo structure regarding scss mixins:

- Mixins should be grouped into categories and placed in a file named for the category under the `Mixins` directory
- Treats the line between mixins and variables as quite blurry

**For consideration:**

- Currently files containing only variables are included under mixins, but probably shouldn't? Should they be given their own `Variables` directory, or would that make so many files it's unwieldy?

### DIFF
--- a/lib/Styles/CONTRIBUTING.md
+++ b/lib/Styles/CONTRIBUTING.md
@@ -40,6 +40,37 @@ CSS style classes, such as `Granite.CssClass.CARD`, should be placed in a
 `_classes.scss` file under the relevant library directory (e.g. Granite, Gtk,
 Adw).
 
+Mixins should be grouped into relevant categories, and placed into a file with
+said category name (e.g. Palette, Typography, Animation)
+
+## Repository Structure
+
+```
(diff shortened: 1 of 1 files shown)


danirabbit:
This makes sense to me! Maybe call it `Common` instead of `Mixins` and put variables in there as well? or just put variables in `Mixins` and say "eh close enough naming". I'm not sure there's enough exported variables to be their own whole directory? I guess we could always do a follow up and move them if there are

Do you want to propose moving those files in a separate branch or do it here as well?

wpkelso:
I like `Common` better, I'll change that.

Let's move the files on this branch. Makes it a nicely packaged commit when merged.

wpkelso:
@danirabbit I moved every file that wasn't directly widget class related out into common. I figure everything else will be caught as we progress further. Should be ready for a final review now.

danirabbit (approved):
Let's go :rocket:

PR #770, opened 2025-03-22
Don't subclass Box, use BoxLayout

danirabbit:


jeremypw:
Conflicts to be resolved.

danirabbit:
Closing since it sounds like we can cross this bridge when we come to it

PR #774, opened 2025-03-25
Styles: use button-spacing in popover, headerbar, overlaybar

danirabbit:


wpkelso (commented):
`$button-spacing` uses em, but the styles we already have here use rem. Should button spacing be switched to use rem as well?

lib/Styles/Common/_common.scss
```diff
@@ -2,4 +2,4 @@ $disabled-opacity: 0.5;
 $window_radius: 9px;
 
 // Spacing between children of buttons, menu items etc
-$button-spacing: 0.5em;
+$button-spacing: 0.5rem;
```
wpkelso, line 5: Is this `rem` the same as `rem()`?

wpkelso (changes requested):

lib/Styles/Common/_common.scss
danirabbit: Yeah that function uses rem as well https://github.com/elementary/granite/blob/main/lib/Styles/Common/_utilities.scss

wpkelso: Ah sorry, didn't see that. My bad

wpkelso (approved):

Issue #768, opened 2025-03-21
Use Granite stylesheets for button styles


wpkelso:
From [Gtk docs](https://docs.gtk.org/gtk4/class.Button.html), these are the relevant nodes and classes:

- [x] `button`

- [ ] `.flat`
- [x] `.image-button`
- [x] `.text-button`
- [ ] `.keyboard-activating`

- [ ] `.toggle`
- [ ] `.popup`
- [ ] `.scale`
- [ ] `.lock`
- [ ] `.color`

danirabbit:
We also support:

- [ ] `.back-button`
- [ ] `.circular`
- [ ] `.destructive`
- [ ] `.suggested`

ztefn:
`.pill`?
https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.3/style-classes.html#pill

PR #775, opened 2025-03-25
Add basic mixins for styling controls

wpkelso:
image

image

This is a little haphazard, but I think it gets a proof-of-concept across right now. I figure we can work from here in bits and pieces.

### DIFF
--- a/lib/Styles/Common/_borders.scss
+++ b/lib/Styles/Common/_borders.scss
@@ -1,0 +1,13 @@
+@mixin border-interactive {
+    box-shadow:
+        outset-highlight("full"),
+        outset-shadow(1);
+
+    border-style: solid;
+    border-color: $border-color;
+    border-width: 1px;
+}
+
+@mixin border-interactive-roundrect {
+    border-radius: 3px;
+}
--- a/lib/Styles/Common/_borders.scss
+++ b/lib/Styles/Common/_borders.scss
@@ -9,5 +9,5 @@
 }
 
 @mixin border-interactive-roundrect {
-    border-radius: 3px;
+    border-radius: rem($window_radius / 3);
 }
--- a/lib/Styles/Common/_borders.scss
+++ b/lib/Styles/Common/_borders.scss
@@ -1,6 +1,6 @@
 @mixin border-interactive {
     box-shadow:
-        outset-highlight("full"),
+        highlight("full"),
         outset-shadow(1);
 
     border-style: solid;
--- a/lib/Styles/Common/_controls.scss
+++ b/lib/Styles/Common/_controls.scss
@@ -1,5 +1,5 @@
 @mixin control-enabled {
-    @extend outset
+    @extend %outset-background
     background-color: bg-color(0);
     background-clip: border-box;
 
--- a/lib/Styles/Common/_controls.scss
+++ b/lib/Styles/Common/_controls.scss
@@ -1,4 +1,5 @@
 @mixin control-enabled {
+    @extend outset
     background-color: bg-color(0);
     background-clip: border-box;
 
--- a/lib/Styles/Common/_controls.scss
+++ b/lib/Styles/Common/_controls.scss
@@ -2,11 +2,18 @@
     background-color: bg-color(0);
     background-clip: border-box;
 
+    box-shadow:
+        outset-highlight(),
+        outset-shadow(1);
+
     border-style: solid;
     border-color: $border-color;
     border-width: 1px;
+
+    transition: all duration("in-place") easing("default");
 }
 
 @mixin control-disabled {
 
 }
+
--- a/lib/Styles/Common/_controls.scss
+++ b/lib/Styles/Common/_controls.scss
@@ -1,5 +1,5 @@
 @mixin control-enabled {
-    @extend %outset-background
+    @extend %outset-background;
     background-color: bg-color(0);
     background-clip: border-box;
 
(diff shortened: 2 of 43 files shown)


lib/Styles/Common/_controls.scss
```diff
@@ -0,0 +1,10 @@
+@mixin control {
+    @extend %outset-background;
+    background-color: bg-color(0);
+    background-clip: border-box;
+
+    @include border-interactive;
+
+    transition: all duration("in-place") easing("default");
```
danirabbit, line 8: Let's use easing function from animate?
```suggestion
    transition: all duration("in-place") $easing;
```

lib/Styles/Common/_borders.scss
```diff
@@ -0,0 +1,13 @@
...
+        outset-highlight("full"),
+        outset-shadow(1);
+
+    border-style: solid;
+    border-color: $border-color;
+    border-width: 1px;
+}
+
+@mixin border-interactive-roundrect {
+    border-radius: 3px;
```
danirabbit, line 12: Let's scale this on window radius for our nested radii theory
```suggestion
    border-radius: rem($window_radius / 3);
```

lib/Styles/Index-dark.scss
```diff
@@ -43,9 +43,3 @@ $warning-icon-color: $BANANA_500;
 // Common styles
 @import '_label.scss';
 @import '_osd.scss';
-
-// Gtk Widgets
-// @import 'Gtk/Index.scss';
-
-// Granite widgets
-@import 'Granite/Index.scss';
```
danirabbit, line 51: This seems unrelated to button stuff. Can you propose this in a separate PR?

meson.options
```diff
@@ -1,4 +1,4 @@
 option('documentation', type: 'boolean', value: false, description: 'generate documentation with gtk-doc and valadoc')
 option('introspection', type: 'boolean', value: true, description: 'Whether to build introspection files')
 option('demo', type: 'boolean', value: true, description: 'Whether to build granite-demo')
-option('gtk-stylesheets', type: 'boolean', value: false, description: 'Whether to include stylesheets for Gtk widgets when initializing stylesheets')
+option('gtk-stylesheets', type: 'boolean', value: true, description: 'Whether to include stylesheets for Gtk widgets when initializing stylesheets')
```
danirabbit, line 4: Oops, still wanna keep this false by default for now

lib/Styles/Gtk/Button.scss
```diff
@@ -3,6 +3,9 @@ button {
     // the necessary SVGs for light/dark and accent color combinations is
     // untenable. Ideally we'd deprecate this in favor of something like a
     // Granite.BackButton with custom drawing; until then, stick an icon in it.
+    @include control;
+    @include border-interactive-roundrect;
```
danirabbit, line 7: Can we scope this down to `&.textbutton`? That's something I wanted to not repeat here is that image buttons never have this kind of raised style and we have to undo all the default styles for image buttons :sweat_smile: 

Also this looks like it got inserted under the comment for the `.back-button` class :grimacing:

danirabbit (changes requested):

meson.options
wpkelso: Oops yup, I really should not put stuff up for review when I checked it over while very tired 😅

lib/Styles/Common/_controls.scss
wpkelso: It does use the function from `_animate`. `$easing` isn't defined outside the easing function

danirabbit: Sorry I don't know how I misread this lmao

danirabbit (changes requested):
Ah I figured out why the highlight colors are working. All the instances of

```
#{'alpha(@highlight_color, 0.3)'}
```

need to be changed to something like

```
scale-color($highlight_color, $alpha: -70%)
```

since we don't have a `@highlight_color` set up

lib/Styles/Common/_lighting.scss
```diff
@@ -61,3 +64,69 @@
...
+        @return
+            0 1px 0 0 #{'alpha(@highlight_color, 0.3)'},
+            inset 0 1px 1px rgba(black, 0.05);
+    }
+
+    @return
+        0 1px 0 0 #{'alpha(@highlight_color, 0.3)'},
+        inset 0 1px 1px rgba(black, 0.05),
+        inset 0 0 1px 1px rgba(black, 0.05);
+}
```
danirabbit, line 79: It doesn't look like this is used anywhere

lib/Styles/Common/_lighting.scss
```diff
@@ -61,3 +64,69 @@
...
+
+@function outset-highlight($sides: "full") {
```
danirabbit, line 81: This is actually a duplicate there's a `highlight` function right above this

danirabbit (changes requested):
Couple more things I noticed

lib/Styles/Common/_borders.scss
```diff
@@ -0,0 +1,13 @@
+@mixin border-interactive {
+    box-shadow:
+        outset-highlight("full"),
```
danirabbit, line 3:
```suggestion
        highlight("full"),
```

danirabbit (changes requested):

lib/Styles/Common/_borders.scss
wpkelso: Totally forgot to check there weren't any errors when actually running it 🙃

danirabbit (approved):
Let's go :rocket:

wpkelso:
related #768

PR #777, opened 2025-03-26
Demo: create DemoPage

danirabbit:
Sets things up for a future where each page might have its own title like Switchboard pages and/or be a ToolbarView subclass. Views no longer add their own scrolledwindow


jeremypw (approved):
Makes sense and no visual differences found

Issue #375, opened 2020-03-13
Deletable listboxrow?


cassidyjames:
A pattern we're using in three places now is a listbox with deletable items, including the little spin-out delete button. It would need to be the listbox as well as the row. The CSS is also currently duplicated, and could either be pulled into Granite (for this widget) or just into the stylesheet.

danirabbit:
I think we want something more generic than just deletable. Looking at Mail for example I think we want like an `ActionableRow`

probably want it to be a Gtk.ListBoxRow subclass so it’s transparent to use with ListBox normally

I should be able to add an action with something like:
`add_action (string label, string detailed_action_name, GLib.Icon? icon = null,  Color? color = null, Gtk.Align? alignment = null)`

All actions should be added to a popover that activates with both right click and long press. Actions with an icon can be added to the popover as little circle buttons maybe? Actions with an align should be added as swipe actions.

There should be maybe some elegant way to handle stateful actions like read/unread for example

danirabbit:
On iOS for example you can have like a shitload of actions per row. A short swipe reveals all the actions and the row snaps in place to reveal them all. A long swipe activates the outermost action

![IMG_6621](https://github.com/user-attachments/assets/beb9fd96-e2de-433c-b5a1-34860daea6ab)

PR #783, opened 2025-03-27
Add ListBox styles, Granite.ListItem

danirabbit:
Screenshot from 2025-03-31 16 02 53

* Create a new ListsView page with box a card/boxed Gtk.ListBox and a scrolled Gtk.ListView.
* `boxed-list` is something from Adw and Gtk.Inspector. We never made a constant for this. I'm proposing we still don't and instead use `.card` here. That seems simpler to me, we already have a style class for cards, this just means put my list on a card :shrug:
* Deprecate `Granite.STYLE_CLASS_RICH_LIST` and replace it with `Granite.ListItem`. The idea being to create predictable widgets we can style the same every time without having to rely on developers creating list item widgets in a magic order to style them correctly. We can probably add properties here to cover all the main major uses cases like icons, buttons, actions, context menus, all kinds of stuff. Even eventually #375 
* I avoided adding a constant for `.frame` for now, but I also avoided deprecating the old `Granite.STYLE_CLASS_FRAME`. I'd like to follow up with this later, but I think we can completely deprecate that constant for `has-frame` properties in GTK widgets (and in our widgets if we want that) in the future.
* I also avoided abstracting `.frame` for now. This class probably means different things for different widgets, so for now I'm considering it an internal detail of the widget.
* Similarly I avoided abstracting `.view` or creating a new constant for it. There has been demand in the past for `view`, `background`, and `sidebar` classes, but I want to revisit that in another branch and just focus on the lists here


danirabbit:
I've actually decided that I hate this and what I really want is to completely deprecate `RICH_LIST` and instead create a Granite.ListItem widget. It would have the same amount of coverage of features but actually be predictable to style. No magic and expecting developers to construct the widget in a very specific way. Just a widget with properties like:
- [ ] string label
- [ ] string description
- [ ] GLib.Icon icon
- [ ] IconSize icon_size or maybe pixel_size?

Functions:
- [ ] `add_button` to add buttons to a button box at the end
- [ ] Maybe a `content` property to add custom widgets like a dropdown or picture 

We could even do things similar to Gtk.PopoverMenu like taking an action and:
- if it's a normal action make the row activatable with like a > indicator
- if it's a stateful action with 2 states give it a switch
- if it's a stateful action with 3 or more states give it a radio button

If you give it a menu model, do all the work of setting up a PopoverMenu automatically with long press and right click

If you give it a cancellable, add a cancel image button

lib/Widgets/ListItem.vala
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public string text { get; set; }
+
+    /**
+     * Small, dim description text shown when active.
+     */
+    public string? description { get; set; }
+
+    class construct {
+        set_css_name ("granite-listitem");
```
danirabbit, line 24: I don't know if there's a better way to name this. I wanted to avoid colliding with Gtk.ListItem which will be the parent of this widget in a ListView

wpkelso (approved):
Looks good to me. I definitely am +1 on less magic and more predictability

danirabbit:
I'm gonna go ahead and merge. Got a lot of acks in Discord as well even if they didn't turn into a dev team review :p

PR #785, opened 2025-03-28
Styles/Controls: merge in background styles

danirabbit:
Since this is just the background style for controls, merge it into the controls mixin.

Headerbars are pretty much flat by default these days since everything is moving to ToolbarView and split views and the old headerbar style is out of fashion


wpkelso:
> Headerbars are pretty much flat by default these days since everything is moving to ToolbarView and split views and the old headerbar style is out of fashion

How would this change affect headerbars on non-gtk windows? I.e. some electron app that isn't doing it's own window control shenanigans

wpkelso (approved):

PR #787, opened 2025-03-28
Styles/Controls: prototype focus styles

danirabbit:
Folks have often complained about focus styles or interpret them as being a bug. I've had this idea for a while about fading out focus styles. Would appreciate feedback :)

[Kooha-2025-03-28-11-48-33.webm](https://github.com/user-attachments/assets/7e6837e0-3d60-4fd5-9a1f-3d81c9922ddc)

### DIFF
--- a/lib/Styles/Common/_controls.scss
+++ b/lib/Styles/Common/_controls.scss
@@ -19,5 +19,29 @@
             scale-color($highlight_color, $alpha: -70%)
         );
     }
+
+    // FIXME: replace with accent color when we merge https://github.com/elementary/granite/pull/704
+    &:focus {
+        @if $color-scheme == "light" {
+            border-color: #{'@BLUEBERRY_500'};
+            outline-color: #{'alpha(@BLUEBERRY_500, 0.3)'};
+        } @else {
+            border-color: #{'@BLUEBERRY_900'};
+            outline-color: #{'alpha(@BLUEBERRY_300, 0.3)'};
+        }
+
+        outline-width: rem(2px);
+        outline-style: solid;
+
+        animation: unfocus 600ms easing() 1s 1;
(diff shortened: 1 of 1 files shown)


lib/Styles/Common/_controls.scss
```diff
@@ -19,5 +19,29 @@
...
+    &:focus {
+        @if $color-scheme == "light" {
+            border-color: #{'@BLUEBERRY_500'};
+            outline-color: #{'alpha(@BLUEBERRY_500, 0.3)'};
+        } @else {
+            border-color: #{'@BLUEBERRY_900'};
+            outline-color: #{'alpha(@BLUEBERRY_300, 0.3)'};
+        }
+
+        outline-width: rem(2px);
```
danirabbit, line 33: Unfortunately we still have https://github.com/elementary/stylesheet/issues/904 with linked buttons. I wonder if maybe we should forget the outline property and use box-shadow?

danirabbit:
It looks like Adw has some more complicated focus ring handling that it's probably worth looking at. They also do the thing of dismissing the focus ring after some time:

https://gitlab.gnome.org/search?search=focus-ring&nav_source=navbar&project_id=14079&group_id=8&search_code=true&repository_ref=main

https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/stylesheet/_drawing.scss#L12

https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/stylesheet/_common.scss#L3

lib/Styles/Common/_controls.scss
wpkelso: I think it's worth exploring. Using box-shadow might also remove some of the funky effects of the shadow and the focus indicator overlapping.

PR #815, opened 2025-04-01
Add basic Shumate support

### LINKED ISSUE #802: Provide native support for libshumate
### Problem

[libshumate](https://gitlab.gnome.org/GNOME/libshumate) is the GTK4 library providing maps. It contains a simple map widget that is not well styled with the elementary stylesheet.
One can use the demo app provided by the library itself.

### Proposal

Implement the styling directly

### Prior Art (Optional)

_No response_

danirabbit:
Fixes #802 

* Does the least to provide basic styles for Shumate SimpleMap
* Add a style for button.osd
* Make sure we don't add spacing to .osd inside .linked
* Only adds a Shumate dep to the Demo. The library does not depend on Shumate

Screenshot from 2025-03-31 20 01 44

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ You'll need the following dependencies:
 * libgee-0.8-dev
 * libgirepository1.0-dev
 * libgtk-4-dev >= 4.4.0
+* libshumate-dev
 * sassc
 * valac
 
--- a/README.md
+++ b/README.md
@@ -16,9 +16,11 @@ You'll need the following dependencies:
 * libgee-0.8-dev
 * libgirepository1.0-dev
 * libgtk-4-dev >= 4.4.0
-* libshumate-dev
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -16,6 +16,7 @@ You'll need the following dependencies:
 * libgee-0.8-dev
 * libgirepository1.0-dev
 * libgtk-4-dev >= 4.4.0
+* libshumate-dev
```
ryonakano, line 19: Tiny thing though, maybe we may want to explicit that this is only necessary when building the demo? Because we can build Granite without the demo these days using the `-Ddemo` option.

wpkelso (approved):
It looks like there is some stuff that could be abstracted, but its probably better to do that in another more general abstraction pass

PR #818, opened 2025-04-01
Styles: Detail pass for buttons

wpkelso:
Little bit of detail work for buttons:

- add color or depressed backgrounds for text buttons and checkbuttons
- subtly darken when active, and lighten when hovered
- reduce opacity and greyscale when disabled
- removed herobrine

image
image


danirabbit (approved):
:rocket:

PR #819, opened 2025-04-02
control: Add libshumate-dev as dependency of demo

ryonakano:
Follow up to #815


danirabbit (approved):
Ah thank you! I forgot to follow up with that

PR #823, opened 2025-04-03
Control: move shumate to build deps

### LINKED ISSUE #822: Gettext workflow is missing dependencies
Image

Looks like both `libshumate` and `cmake` interestingly enough

danirabbit:
Shumate was added to demo runtime deps, but we actually need to make it a build time dep

Fixes #822

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -7,6 +7,7 @@ Build-Depends: debhelper (>= 10.5.1),
                libgee-0.8-dev,
                libgirepository1.0-dev,
                libgtk-4-dev (>= 4.12.0),
+               libshumate-dev,
                meson (>= 0.57.0),
                sassc,
                valac (>= 0.40)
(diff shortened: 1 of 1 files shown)


ryonakano:
Doesn't this because the demo option is enabled by default and gettext actions tries to build the demo……?

ryonakano:
IIRC we don't support translations in the demo so we can disable building the demo in the gettext actions.

danirabbit:
@ryonakano I'm not sure if we can do that in the gettext action currently. I think it just builds whatever the package builds. So we'd have to disable demo building in the package no?

ryonakano:
> So we'd have to disable demo building in the package no?

Sorry, I realized your PR is correct because it is the Depends section of the demo, not the Build-Depends section, where I added libshumate-dev in the previous PR 😵

So I'm approving your PR. Thank you for your fix.

ryonakano (approved):

PR #829, opened 2025-04-08
Styles/Scale: style values

### LINKED ISSUE #800: Style scale value
### Problem

We don't current ever show scale values, but in same cases it might be nice to show the value while dragging the slider

### Proposal

Style scale value as an .osd element:

Screenshot from 2023-06-24 16 34 58

```
scale value {
background-color: alpha(@BLACK_700, 0.9);
border-radius: 3px;
box-shadow: 0 1px 3px rgba(0, 0, 0, 0.12), 0 1px 2px rgba(0, 0, 0, 0.24);
color: white;
padding: 3px 6px;
margin: -6px -6px -3px -6px;
transition: all 200ms ease-in-out;
}

scale:not(.dragging) value {
color: transparent;
background: transparent;
box-shadow: none;
}
```

### Prior Art (Optional)

https://github.com/elementary/switchboard-plug-keyboard/pull/447

danirabbit:
Fixes #800 

Screenshot from 2025-04-08 13 23 16

### DIFF
--- a/demo/Views/ControlsView.vala
+++ b/demo/Views/ControlsView.vala
@@ -117,7 +117,8 @@ public class ControlsView : DemoPage {
         var scale_header = new Granite.HeaderLabel ("Scale");
 
         var hscale = new Gtk.Scale.with_range (HORIZONTAL, 0, 100, 1) {
-            hexpand = true
+            hexpand = true,
+            draw_value = true
         };
         hscale.adjustment.value = 50;
 
--- a/lib/Styles/Gtk/Scale.scss
+++ b/lib/Styles/Gtk/Scale.scss
@@ -25,9 +25,40 @@ scale {
         min-width: $trough-width;
     }
 
+    value {
+        @extend .small-label;
+    }
+
     &.horizontal {
         slider {
             margin: $slider-margin -1px;
+        }
+
+        value {
+            @extend .osd;
+
+            // Padding + text height + margin
+            $value-osd-height: (rem(6px) * 2) + 0.85rem + rem(6px);
+
+            &.bottom,
+            &.top {
+                opacity: 0;
+                transition: all duration("in-place") easing();
+            }
+
+            &.bottom {
+                margin-bottom: $value-osd-height * -1;
+                margin-top: rem(3px);
+            }
+
+            &.top {
+                margin-bottom: rem(3px);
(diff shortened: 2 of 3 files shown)


danirabbit:
@wpkelso One issue here is only TOP and BOTTOM positions follow the slider, so I don't want to style LEFT and RIGHT as `osd`, but when I add the extends inside `&.bottom, &.top` the margins break. Any ideas?

wpkelso:
> One issue here is only TOP and BOTTOM positions follow the slider, so I don’t want to style LEFT and RIGHT as `osd`, but when I add the extends inside `&.bottom, &.top` the margins break. Any ideas?

I don't see much of a visual difference, but there seems to be an error thrown with it in its current location, while that error goes away when it is moved to be inside `&.bottom, &.top`. E.g.

Screenshot from 2025-04-09 14 39 05
versus
Screenshot from 2025-04-09 14 38 03

danirabbit:
@wpkelso look at the vertical alignment of the horizontal scale in your second screenshot. It's way off center compared to the top screenshot

wpkelso:
@danirabbit It looks like when the extends is placed inside the `&.bottom, &.top` the overrides for `top-margin` and `bottom-margin` aren't cascading correctly. Within `Gtk.css` they're calculated out to the exact same rem values in the exact same place. There isn't anything further down the file that touches those values. 

I haven't been able to figure out why they then get calculated to different pixel values, so maybe it's some weird edge case in rem conversion? The following are the margins and padding for the `value` elements in either case. I used 1.25 text scaling on the top image, but that shouldn't create the specific behaviours here.

image

image

wpkelso:
Thinking about it, rem calculation error is most likely the wrong conclusion. It looks like the overrides just aren't getting applied for some reason.

danirabbit:
@wpkelso I think it was fixed in #840

wpkelso:
> I think it was fixed in #840

Ah yup, looks like it. Do we know specifically which change fixed it? Was it the thing where `osd`s couldn't be modified?

danirabbit:
@wpkelso yeah exactly. Setting the provider priority so that it wasn't always overriding the gtk styles

wpkelso (approved):
Looks all tidy to me 🚀

PR #830, opened 2025-04-08
Styles/Scale: add dragging style for slider

danirabbit:
Please excuse the little GTK rendering bug in the shadow :sweat_smile: 

https://github.com/user-attachments/assets/e4995943-7101-40f9-b796-0f032341b801

### DIFF
--- a/demo/DemoPage.vala
+++ b/demo/DemoPage.vala
@@ -4,6 +4,8 @@
  */
 
 public class DemoPage : Gtk.Box {
+    public string title { get; set; }
+
     public Gtk.Widget content {
         set {
             scrolled_window.child = value;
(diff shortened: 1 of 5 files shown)


lib/Styles/Gtk/Scale.scss
```diff
@@ -25,6 +29,19 @@ scale {
         min-width: $trough-width;
     }
 
+    &.dragging {
+        slider {
+            transform: scale(1.15);
```
wpkelso, line 34: What're your thoughts on making this slightly subtler? Playing around, I found I liked it in the `1.05` to `1.10` range

wpkelso (changes requested):
Very satisfying, just a small inquiry from playing around with the scaling values on my own

lib/Styles/Gtk/Scale.scss
danirabbit: I personally thought 1.05 was too weak, so happy to compromise at 1.1 :)

wpkelso (approved):
🚀

PR #831, opened 2025-04-08
Constants: deprecate scale styles

danirabbit:
These constants were never used outside of Granite.

The `warmth` style was added for Nightlight, so it's only used in settings and the indicator. We should just add the styles there. I don't think they're generally useful for apps: https://github.com/search?q=org%3Aelementary+class+%28%22warmth%22%29%3B+language%3AVala&type=code

Same story with the `temperature` style. It was only used in Photos: https://github.com/search?q=org%3Aelementary+add_class+%28%22temperature%22%29+language%3AVala&type=code


danirabbit:
Yeah it's only a couple lines to add your own background gradient to a scale trough, so I think it's not a big deal

PR #834, opened 2025-04-09
Header: modernize h2 styles

danirabbit:
Really thin header styles are out of fashion. We are currently carrying custom CSS for H2 headers in Icon Browser and Settings for example. So update to match those:

Screenshot from 2025-04-08 17 53 24

Add a title to demo page so we can use it in context

### DIFF
--- a/demo/DemoPage.vala
+++ b/demo/DemoPage.vala
@@ -19,7 +19,8 @@ public class DemoPage : Gtk.Box {
             hexpand = true,
             size = H2,
             margin_top = 3,
-            margin_bottom = 3
+            margin_bottom = 3,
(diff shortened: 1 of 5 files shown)


wpkelso (commented):
I like this styling more than the old version, but it seems like we're expecting developers to construct the window in a specific way to achieve this. Is that actually the case? And if it is, should we be thinking about a more ergonomic way to let developers use this?

danirabbit:
@wpkelso It's way more fucky in the demo because we don't have Adw.ToolbarView. Here's an example of what the diff looks like in practice: https://github.com/elementary/iconbrowser/pull/64

PR #835, opened 2025-04-09
Styles: Add padding and minimum sizes for buttons

wpkelso:
Screenshot from 2025-04-09 15 06 54
Screenshot from 2025-04-09 15 07 39

### DIFF
--- a/lib/Styles/Gtk/Button.scss
+++ b/lib/Styles/Gtk/Button.scss
@@ -1,10 +1,7 @@
 button {
 
-    min-height: rem(16px);
-    min-width: rem(16px);
-
     &.text-button {
-        padding: rem(2px) rem(6px);
+        padding: rem(2px) rem(4px);
     }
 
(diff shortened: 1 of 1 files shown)


lib/Styles/Gtk/Button.scss
```diff
@@ -1,9 +1,16 @@
 button {
+
+    min-height: rem(16px);
+    min-width: rem(16px);
```
danirabbit, line 4: This makes window controls super blurry

danirabbit (changes requested):
Ah wow these are really huge :grimacing: I think maybe less padding would be better. Compared to the old stylesheet:

Screenshot from 2025-04-09 12 17 04
Screenshot from 2025-04-09 12 17 34

wpkelso:
image

danirabbit (approved):
Let's go :tada:

PR #840, opened 2025-04-10
Styles: add Gtk.Video

### LINKED ISSUE #804: Add GTK4 video styles

danirabbit:
Fixes #804 

https://github.com/user-attachments/assets/0f782835-3709-4a8c-9af6-01222a96c670

* Add Gtk.Video to Demo
* Adjust provider priority. Fixes a bug where `.osd` styles were way too OP and couldn't be overridden
* Move `.circular` because this isn't just a thing that applies to buttons

Button:
* Make sure we set `color` in `button.osd` so that it doesn't inherit a non-contrasty color
* reset margins in `button.osd` instead of using expensive `:not` selector in `.osd`
* Add the same fun bounce transition we use in window controls and popovers to OSD buttons, but not when they're linked

Scale:
* add a transition for opacity to sliders
* Set rounded radius from `fill` origin, but make sure it's still square on the end

Slider:
* Don't use 50% `border-radius` in case our slider is non-square

Video:
* Add Styles for Gtk.Video and Gtk.MediaControls based on the custom mediacontrols we did for Videos
* Make sure video widgets on a square still get a black letter/pillarbox
* Use tabular numbers so the scale position doesn't jump around
* In this context, Fill must mean we're downloading a video so add a fun pulse animation here. I'm not sure if Gtk.Video actually uses this, I just messed around in inspector. I can remove this if it's excessive :sweat_smile:

### DIFF
--- a/demo/GraniteDemo.vala
+++ b/demo/GraniteDemo.vala
@@ -31,6 +31,7 @@ public class Granite.Demo : Gtk.Application {
         var settings_uris_view = new SettingsUrisView ();
         var style_manager_view = new StyleManagerView ();
         var utils_view = new UtilsView ();
+        var video_view = new VideoView ();
         var placeholder = new WelcomeView ();
         var dialogs_view = new DialogsView (window);
         var application_view = new ApplicationView ();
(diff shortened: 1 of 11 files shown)


wpkelso (approved):
🚀

PR #848, opened 2025-04-21
Styles: Add lighting to progress/scale bars

wpkelso:
Basically just transferring the checked styles, using a little bit of lighting to make the bars stand out more. It may be the wrong path to follow, given there seemed to be some support for keeping the initial styles in the discord.

<details>
<summary>Initial Proposal</summary>

</details>

image
image

### DIFF
--- a/lib/Styles/Common/_lighting.scss
+++ b/lib/Styles/Common/_lighting.scss
@@ -92,13 +92,3 @@
         inset 0 0 1px 1px rgba(black, 0.05);
 }
 
-@mixin lighting-protrude {
-    background-clip: border-box;
-    background-image:
-        linear-gradient(
-            to bottom,
-            scale-color($highlight_color, $alpha: -80%),
-            rgba(white, 0)
-        );
-}
-
--- a/lib/Styles/Common/_lighting.scss
+++ b/lib/Styles/Common/_lighting.scss
@@ -91,3 +91,14 @@
         inset 0 1px 1px rgba(black, 0.05),
         inset 0 0 1px 1px rgba(black, 0.05);
 }
+
+@mixin lighting-protrude {
+    background-clip: border-box;
+    background-image:
+        linear-gradient(
+            to bottom,
+            scale-color($highlight_color, $alpha: -80%),
+            rgba(white, 0)
+        );
+}
+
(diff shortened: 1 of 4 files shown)


lib/Styles/Gtk/ProgressBar.scss
```diff
@@ -32,7 +32,7 @@ progressbar {
                         rgba(white, 0)
                     );
                 background-repeat: no-repeat;
-                background-size: 48px 100%;
+                background-size: 100% 100%;
```
danirabbit, line 35: I don't think you meant to do this. Breaks animations

lib/Styles/Common/_lighting.scss
```diff
@@ -91,3 +91,14 @@
         inset 0 1px 1px rgba(black, 0.05),
         inset 0 0 1px 1px rgba(black, 0.05);
 }
+
+@mixin lighting-protrude {
```
danirabbit, line 95: Probably don't want this because, for example, the progressbar uses multiple background gradients. In the old stylesheet we had:

```
// Background for elements with outset style like headerbars, buttons, checkboxes, etc
%outset-background {
    background-image:
        linear-gradient(
            to bottom,
            #{'alpha(@highlight_color, 0.2)'},
            rgba(white, 0)
        );

    &:backdrop {
        background-image:
            linear-gradient(
                to bottom,
                #{'alpha(@highlight_color, 0.35)'},
                #{'alpha(@highlight_color, 0.3)'}
            );
    }
}
```

danirabbit (changes requested):
I don't like the protruded styles for the scale at all. I think the flat style worked better here for scale fill

I'm not super opposed to the progressbar highlight being protruded, but there's some bugs introduced here with the animation

lib/Styles/Gtk/ProgressBar.scss
wpkelso: Yup, forgot to clean that up

wpkelso:
I've changed it so now we only have a bordered edge where the fill meets the trough&mdash;very subtle but I think it does add a little, especially in light theme

image
image

stsdc:
If the fill gets a volume then the trough needs to have a bit of shadow.
These designs reminded me of this:
obraz

obraz

wpkelso:
> If the fill gets a volume then the trough needs to have a bit of shadow.

You're right, that helps a ton. Thanks for the reminder

image
image

PR #849, opened 2025-04-21
Styles: Use slider handle styles for scrollbar handle

wpkelso:
Not 100% sure about this, but wanted to explore the idea given they're very similar interactions.

[Kooha-2025-04-21-16-05-53.webm](https://github.com/user-attachments/assets/e8efd26b-924f-40d6-ba2d-93111436f10c)


danirabbit:
Hrm yeah I dunno. It's interesting, but I don't think I love it.

PR #850, opened 2025-04-23
Style (pane) separators and sidebars

wpkelso:
image

### DIFF
--- a/lib/Styles/Granite/Sidebar.scss
+++ b/lib/Styles/Granite/Sidebar.scss
@@ -10,6 +10,7 @@
         }
 
         &:selected {
+            // FIXME: replace with translucent/brighter accent color
             background-color: bg-color(4);
         }
     }
--- a/lib/Styles/Granite/Sidebar.scss
+++ b/lib/Styles/Granite/Sidebar.scss
@@ -1,0 +1,16 @@
+.sidebar {
+    background-color: bg-color(3);
+
+    & row {
+        padding: rem(6px) rem(12px);
+        background-color: inherit;
+
+        &:hover {
+            background-color: $highlight_color;
+        }
+
+        &:selected {
+            background-color: bg-color(4);
+        }
+    }
+}
--- a/lib/Styles/Gtk/Separator.scss
+++ b/lib/Styles/Gtk/Separator.scss
@@ -1,0 +1,5 @@
+separator {
+    background-color: $toplevel-border-color;
+    min-width: 1px;
+    border-right: solid 1px $highlight_color;
+}
--- a/lib/Styles/Gtk/Separator.scss
+++ b/lib/Styles/Gtk/Separator.scss
@@ -1,5 +1,11 @@
-separator {
-    background-color: $toplevel-border-color;
-    min-width: 1px;
-    border-right: solid 1px $highlight_color;
+$separator-color: rgba(black, 0.15);
+@if $color-scheme == "dark" {
+    $separator-color: rgba(black, 0.35);
 }
+
+separator {
+    background-color: $separator-color;
+    min-height: 1px;
+    min-width: 1px;
+}
+    
--- a/lib/Styles/Gtk/Sidebar.scss
+++ b/lib/Styles/Gtk/Sidebar.scss
@@ -1,3 +1,8 @@
+.sidebar {
+    background-color: bg-color(3);
+    box-shadow: highlight();
+}
+
 stacksidebar.sidebar {
     background: none;
     box-shadow: none;
--- a/lib/Styles/Gtk/Sidebar.scss
+++ b/lib/Styles/Gtk/Sidebar.scss
@@ -11,6 +11,8 @@ stacksidebar.sidebar {
         padding: rem(6px) rem(12px);
         background-color: inherit;
 
+        transition: all duration("in-place") easing();
+
         &:hover {
             background-color: $highlight_color;
         }
--- a/lib/Styles/Gtk/Sidebar.scss
+++ b/lib/Styles/Gtk/Sidebar.scss
@@ -7,19 +7,20 @@ stacksidebar.sidebar {
     background-color: bg-color(3);
     box-shadow: highlight();
 
-    & row {
-        padding: rem(6px) rem(12px);
-        background-color: inherit;
+}
 
-        transition: all duration("in-place") easing();
+stacksidebar.sidebar row {
+    padding: rem(6px) rem(12px);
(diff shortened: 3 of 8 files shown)


meson_options.txt
```diff
@@ -1,4 +1,4 @@
 option('documentation', type: 'boolean', value: false, description: 'generate documentation with gtk-doc and valadoc')
 option('introspection', type: 'boolean', value: true, description: 'Whether to build introspection files')
 option('demo', type: 'boolean', value: true, description: 'Whether to build granite-demo')
-option('gtk-stylesheets', type: 'boolean', value: false, description: 'Whether to include stylesheets for Gtk widgets when initializing stylesheets')
+option('gtk-stylesheets', type: 'boolean', value: true, description: 'Whether to include stylesheets for Gtk widgets when initializing stylesheets')
```
danirabbit, line 4: Looks like this accidental change got in again :p

wpkelso: 🥲

lib/Styles/Gtk/Separator.scss
```diff
@@ -0,0 +1,5 @@
+separator {
+    background-color: $toplevel-border-color;
+    min-width: 1px;
+    border-right: solid 1px $highlight_color;
```
danirabbit, line 4: Probably should make sure this is wrapped in `&.vertical`

lib/Styles/Gtk/Separator.scss
```diff
@@ -0,0 +1,5 @@
+separator {
```
danirabbit, line 1: In the old stylesheet we had separately defined separator colors that were more subtle. Toplevel border color is pretty harsh:

```scss
$separator-color: rgba(black, 0.15);
$separator-highlight-color: rgba(white, 0.8);

@if $color-scheme == "dark" {
    $separator-color: rgba(black, 0.35);
    $separator-highlight-color: rgba(white, 0.05);
}
```

lib/Styles/Granite/Sidebar.scss
```diff
@@ -0,0 +1,17 @@
+.sidebar {
```
danirabbit, line 1: This is actually not a granite style. It's from Gtk.StackSidebar

lib/Styles/Granite/Sidebar.scss
```diff
@@ -0,0 +1,17 @@
...
+    background-color: bg-color(3);
```
danirabbit, line 2: Using background color blocks the inset box-shadow on windows. You can see it in dark mode. So we need to add `box-shadow: highlight();` here.

But beware that if you have the whole pane with `.sidebar` style class and the it contains Gtk.StackSidebar then you get double box shadow. So specifically `stacksidebar.sidebar` needs no box shadow and probably no background so it doesn't cover up the shadow again. Alternatively you could clear styles for `.sidebar .sidebar`

danirabbit (changes requested):

lib/Styles/Granite/Sidebar.scss
wpkelso: Ah, I couldn't tell whether it was the selector from Gtk or the selector for `CssClass.SIDEBAR`

danirabbit: Yeah we were carrying a STYLE_CLASS_SIDEBAR and kind of abusing it. But maybe we should add `CssClass.SIDEBAR = sidebar-pane` and make sure we're actually not creating this kind of issue with nested styles of the same class

lib/Styles/Gtk/Separator.scss
wpkelso: It doesn't seem like the paned separators have horizontal or vertical style classes attached to them, so I've removed the highlight/shadow effect for now

danirabbit: Yeah if you're specifically wanting to style pane seps and not all seps (in menus and all else) probably should create a `Paned.scss` and do `paned.vertical > separator`

lib/Styles/Gtk/Separator.scss
```diff
@@ -0,0 +1,11 @@
+$separator-color: rgba(black, 0.15);
+@if $color-scheme == "dark" {
+    $separator-color: rgba(black, 0.35);
+}
+
+separator {
```
danirabbit, line 6: If you're specifically wanting to style pane seps and not all seps (in menus and all else) probably should create a Paned.scss and do `paned.vertical > separator`

lib/Styles/Gtk/Sidebar.scss
```diff
@@ -0,0 +1,25 @@
...
+
+        &:hover {
+            background-color: $highlight_color;
+        }
+
+        &:selected {
+            // FIXME: replace with translucent/brighter accent color
+            background-color: bg-color(4);
+        }
+    }
```
danirabbit, line 24: This seems like it should be in `stacksidebar.sidebar` since otherwise this would effect Granite.ListItem styles in a sidebar using those widgets

danirabbit (changes requested):

lib/Styles/Gtk/Sidebar.scss
danirabbit: There's probably a better way to abstract these styles, but what do you think about how this looks?
```
stacksidebar.sidebar {
    background: none;
    box-shadow: none;

    .navigation-sidebar {
        margin: $button-spacing;

        row {
            @extend .model;

            border-spacing: $button-spacing;
            min-width: rem(150px);

            &:active {
                // There's an optical illusion because items are wider than
                // they are tall, so compensate by scaling y a little extra
                transform: scale(0.98, 0.96);
            }

            &:selected {
                @extend selection;
            }
        }
    }
}
```

wpkelso:
image

I've initially taken a flatter approach to this, but could re-add the shadow we had in the old stylesheet to the separators; I have no preference between the two approaches.

lib/Styles/Gtk/Sidebar.scss
wpkelso: I like it&mdash;the bounce is very satisfying. `selected` looks pretty awful in dark mode for text, though:
image

I think in this case, we probably want to use the standard `$fg-color` for selected instead of the shaded accent colour.

danirabbit: Ah yeah I'll propose a branch to fix the selected styles. I forgot to add the dark one

danirabbit (approved):
Let's go :rocket:

PR #861, opened 2025-05-06
Add `.monospace` class

wpkelso:


danirabbit (changes requested):
Don't forget to add a constant in `Granite.CssClass` and add an example to the Demo!

danirabbit:
@wpkelso It looks like [we did support it before](https://github.com/elementary/stylesheet/blob/ec0048924a19425c897182a87a12afe9e92bfa1e/src/gtk-4.0/_typography.scss#L125) and Adw supports it as well, so we might as well

danirabbit (changes requested):
Looks good, just one small thing

danirabbit (approved):
looks good! Nice job

PR #862, opened 2025-05-06
GraniteDemo: add color scheme switcher to sidebar

danirabbit:
Makes it easier to switch color schemes which comes in handy to test various views and see how they'll look in light and dark modes:

Screenshot from 2025-05-06 13 37 18


zeebok (approved):
This is slick. These stateful actions are pretty rad!

PR #864, opened 2025-05-13
Styles: Add `terminal` style class

wpkelso:
- Deprecate `Granite.STYLE_CLASS_TERMINAL`
- Add undocumented `terminal` style class

image


wpkelso:
Open question: Would this be better used as an internal style to a new `TerminalView` widget? Every time this is used in our repos, it's in the context of a `ScrollBox` with a `TextView` child.

danirabbit:
Yeah I actually wonder if we should have this public facing at all. Because we mainly use it in messagedialog. The sideload and installer cases I think I'd be okay with them using undocumented API or just supplying their own styles. I'm not sure this style is actually useful outside of system apps

wpkelso:
I'm happy with the style class being undocumented or internal. It's still worth it to abstract all of it into a widget, as even in the system apps and widgets the use cases for a terminal view are so similar that it should make the code cleaner. I've started another branch off this one to play around with that idea in #867

PR #865, opened 2025-05-13
accel_to_string: ⌘ → Super

danirabbit:
The original thought process behind using "⌘" was that:
1. No keyboards have the word "Super"
2. We can't use the Windows logo because of trademark
3. Some keyboards have the "⌘" symbol

But now, we have hardware partners like StarLabs which do print keyboards with the word "Super" on the super key. Plus the screen reader announces "⌘" as "Mac command key" instead of "Super". So it would probably be better at this point to switch to the word "Super"


leonardo-lemos (approved):

ryonakano:
Yes, that would be helpful!

ryonakano (approved):

wpkelso (approved):

PR #866, opened 2025-05-14
Settings: remove AccountsService fallback

lenemter:
It doesn't make sense to keep this, we use settings portal nowadays

### DIFF
--- a/lib/Widgets/Settings.vala
+++ b/lib/Widgets/Settings.vala
@@ -4,16 +4,6 @@
  */
 
 namespace Granite {
-    [DBus (name = "io.elementary.pantheon.AccountsService")]
-    private interface Pantheon.AccountsService : Object {
-        public abstract int prefers_color_scheme { get; set; }
-    }
-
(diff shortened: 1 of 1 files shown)


tintou:
Note that non-flatpak apps don't use portals

lenemter:
@tintou That's not true. Even non-sandboxed apps use Portals, that's the reason why if you launch granite apps without portals, you get 30 seconds startup delay. I just ported this patch to 7.6.0 to build it on my OS 7 and I can confirm that.

jeremypw:
Not sure why but Terminal headless unit tests fail on unstable build in Gtk4 with the message:
```
# DEBUG: Settings.vala:172: cannot use the portal, using the AccountsService: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.portal.Desktop was not provided by any .service files
not ok /application/action/new-window - FATAL-CRITICAL: Settings.vala:114: Could not connect: No such file or directory
```

So the AccountsService seems to be needed in that case, unless someone can advise how to get it to use the portal.

PR #867, opened 2025-05-14
Widgets: Add TerminalView

wpkelso:
Exploring an abstraction for a Terminal view, primarily for use in system apps and MessageDialogs

### DIFF
--- a/lib/Constants.vala
+++ b/lib/Constants.vala
@@ -131,6 +131,7 @@ namespace Granite {
      * When used with {@link Gtk.Label} this style includes internal padding. When used with {@link Gtk.TextView}
      * interal padding will need to be set with {@link Gtk.Container.border_width}
      */
+    [Version (deprecated = true, deprecated_since = "7.7.0", replacement = "Granite.CssClass.TERMINAL")]
     public const string STYLE_CLASS_TERMINAL = "terminal";
     /**
      * Style class for title label text in a {@link Granite.MessageDialog}
@@ -290,6 +291,15 @@ namespace Granite {
          * Style class for non-terminal text that uses a monospace font.
          */
         public const string MONOSPACE = "monospace";
+
+        /**
+         * Style class for {@link Gtk.Label} or {@link Gtk.TextView} to emulate the appearance of Terminal. This includes
+         * text color, background color, selection highlighting, and selecting the system monospace font.
+         *
+         * When used with {@link Gtk.Label} this style includes internal padding. When used with {@link Gtk.TextView}
+         * interal padding will need to be set with {@link Gtk.Container.border_width}
+         */
+        public const string TERMINAL = "terminal";
     }
 
     /**
--- a/lib/Widgets/TerminalView.vala
+++ b/lib/Widgets/TerminalView.vala
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2025 elementary, Inc. (https://elementary.io)
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+[Version (since = "7.7.0")]
+public class Granite.TerminalView : Gtk.ScrolledWindow {
+
+    public TerminalView () {
+        this.add_css_class (Granite.CssClass.TERMINAL);
+    }
+    
+    public TerminalView.with_textview (
+        Gtk.TextView view = new Gtk.TextView () {
+            cursor_visible = true,
+            margin_end = 6,
+            editable = false,
+            margin_start = 6,
+            monospace = true,
+            pixels_below_lines = 3,
+            wrap_mode = Gtk.WrapMode.WORD
+        }
+    ) {
+        this.child = view;
+    }
+    
+    // We may not need this, as we could use an un-editable TextView and
+    // simplify the API
+    public TerminalView.with_label (Gtk.Label label) {
+        this.child = label;
+    }
+}
--- a/demo/Views/CSSView.vala
+++ b/demo/Views/CSSView.vala
@@ -84,7 +84,8 @@ public class CSSView : DemoPage {
         card_box.append (card);
         card_box.append (card_checkered);
 
-        var terminal_label = new Granite.HeaderLabel ("\"terminal\" style class");
+        var terminal_label = new Granite.HeaderLabel
+        ("\"Granite.CssClass.TERMINAL\"");
 
         var terminal = new Gtk.Label ("[ 73%] Linking C executable granite-demo\n[100%] Built target granite-demo") {
             selectable = true,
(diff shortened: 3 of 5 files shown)


lib/Widgets/TerminalView.vala
```diff
@@ -0,0 +1,32 @@
...
+ * Copyright 2025 elementary, Inc. (https://elementary.io)
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+[Version (since = "7.7.0")]
+public class Granite.TerminalView : Gtk.ScrolledWindow {
+
+    public TerminalView () {
+        this.add_css_class (Granite.CssClass.TERMINAL);
+    }
```
danirabbit, line 11: This should be in a `construct` block. That's the block that runs for every constructor

lib/Widgets/TerminalView.vala
```diff
@@ -0,0 +1,81 @@
...
+    private void update_autoscroll () {
+        // FIXME: this disjoints the window closing and the application finishing
+        Idle.add (() => {
+            attempt_scroll ();
+            if (this.autoscroll) { return GLib.Source.CONTINUE; }
+            else {return GLib.Source.REMOVE; }
+        });
+    }
+
+    public void attempt_scroll () {
```
danirabbit, line 54: Does this need to be public? All public methods need documentation

lib/Widgets/TerminalView.vala
```diff
@@ -0,0 +1,81 @@
...
+    /* This is useful for specific feedback such as in MessageDialogs
+     *
+     * Might also be useful to have a way of enabling direct input for stdout or
+     * stderr
+     */
+    public void append_text (string text) {
+        buffer.insert_at_cursor (text, -1);
+    }
+
+    public void replace_text (string text) {
```
danirabbit, line 78: This should be a property and bind it (`BIDIRECTIONAL | SYNC_CREATE`) to the buffer text property: https://valadoc.org/gtk+-3.0/Gtk.TextBuffer.text.html

lib/Widgets/TerminalView.vala
```diff
@@ -0,0 +1,81 @@
...
+            adj.value = adj.upper;
+        }
+
+        prev_upper_adj = adj.upper;
+    }
+
+    /* This is useful for specific feedback such as in MessageDialogs
+     *
+     * Might also be useful to have a way of enabling direct input for stdout or
+     * stderr
```
danirabbit, line 72: This is going to show up in Valadoc :p The comments here should be about how to use the function

lib/Constants.vala
```diff
@@ -285,6 +286,15 @@ namespace Granite {
          * Style class for non-terminal text that uses a monospace font.
          */
         public const string MONOSPACE = "monospace";
+
+        /**
+         * Style class for {@link Gtk.Label} or {@link Gtk.TextView} to emulate the appearance of Terminal. This includes
+         * text color, background color, selection highlighting, and selecting the system monospace font.
+         *
+         * When used with {@link Gtk.Label} this style includes internal padding. When used with {@link Gtk.TextView}
+         * interal padding will need to be set with {@link Gtk.Container.border_width}
```
danirabbit, line 295: This documentation seems outdated now. This should probably just reference TerminalView?

lib/Widgets/MessageDialog.vala
```diff
@@ -169,7 +169,8 @@ public class Granite.MessageDialog : Granite.Dialog {
      */
     private Gtk.Grid message_grid;
 
-    private Gtk.Label? details_view;
+    //private Gtk.Label? details_view;
```
danirabbit, line 172: commented code

lib/Widgets/MessageDialog.vala
```diff
@@ -333,6 +335,6 @@ public class Granite.MessageDialog : Granite.Dialog {
             }
         }
 
-        details_view.label = error_message;
+        details_view.append_text (error_message);
```
danirabbit, line 338: This should still be a set/replaces. Can't change API out from under apps that are using it

danirabbit (changes requested):

stsdc:
If it's possible with this widget, maybe it would be worth to show colour palette?

Small snippets on how to archive those colours. e.g. for `[PASS], [FAIL], [WARNING]`.

lib/Widgets/TerminalView.vala
```diff
@@ -0,0 +1,74 @@
+/**
+ * Copyright 2025 elementary, Inc. (https://elementary.io)
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+[Version (since = "7.7.0")]
+public class Granite.TerminalView : Granite.Bin {
```
danirabbit, line 7: This needs documentation. See some of the other widgets for an example. Just needs a simple description of what this widget is and what it's used for

lib/Constants.vala
```diff
@@ -285,6 +286,12 @@ namespace Granite {
          * Style class for non-terminal text that uses a monospace font.
          */
         public const string MONOSPACE = "monospace";
+
+        /**
+         * Internal style class for {@link Granite.TerminalView} to emulate the appearance of Terminal. This includes
+         * text color, background color, selection highlighting, and selecting the system monospace font.
+         */
+        internal const string TERMINAL = "terminal";
```
danirabbit, line 294: I don't think this should be added as a constant here. We usually don't make constants for things that aren't used outside of their own class

lib/Widgets/TerminalView.vala
```diff
@@ -0,0 +1,74 @@
...
+
+        scrolled_window = new Gtk.ScrolledWindow () {
+            child = view,
+            hexpand = true,
+            vexpand = true,
+            hscrollbar_policy = NEVER,
+        };
+
+        child = scrolled_window;
+        add_css_class (Granite.CssClass.TERMINAL);
```
danirabbit, line 39: Instead of adding a style class, this should be the css name. Right now you have css nodes like `widget.terminal` but it should just be `terminal` So you need to remove this css class and add something like:
```
    class construct {
        set_css_name ("terminal-view");
    }
```

danirabbit (changes requested):
I still don't think this should be a public widget. The previous style class was really only used in MessageDialogs, it wasn't supposed to be a style that app developers were using regularly. So personally I'm -1 on this altogether. This isn't something I think we should be doing or supporting.

Tbh even in MessageDialogs it's been a bit of an antipattern. People have been really bad about not reading these messages, even when reporting an issue they won't expand the terminal view. So I'm kind of in favor of removing this feature altogether in a future version of MessageDialog and forcing developers to write some better error handling

PR #870, opened 2025-05-23
Styling: Add `.destructive` class for `button.text-button`

wpkelso:
image

Image excluded for now because I don't believe there is a precedent style.


danirabbit:
I don't think this works in the dark style:

Screenshot from 2025-07-01 11 51 56

Compared to the old stylesheet:

Screenshot from 2025-07-01 11 53 18

PR #872, opened 2025-05-28
Move Granite specific button styles to Granite Dir

wpkelso:
Rel: feedback on #870


lib/Styles/Granite/Button.scss
```diff
@@ -0,0 +1,83 @@
...
+            filter: brightness(95%);
+        }
+
+        &:disabled {
+            @include control-disabled;
+        }
+
+        &:hover {
+            filter: brightness(105%);
+        }
```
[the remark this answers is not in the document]
wpkelso: Yeah, that would make more sense

lib/Styles/Granite/Button.scss
```diff
@@ -0,0 +1,76 @@
+button {
+    .linked &,
+    .linked &.image-button,
+    .linked &.image-button.toggle {
+        @extend .text-button; 
+
+        // Overwrite changes made further down the cascade due to the way
+        // @extend works
+        padding: 0;
```
danirabbit, line 9: This padding removal seems like an error?

Main:
Screenshot from 2025-06-03 11 42 00
Screenshot from 2025-06-03 11 44 51

This branch:
Screenshot from 2025-06-03 11 42 23
Screenshot from 2025-06-03 11 44 16

danirabbit (changes requested):

lib/Styles/Granite/Button.scss
wpkelso: `button.text-button` sets a padding of `rem(2px) rem(4px)`, which then gets inherited when we extend it. This is supposed to override that and reset the padding. Before, we got around this by setting the padding for `.text-button` in a block specific to `.text-button`, and setting everything else in a shared block. 

Compiled as it currently is, the padding is not being overridden. I'm not really sure why, though. Sheets might be cascading in a way I'm not expecting?

image

danirabbit: I see what you're saying. Because we want text buttons to have more horizontal padding and image buttons to have even padding all the way around.

Maybe what we need to do instead then is have a `.raised` style that just handles the borders and have .linked and .text-button extend that, separately

wpkelso: I've split it out using a placeholder class. For some reason the checked and active selectors still had to be overriden

PR #876, opened 2025-06-16
StyleManager: Allow the GLib.Application to be null

tintou:
This is for instance the case for the portal application.


danirabbit (approved):
Thanks for fixing this!

PR #892, opened 2025-07-09
Release 7.7.0

danirabbit:
There's a ton of useful new widgets here and also a lot of deprecations. I'd like approves from multiple people here just to make sure since Granite is very consequential :)

### DIFF
--- a/data/granite.metainfo.xml.in
+++ b/data/granite.metainfo.xml.in
@@ -27,7 +27,7 @@
   <update_contact>contact_at_elementary.io</update_contact>
 
   <releases>
-    <release version="7.7.0" date="2025-09-28" urgency="medium">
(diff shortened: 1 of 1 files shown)


ryonakano:
I think we shouldn't use the deprecated by ourselves?

```
user@test:~/work/granite$ git grep -n Granite.STYLE_CLASS_FRAME -- demo
demo/Views/HyperTextViewGrid.vala:32:        hypertext_scrolled_window.add_css_class (Granite.STYLE_CLASS_FRAME);
demo/Views/WelcomeView.vala:48:        listbox.add_css_class (Granite.STYLE_CLASS_FRAME);
user@test:~/work/granite$ 
```

ryonakano:
It fails to build with documentation so I proposed #910

lenemter (approved):

PR #896, opened 2025-07-09
MapsView: add marker demos

### LINKED ISSUE #894: Shumate: A big circle in sky blue is shown
### What Happened?

I'm not sure where I should report this, but a big circle in sky blue is shown under the location picker with our stylesheet:

Image

### Steps to Reproduce

1. Launch Tasks
2. Create a new task
3. Click "Set Location"
4. See a big circle in sky blue is shown under the location picker

### Expected Behavior

The big circle in sky blue is not shown under the location picker.

### OS Version

Early Acccess

### OS Architecture

amd64 (on most hardwares)

### Session Type

Classic Session (X11, This is the default)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

### Hardware Info

_No response_

danirabbit:
Fixes #894 

Move marker styles into the demo as an example instead of styling all markers

Screenshot from 2025-07-09 12 15 04


tintou:
In fact, I believe that we should use [Shumate.Point](https://gnome.pages.gitlab.gnome.org/libshumate/class.Point.html) for this

PR #906, opened 2025-08-20
Styles: Add Undershoot styles for Scrolled Window

### LINKED ISSUE #806: Add underscroll with style class
### Problem

I've seen devs asking how they can add underscroll to a specific view like they get in popovers (see: https://github.com/hezral/quickword/issues/5); however, in the past we've been concerned about being able to reliably add it by default (due to differing view background colors).

### Proposal

We could at least add it as a style class that can be added so devs can opt into it.

### Prior Art (Optional)

Popovers have underscroll fade to the background color.

wpkelso:
image

I believe this would close #806


danirabbit (changes requested):
If we're adding a global undershoot here we should remove:
- [ ] The duplicate one in Popovers: https://github.com/elementary/granite/blob/fde869baeb961774afd40d7e6af12d035da0a528/lib/Styles/Gtk/Popover.scss#L20
- [ ] and the one in ToolBox: https://github.com/elementary/granite/blob/main/lib/Styles/Granite/ToolBox.scss

Though personally I like the fade out style of undershoot in framed scrolled windows better. I think especially when you've got a situation like a card with a scrolled window in it, then this style of undershoot gives you weird conflicting levels

PR #943, opened 2026-01-29
Rename RDNN to io.elementary.contractor

### LINKED ISSUE #30: wrong RDNN for dbus service
The dbus service file is still named `org.elementary.contractor.service`, everything is being moved to `io.elementary.*` namespace.

drowsymike:
In discussions elementary/contractor#31 and elementary/contractor#41, @ryonakano pointed out the need to transition from `org.elementary.Contractor` to `io.elementary.Contractor` as part of the RDNN rename for OS 9

This PR updates the occurrences in lib/Services/ContractorProxy.vala to match the changes in the Contractor daemon

Note: This change is part of a cross-repo transition and should be merged in coordination with the Contractor PR

### DIFF
--- a/lib/Services/ContractorProxy.vala
+++ b/lib/Services/ContractorProxy.vala
@@ -68,7 +68,7 @@ namespace Granite.Services {
         string icon;
     }
 
-    [DBus (name = "org.elementary.Contractor")]
+    [DBus (name = "io.elementary.Contractor")]
     internal interface ContractorDBusAPI : Object {
         public signal void contracts_changed ();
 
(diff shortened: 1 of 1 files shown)


danirabbit (commented):
Since this is an API breaking change we also need to bump the library major version. So this would have to come after we change to `granite-9`

ryonakano:
Since now we bumped to `granite-9` in the main branch, maybe we can go ahead and merge this?

danirabbit (approved):

ryonakano (approved):

ryonakano:
@danirabbit

> Since this is an API breaking change we also need to bump the library major version. So this would have to come after we change to `granite-9`

I'm thinking this back after merging this PR; is this really an API breaking change? As I understand correctly apps that want to use Contractor communicate with it through the APIs in Granite. Here, all of our apps that use Contractor still uses Granite 6 because they have not been ported to GTK 4. So, I wonder possibly we need to rather backport this to Granite 6, otherwise we can never release new version of Contractor with the new RDNN until we port these apps to GTK 4 and Granite 9.

ryonakano:
> So, I wonder possibly we need to rather backport this to Granite 6, otherwise we can never release new version of Contractor with the new RDNN until we port these apps to GTK 4 and Granite 9.

Opened https://github.com/elementary/granite/pull/1051 and https://github.com/elementary/granite/pull/1052.

PR #971, opened 2026-03-23
Demo/Controls: add more image button examples

danirabbit:
Make sure that we have equivalent examples for text and image buttons:

Screenshot from 2026-03-23 10 21 49


zeebok:
Should the call buttons look different from the other icon buttons due to them from having the css styles applied?

danirabbit:
@zeebok eventually yes, but we don't have styles for it yet

zeebok (approved):

Issue #471, opened 2021-03-02
Play a sound for Granite.MessageDialog


hanaral:
## Problem

Many users may react slowly or not be able to immediately notice an error popup. The user has to be staring at their to notice a popup, which is pretty bad if it is a critical error.

## Proposal

Gala should play a crisp 'dunn' sound effect when a Granite.MessageDialogue with an error appears, and for when a Granite error views appears. 
This should be done on an opt-in basis to avoid accidentally confusing users.
I'd also suggest adding a lighter sound effect for authentication popups.

tl;dr
Error views and granite error dialogues should have unique audio cues 

## Prior Art

_Maybe I'll add some later_

danrabbit:
Since this would need context, it would have to happen in Granite. The Window Manager isn't aware of the dialog's context, just that it's a dialog.

This would be hard to enforce consistently since apps may choose not to use Granite.MessageDialog to show errors

cassidyjames:
I’m pretty :-1: about dialogs playing sounds in general; notifications are the platform way to notify for background tasks, and dialogs generally assume the user is present. I believe GNOME used to (optionally) play an FD.o sound when dialogs were thrown, but it was removed because it was annoying. It's probably worthwhile to investigate when and why they removed it if you want to petition for it to be re-added.

hanaral:
This isn't about notifying the user about a backrgound activity, you've already pointed out that the general way to notify is through notifications. 
This is supposed to be a way for there to automatically be an auditory cue to fit with something of importance on screen - yes, I could use a screen reader, but what happens to the using accessibility features as features?
I think that this should just be something the user can disable like window animations.

Also is there no way for granite.messagedialog to emit a dbus signal or something that gala or even the notification server(which already plays sounds) can catch?

PR #972, opened 2026-03-23
Create AlertDialog

danirabbit:
A new AlertDialog based on PortalDialog to replace MessageDialog since it's Gtk.Dialog based on Gtk.Dialog is deprecated.

Screenshot from 2026-03-23 12 10 36

Todo:
- [ ] button box isn't always valigned to the end. Bug in ToolBox?
- [ ] Handle no secondary icon
- [ ] Improve documentation
- [ ] Mark MessageDialog deprecated
- [ ] Wrap buttons if the text is long

Things to figure out:
- [ ] Do we require an icon to construct? Adw and GTK don't. Alerts on Android and iOS don't either
- [ ] Adw and GTK both have an async API https://valadoc.org/gtk4/Gtk.AlertDialog.choose.html and https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/method.AlertDialog.choose.html
- [ ] Adw.AlertDialog uses a `string` ID which is good for making code more maintainable and on our side it makes it easer t use GLibAction for things like action enabled, but in places where we want to use alerts (like portals) an int response is required (and Gtk uses int still). Which use case do we optimize for?
- [ ] Gtk automatically creates a "Close" button if no buttons are added. Probably a good feature
- [ ] Adw.AlertDialog allows removing buttons. Is that good/necessary? Examples?
- [ ] Should button labels be settable?
- [ ] Should button styles be settable?
- [ ] Property and function names
- [ ] Padding for content area. Should we pack everything into a scrolled by default?
- [ ] Do we want to keep the error output area/details view feature? Is this an anti-pattern? Should it just be a separate widget with no collapsing like https://github.com/elementary/granite/pull/867?
- [ ] Should alert dialogs make noise? https://github.com/elementary/granite/issues/471

### DIFF
--- a/demo/Views/DialogsView.vala
+++ b/demo/Views/DialogsView.vala
@@ -115,6 +115,8 @@ public class DialogsView : DemoPage {
         dialog.add_button ("_Cancel", "cancel");
         dialog.add_button ("_Take Action", "accept", SUGGESTED);
 
+        dialog.set_response_enabled ("accept", false);
+
         dialog.response.connect ((response_id) => {
            if (response_id == "accept") {
                toast.send_notification ();
@@ -124,6 +126,10 @@ public class DialogsView : DemoPage {
         });
 
         dialog.present ();
+
+        Timeout.add_once (3000, () => {
+            dialog.set_response_enabled ("accept", true);
+        });
     }
 
     private void show_message_dialog () {
--- a/demo/Views/DialogsView.vala
+++ b/demo/Views/DialogsView.vala
@@ -15,6 +15,10 @@ public class DialogsView : DemoPage {
     construct {
         var dialog_button = new Gtk.Button.with_label ("Show Dialog");
 
+        var alert_button = new Gtk.Button.with_label ("AlertDialog");
+
+        var alert_button_pro = new Gtk.Button.with_label ("AlertDialog+ Pro Ultra Max");
+
         var message_button = new Gtk.Button.with_label ("Show MessageDialog");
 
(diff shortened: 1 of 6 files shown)


bugaevc:
I have experience implementing another dialog API at `$dayjob`, also modeled after modern GTK & libadwaita dialogs.

All of GTK's recent dialogs have an async API, as opposed to signal-based `response.connect` + `present`. See [`Gtk.FileDialog`](https://valadoc.org/gtk4/Gtk.FileDialog.html) for example.

In practice, if you want to obtain a response, you await (`yield`) the call that shows the dialog. For a message dialog, I'd expect this to return the chosen response either as a string or an int. If you don't want a response and just want to let the user know about something, you do `dialog.choose.begin ()`.

Triggering the cancellable passed to the async call causes the GTK dialogs to pop down and throw [`Gtk.DialogError.CANCELLED`](https://valadoc.org/gtk4/Gtk.DialogError.html). However: note that the implicit `GLib.Task` Vala creates when you write an async method has the [`check_cancellable`](https://valadoc.org/gio-2.0/GLib.Task.set_check_cancellable.html) flags set (which is the default... unfortunately). The effect of this is that the async method will be seen as throwing [`GLib.IOError.CANCELLED`](https://valadoc.org/gio-2.0/GLib.IOError.html) if the cancellable has been triggered, even if the implementation did react differently, such as in this case by throwing a different error, `Gtk.DialogError.CANCELLED`. I don't immediately know of a way to work around this, except writing the function in C.

PR #973, opened 2026-03-30
Move design goals to be in-tree

wpkelso:
This file formerly lived in GitHub's wiki system. Moving it in tree enables actual proposal/approval processes via pull request, improves portability of the repo, and improves discoverability of the standards.

### DIFF
--- a/lib/Styles/STANDARDS.md
+++ b/lib/Styles/STANDARDS.md
@@ -1,3 +1,5 @@
+# Granite Design System Standards
+
 The following is a summary of the design targets and goals for Granite. 
 When adding new styles, they should meet these goals to form a cohesive and predictable whole.
 
@@ -8,7 +10,7 @@ When adding new styles, they should meet these goals to form a cohesive and pred
   - In cases where a Granite widget is not available and an Adw widget needs to be used, styling of those Adw widgets should be done primarily at the application level. Adw widgets can be considered for styling within Granite on a case-by-case basis.
   - Adw widgets should not be added to Granite.Demo.
 
-# Levels Theory
+## Levels Theory
 
 Ideally, widget backgrounds should be styled according to their "level" or "elevation".
 The further away a widget is in the interaction hierarchy, the darker it's background color should be.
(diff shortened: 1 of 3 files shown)


lib/Styles/STANDARDS.md
```diff
@@ -0,0 +1,122 @@
...
+
+For example, interactive widgets like buttons and entries should be the lightest, while containers like sidebars should be some of the darkest widgets.
+Additionally, `backdrop` and `disabled` states should be considered "further away" in the interaction hierarchy, and should thus be on the darker end of the scale.
+
+## Color
+
+Colors should all be derived from one of the colors in the Granite palette &mdash; an extended version of the [elementary brand palette](https://elementary.io/brand), pure white, or pure black.
+Any other necessary shades or highlights should be achieved using `color` functions, and overlaying translucent black or white elements.
+
+## (Proposal) Blur & Transparency
```
danirabbit:
```suggestion
## Blur & Transparency
```

lib/Styles/STANDARDS.md
```diff
@@ -0,0 +1,122 @@
...
+
+Blur and transparency should be used primarily to communicate that an element is "transient", or overlays another element. Good examples of elements that display these traits are system notifications, OSDs, and &mdash; under default settings &mdash; the system dock.
```
danirabbit: I don't think we need to include this aside. We already call out that transparent elements should have a non-transparent style for reduce transparency in the a11y section

```suggestion
Blur and transparency should be used primarily to communicate that an element is "transient", or overlays another element. Good examples of elements that display these traits are system notifications, OSDs, and the system dock.
```

lib/Styles/STANDARDS.md
```diff
@@ -0,0 +1,122 @@
...
+
+Where reasonable, border radii should increase as they nest. For instance, button radii are 4px so they nest neatly into 8px radii windows, popovers, and OSDs.
+To facilitate this, `$window_radius` has been defined as a variable, and all other radii should be derived from it (e.g. `$button_radius = $window_radius / 2`).
+
+## States & Pseudo Classes
+
+### Active
+
+Interactive widgets must respond to activation.
+In many cases, this response will be through a skeuomorphic animation.
```
danirabbit: This is not really what `skeuomorphic` means. Most of the activation animations we're using are springy and don't reflect how physical buttons work. The slider animations are definitely not how sliders work in analog devices

```suggestion
```

lib/Styles/STANDARDS.md
```diff
@@ -0,0 +1,122 @@
...
+- If a `high-contrast` style class exists, it must pass WCAG AAA
+- When transparency is used, an opaque `high-contrast` style class must also be added
+- Excepting app icons, symbolic icons should be forced when `high-contrast` is used.
+
+### Reduced Motion
+
+- Interactive components should still be stateful, but translation should not occur between states. Buttons should press in, but they shouldn't bounce.
+- Widgets should still react to multitouch as normal since they behave more like drag handles, but don't animate when interacted in other ways like keyboard shortcuts
+
+### Texting Scaling
```
danirabbit: this got an `ing` added somehow
```suggestion
### Text Scaling
```

danirabbit (changes requested):
emdashes shouldn't have spaces around them :)

danirabbit (approved):

PR #978, opened 2026-04-26
Fix radius and add subtle sculpting to `.keycap`

wpkelso:
- Adjust border radius and padding to align with 2px grid
- Add some subtle shading to emulate sculpting because I think its fun

## Before

image

image

## After

image

image


danirabbit (approved):
Looks cute :)

PR #980, opened 2026-04-27
Styles: Adjust and even-out padding for buttons and entry

wpkelso:
Might be a little contentious? I think the little bit of extra room on buttons gives a little better rhythm when types are mixed, but I tried not to turn them into balloons.

- Squared up image buttons linked to entries
- Evened out padding

## Before

image

image

## After

image

image


lib/Styles/Gtk/Button.scss
```diff
@@ -9,8 +9,8 @@ button {
         @extend .circular;
 
         padding: rem(4px);
-        min-height: 16px;
-        min-width: 16px;
+        min-height: 24px;
+        min-width: 24px;
```
danirabbit: I think these are min 16px because of icons. Probably don't want to touch the min size here and just change padding?

danirabbit (commented):
Looks like something has gone wrong here. I'm guessing you're using some kind of text scaling? These don't match up with default text scaling on my end

Image

wpkelso:
@danirabbit Yeah I forgot I use text scaling. It's properly aligned at 1x now, but because icons resize at thresholds and don't add any padding to compensate between icon and text size differences, image buttons linked or inline with a text button will be rectangular. Setting a minimum size of 16px for other types of buttons and such fixes the problem on the minimized end, but I'm not sure how to compensate on the magnified end.

Ex.
image

image

image

(Looks like we might have some double border issue with the linked button borders 🙃 )

lib/Styles/Gtk/Button.scss
```diff
@@ -8,9 +8,10 @@ button {
     &.image-button {
         @extend .circular;
 
-        padding: rem(4px);
+        padding: calc(0.5rem - 2px);
```
danirabbit, line 11: Can we make a constant for this and comment what the magic number here is?

lib/Styles/Gtk/Button.scss
```diff
@@ -43,7 +44,9 @@ button {
     &.text-button {
         @extend %raised;
 
-        padding: rem(2px) rem(4px);
+        min-height: 16px;
+        min-width: 16px;
+        padding: calc(0.5rem - 2px);
```
danirabbit, line 49: In text buttons its intentional for x padding to be double

danirabbit (changes requested):

danirabbit:
Looks good I'd just like to see this abstracted so it stays in sync

danirabbit (approved):
I'm just gonna merge this and we can revisit abstracting values. This is a big improvement on its own

PR #981, opened 2026-04-29
Styles: `suggested` class for text buttons

wpkelso:
image

image


danirabbit (approved):
Looks good. Nice work!

PR #983, opened 2026-06-11
SwitchModelButton: Focus the switch

### LINKED ISSUE #807: Consider button focus rings on immediate child
## Problem

We have some issues with focus rings being outside of buttons like clipping in linked styles or when there isn’t technically any margin around the button or some issue with z axis ordering or clipping in overlays. Lots of little edge cases that make focus rings outside of the button break

## Proposal

Maybe instead, we should apply the focus style to the immediate child of the button (likely a label or icon) as a background 

## Prior Art

This would look more similar to the flat button focus styles and indeed maybe we could combine a lot of the focus style code this way as well

jeremypw:
Fixes #807 

Prompted by https://github.com/elementary/settings-network/issues/450

### DIFF
--- a/lib/Widgets/SwitchModelButton.vala
+++ b/lib/Widgets/SwitchModelButton.vala
@@ -36,6 +36,8 @@ public class Granite.SwitchModelButton : Gtk.ToggleButton {
     }
 
     construct {
+        focusable = false;
+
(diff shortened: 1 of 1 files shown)


danirabbit:
I think we probably want to leave focus how it is and fix this in the stylesheet instead

PR #989, opened 2026-08-20
ListItem: Ellipsize instead of wrap

leolost2605:
I think in lists you usually want to have items of the same size so use ellipsize instead of wrapping.

This also improves performance (by a lot) because list views need to know the height of items to guess the size of the full list and layouting text is very expensive whereas this way you have a fixed height. This also improves predictability because the listview won't get confused when suddenly an item takes up 10 times the height of others (without this you get a sudden change in the scrollbar position and more weirdeness when you scroll an item with a much bigger height into view).

### DIFF
--- a/lib/Widgets/ListItem.vala
+++ b/lib/Widgets/ListItem.vala
@@ -76,13 +76,13 @@ public class Granite.ListItem : Gtk.Widget {
         var label = new Gtk.Label ("") {
             hexpand = true,
             vexpand = true,
-            wrap = true,
+            ellipsize = END,
             xalign = 0,
             mnemonic_widget = this
         };
 
         description_label = new Gtk.Label ("") {
-            wrap = true,
+            ellipsize = END,
             xalign = 0
         };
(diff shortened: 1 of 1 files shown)


lib/Widgets/ListItem.vala
```diff
@@ -73,14 +73,14 @@ public class Granite.ListItem : Gtk.Widget {
         var label = new Gtk.Label ("") {
             hexpand = true,
             vexpand = true,
-            wrap = true,
             xalign = 0,
-            mnemonic_widget = this
+            mnemonic_widget = this,
+            ellipsize = END,
```
danirabbit, line 85: Can we add a comment here about why we don't wrap?

danirabbit (changes requested):

lib/Widgets/ListItem.vala
leolost2605: I added a comment. Lmk if I can improve clarity or if it's too wordy :sweat_smile:

danirabbit (approved):

PR #991, opened 2026-08-25
Add Granite.Symbol

### LINKED ISSUE #919: Animated Icons
GTK 4.22 will ship with a new native SVG renderer that supports animation: https://blogs.gnome.org/gtk/2025/10/23/svg-in-gtk/

We should resource animated icons here and make it easy for developers to use them in their apps

danirabbit:
Fixes #919 
Screenshot from 2026-09-07 12 45 35

Todo:
- [ ] Documentation

Examples:
 - https://github.com/elementary/panel-bluetooth/pull/238
 - https://github.com/elementary/panel-sound/pull/332


lib/Widgets/Symbol.vala
```diff
@@ -0,0 +1,78 @@
...
+    // Disabled state represented by a slash
+    public const string DISABLED = "disabled";
+    // e.g. paired, connected, needs attention
+    public const string ACTIVE = "active";
+}
+
+public class Granite.Symbol : Granite.Bin {
+    public string resource_path { get; construct; }
+
+    public int pixel_size {
```
danirabbit, line 18: I wonder do we actually want this or should we have an enum or at least constants? The icon sizes we use symbolics in so far are:

1. 16px for buttons
2. 24px for tool bars
3. 32px in lists/selections (Screenshot, Settings)
4. 48px in Notifications

zeebok: I would say follow the SymbolState and use namespaced constants. That way we have important sizes but if we need to handle more it is possible

lib/Symbols/CONTRIBUTING.md
```diff
@@ -0,0 +1,22 @@
+## Canvas Size
+
+16px x 16px
+
+## Fill
+
+10% opacity
```
danirabbit, line 7: So we had talked about going duotone by default. It doesn't look like we can have outline or solid style options at this time. Do we still wanna do duotones if that's our only option?

PR #994, opened 2026-08-27
 ListItem: Add a setup_menu signal 

### LINKED ISSUE #992: ListItem: setup_menu () function
### Problem

Sometimes we need to change menu actions (like setting enabled/disabled) depending on the selected item

### Proposal

A function that runs before the popover is shown where we can make changes

### Prior Art (Optional)

https://valadoc.org/libadwaita-1/Adw.TabView.setup_menu.html

https://github.com/elementary/terminal/blob/c2be61f5bc1d017f44520be044569579eab2dd22/src/Widgets/TerminalView.vala#L266

leolost2605:
Fixes #992 

I actually also wanted this in the past for another reason because I wanted to select the item when the menu was opened.

I added a little demo if somebody has a better idea for a demo lmk.

### DIFF
--- a/demo/Views/ListsView.vala
+++ b/demo/Views/ListsView.vala
@@ -4,6 +4,8 @@
  */
 
 public class ListsView : DemoPage {
+    private int menu_opened_count = 0;
+
     construct {
         title = "Lists & Grids";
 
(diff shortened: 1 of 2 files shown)


danirabbit (approved):
Works as expected, nice!

PR #996, opened 2026-08-27
ListItem: add version tag to setup_menu

danirabbit:


leolost2605 (approved):
I didn't add it because I thought since 9 is a breaking version bump we don't need it but I don't know about this stuff :)

danirabbit:
Hm yeah that's a good point but we do still have them for versions before 9 so I think probably just for completeness in the docs :woman_shrugging:

PR #997, opened 2026-08-28
Styles: Use purple grid styling for development windows

wpkelso:
Exploring an idea for putting "development" styling in headerbar areas like gnome. Expands on the work done in #988 .

image
image

<details>
<summary>Incomplete iteration history</summary>
image
image
image
</details>


teamcons:
i dont want to catch your thunder, but wouldnt both be too much?

id have been fine with discussing which one between two ideas or even the concept be rejected,

but redundancy for a feature that elementary OS doesnt use itself (there is no devel option in any apps ive seen) may be a bit much

teamcons:
thats fire tho

wpkelso:
@teamcons This wouldn't create a parallel style class, but rather change the extant one. I realize I did not explain it very clearly in the op

teamcons:
Ah!
But then why merge it if you werent on board with the style?

danirabbit:
I kinda like the stripes tbh. Less busy

Screenshot from 2026-08-28 15 04 51

wpkelso:
@teamcons I thought it better to treat it as two separate questions: "should it exist?", and "what should it look like?" I decided to treat 988 as asking the first question, to which my answer is "yes", so I merged it. This PR, then, asks the second question. The yellow outline does the job well, but I think there is also room to explore what else it _could_ look like. 

Question 2 is dependent on question 1, but question 1 is in no way dependent on question 2 having an answer. If we don't need that answer right away, then I don't see why block on something we don't need to block on.

teamcons:
Yeah thats very fair

The thought process on the outline was:
-Adwaita already did warning stripes in headerbar and devel icon, so just doing same but elementary felt uninspired/cheaps
-It was for Jorts, whose design has the headerbar being an integral part of the rest, destroying the look for the sake of warning felt bad
-KDE used with Oxygen to have a cool glow aura around focused windows, which really helped distinguish non-focused with focused. The idea was to replicate it with the warning color of those "DO NOT CROSS" bands. Like you were in the warning zone. You had a really cool draft for Jorts using that color and metaphor

leaving yall to decide, got no bacon in this. Whatever wpkelso signs off on is a good choice anyway cuz that otter got game

PR #999, opened 2026-08-28
Styles use dark style media query

### LINKED ISSUE #998: Replace dark stylesheet with media query
We have media queries now so we can replace all the stylesheet loading and unloading etc with:

```
@media (prefers-color-scheme: dark) {
}
```

danirabbit:
Fixes #998


danirabbit:
Need to switch everything to using CSS vars and functions first

PR #1000, opened 2026-08-28
Styles: export css style palette variables

danirabbit:
Use CSS variables for palette colors and bump GTK requirement


wpkelso (approved):
Css variables lesgoooooooooo 🚀

PR #1005, opened 2026-08-29
styles/exported: use CSS vars

danirabbit:
This doens't fix all the variable usage yet, but it does fix exported variables


lib/Styles/Granite/Button.scss
```diff
@@ -13,13 +13,13 @@ button {
         }
 
         &:active {
-            background-color: mix($error_color, $STRAWBERRY_900, 90%);
+            background-color: #{'mix(var(--error-color), var(--STRAWBERRY_900), 0.1)'};
```
wpkelso: I think we could probably switch this to use `color-mix()` while we're here, rather than the gtk `mix()` extension. Is there any advantage to using `mix()` here?

https://docs.gtk.org/gtk4/css-properties.html#css-colors

lib/Styles/Gtk/Popover.scss
```diff
@@ -79,18 +79,18 @@ popover {
...
-        background: rgba($fg_color, 0.05);
+        background: #{'alpha(var(--fg-color), 0.05)'};
         padding: rem(9px);
     }
 
     // Focus and hover are synonymous in Popovers
     &:hover,
     &:focus-visible {
-        background: rgba($fg_color, 0.1);
+        background: #{'alpha(var(--fg-color), 0.1)'};
```
wpkelso: Would a `color-mix()` with `transparent` achieve the same thing here?

wpkelso (commented):
Looks overall good, just a couple of small questions I still have. I think overall we should be shifting away from the gtk css color extensions toward native css functions, but I don't know if you consider that in-scope for this PR or not.

lib/Styles/Granite/Button.scss
danirabbit: Oh cool I didn't know GTK supports color-mix these days will do

danirabbit: @wpkelso is this the intended mix or was it the other way around?

danirabbit:
@wpkelso I agree and I think it makes sense to switch the ones that we're touching at least. Good call

danirabbit:
@wpkelso okay this now uses color-mix for all changed lines

wpkelso (approved):
Lets go 🚀

PR #1006, opened 2026-08-29
StyleManager: switch to new GTK color scheme preference

danirabbit:
Switch to the new way to set color scheme preferences. This makes media queries work

### DIFF
--- a/lib/StyleManager.vala
+++ b/lib/StyleManager.vala
@@ -63,7 +63,7 @@ public class Granite.StyleManager : Object {
 #if INCLUDE_GTK_STYLESHEETS
         gtk_settings.gtk_theme_name = "Granite-empty";
 #endif
-        gtk_settings.notify["gtk-application-prefer-dark-theme"].connect (set_provider_for_display);
+        gtk_settings.notify["gtk-interface-color-scheme"].connect (set_provider_for_display);
         set_provider_for_display ();
 
         var granite_settings = Granite.Settings.get_default ();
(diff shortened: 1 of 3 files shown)


lib/StyleManager.vala
```diff
@@ -63,14 +63,13 @@ public class Granite.StyleManager : Object {
 #if INCLUDE_GTK_STYLESHEETS
         gtk_settings.gtk_theme_name = "Granite-empty";
 #endif
-        gtk_settings.notify["gtk-application-prefer-dark-theme"].connect (set_provider_for_display);
+        gtk_settings.notify["gtk-interface-color-scheme"].connect (set_provider_for_display);
         set_provider_for_display ();
 
         var granite_settings = Granite.Settings.get_default ();
-        granite_settings.notify["prefers-color-scheme"].connect (update_color_scheme);
+        granite_settings.notify["prefers-color-scheme"].connect (set_provider_for_display);
```
leolost2605: This was removed wasn't it?

leolost2605 (changes requested):
LGTM, only one little comment. Also does this have any effect on non desktop session like greeter? Not sure whether portals are available there or how that's handled?

lib/StyleManager.vala
danirabbit: Good catch!

danirabbit:
@leolost2605 This won't effect anything still using Granite 7 :) But I think if the greeter needs special access outside the portal, it should probably do its own implementation and let the library cater to the typical sandboxed case

manu20247:
I'm very happy with the progress of the OS. I'd like to know if there will be an option to "turn off" the transparency of the top bar and other parts of the system like notifications and alerts. Of course, I found it very attractive, but it would be good to be able to turn it off completely. And while we're at it, could you please put the screen brightness control in the quick access menu above the system volume control? 😊 I'm really looking forward to EOS 9.

danirabbit:
@manu20247 glad to hear it!

But please don't ask support questions on pull requests. You can ask for help on Discussions pages or in Discord here: https://discord.gg/pCenUBu

leolost2605 (approved):
LGTM

PR #1008, opened 2026-08-29
Add `adw-entry-edit-symbolic` icon

### LINKED ISSUE #1003: Missing icon (alias) `adw-entry-edit-symbolic`
I was testing out the newly released platform 9 and I noticed the pencil icon was missing in [Adw.EntryRow](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/class.EntryRow.html) fields:

Image

The problem appears to be that the icon name used in libadwaita switched to a more specific `adw-entry-edit-symbolic` instead of `document-edit-symbolic` used before, which worked fine in 8.x because it's a standard icon name:

Image

ztefn:
Fixes #1003


lib/Icons/icons.gresource.xml
```diff
@@ -2,6 +2,7 @@
 <gresources>
   <gresource prefix="/io/elementary/granite">
```
danirabbit, line 3: Can you also add the line for `document-edit-symbolic`?

danirabbit (commented):
Thanks! Just one small request 🙏🏼

lib/Icons/icons.gresource.xml
ztefn: Sure, done.

danirabbit (approved):
Thanks!

PR #1009, opened 2026-08-29
HeaderLabel: add justify property

### LINKED ISSUE #977: HeaderLabel: Expose label xalign property
### Problem

I cannot properly, fully center the placeholder text here. The wrap is done assuming the text is left-aligned, leaving the placeholder here strangely bungled

the style class being deprecated, it would be nifty to use headerlabel here, but it lacks the granularity of Gtk.Label

Image

### Proposal

expose the base Gtk.Label xalign property for Granite.HeaderLabel
I suppose one could want the wrap property too to control how it is displayed

current workaround is to
```
var a = (Gtk.Label)placeholder.get_last_child ();
a.xalign = 0.5f;
```

### Prior Art (Optional)

_No response_

nityanand123gupta:
## Description

`Granite.HeaderLabel` hardcoded `xalign = 0` on both its primary and secondary internal `Gtk.Label`s, with no way to change it short of reaching into the widget tree via `get_last_child()`.

## Change

- Added a public `xalign` property (`since = "9.0.0"`, matching the current unreleased `meson.build` version), defaulting to `0` to preserve existing behavior.
- Bound it to both the primary label and the secondary label (when present) via `bind_property (..., SYNC_CREATE)`, following the same pattern already used for `ellipsize`.

Note: I don't have a Vala/GTK4 build environment to compile locally, so I've relied on matching the existing binding pattern exactly — happy to iterate on CI/review feedback.

Fixes #977

### DIFF
--- a/lib/Widgets/HeaderLabel.vala
+++ b/lib/Widgets/HeaderLabel.vala
@@ -56,6 +56,12 @@ public class Granite.HeaderLabel : Gtk.Widget {
     [Version (since = "7.8.0")]
     public Pango.EllipsizeMode ellipsize { get; set; default = NONE; }
 
+    /**
+     * The horizontal alignment of the label text, forwarded to the underlying {@link Gtk.Label}.
+     */
+    [Version (since = "9.0.0")]
+    public float xalign { get; set; default = 0; }
+
     private Gtk.Label? secondary_label = null;
     /**
      * Optional secondary label string displayed below the header
@@ -76,12 +82,12 @@ public class Granite.HeaderLabel : Gtk.Widget {
             } else if (value != null) {
                 secondary_label = new Gtk.Label (value) {
                     use_markup = true,
-                    wrap = true,
-                    xalign = 0
+                    wrap = true
                 };
                 secondary_label.add_css_class ("subtitle");
 
                 bind_property ("ellipsize", secondary_label, "ellipsize", SYNC_CREATE);
+                bind_property ("xalign", secondary_label, "xalign", SYNC_CREATE);
 
                 secondary_label.set_parent (this);
             }
@@ -107,8 +113,7 @@ public class Granite.HeaderLabel : Gtk.Widget {
 
     construct {
         var label_widget = new Gtk.Label (label) {
-            wrap = true,
-            xalign = 0
+            wrap = true
         };
         label_widget.add_css_class ("heading");
         label_widget.set_parent (this);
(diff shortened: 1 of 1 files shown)


lib/Widgets/HeaderLabel.vala
```diff
@@ -56,6 +56,12 @@ public class Granite.HeaderLabel : Gtk.Widget {
     [Version (since = "7.8.0")]
     public Pango.EllipsizeMode ellipsize { get; set; default = NONE; }
 
+    /**
+     * The horizontal alignment of the label text, forwarded to the underlying {@link Gtk.Label}.
+     */
+    [Version (since = "9.0.0")]
+    public float xalign { get; set; default = 0; }
```
danirabbit: I think we probably want to use `justify` rather than xalign https://valadoc.org/gtk+-3.0/Gtk.Label.justify.html

danirabbit (commented):
Hey thanks for taking this on! Just one suggestion

danirabbit:
Also OP just want to confirm that you did not use an LLM here in accordance with our contributor policy?  https://docs.elementary.io/contributor-guide/development/generative-ai-policy

lib/Widgets/HeaderLabel.vala
nityanand123gupta: Good call, updated to use justify instead — pushed

nityanand123gupta:
> Also OP just want to confirm that you did not use an LLM here in accordance with our contributor policy? https://docs.elementary.io/contributor-guide/development/generative-ai-policy

Yes, I confirm I did not use an LLM for this contribution

danirabbit (changes requested):
Looks like this needs a little more to it since it contains two separate labels we probably need to set xalign at the same time as justify in the labels.

Current result:

Image

nityanand123gupta:
Added `xalign` back alongside `justify` — bound both to each internal label now, since they control different things (xalign positions the block, justify aligns wrapped lines within it). Should fully center wrapped multi-line text now.

Also: the `build (stable)` / `build (unstable)` failures look unrelated to this PR — they're failing at the meson dependency-check stage (`gtk4 >=4.22` required, but those runners have `4.14.5` installed), before any code compiles. `build (development-target)` passed cleanly.

danirabbit (approved):
This works for me. Thanks for addressing it and being responsive to change requests!

nityanand123gupta:
Thanks for the review and guidance! Really appreciate it.

PR #1013, opened 2026-08-30
menubutton: arrow: Use open-menu-symbolic when no direction

### LINKED ISSUE #1012: Use open-menu-symbolic instead of pan-down-symbolic when Gtk.ArrowType.NONE is set
### Problem

Originally from https://github.com/elementary/fileroller/pull/76#issuecomment-5464955773

`pan-down-symbolic` is used for the menu icon if an app set `GTK_ARROW_NONE` (`Gtk.ArrowType.NONE` in Vala) to `Gtk.MenuButton.direction`:

https://github.com/elementary/granite/blob/952337ea74b925c06088e72d10fb11739fd22dbd/lib/Styles/Gtk/Arrow.scss#L2

---

An example in our ecosystem is File Roller:

https://gitlab.gnome.org/GNOME/file-roller/-/blob/44.7/src/fr-window.c?ref_type=tags#L4948

I don't think it's intuitive if the menu button uses `pan-down-symbolic`:

Image

### Proposal

Use `open-menu-symbolic` instead of `pan-down-symbolic` when `Gtk.ArrowType.NONE` is set.

Demonstration using File Roller:

Image

### Prior Art (Optional)

libadwaita uses `open-menu-symbolic`:

https://gitlab.gnome.org/GNOME/libadwaita/-/blob/42dc1102d07b75638abb283ac7ae7765baf23638/src/stylesheet/widgets/_buttons.scss#L454-456

ryonakano:
Fixes #1012

## Checklist
- [x] Confirmed the expected icon is used when building our File Roller with the following diff and changing GTK Theme to Granite using GTK Inspector  
  image
- [x] Confirmed other widgets like dropdowns is not affected to this change  
  image

<details>
<summary>A patch to build this branch inside the sanbox</summary>

```diff
diff --git a/org.gnome.FileRoller.json b/org.gnome.FileRoller.json
index c3ccc8e..1713862 100644
--- a/org.gnome.FileRoller.json
+++ b/org.gnome.FileRoller.json
@@ -99,6 +99,68 @@
                 }
             ]
         },
+        {
+            "name": "granite-7",
+            "buildsystem": "meson",
+            "config-opts": [
+                "-Ddemo=false"
+            ],
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://github.com/elementary/granite.git",
+                    "commit": "bc4455f175f5e56da9632f5139f6dfc57c02d03f",
+                    "x-checker-data": {
+                        "type": "git",
+                        "tag-pattern": "^([\\d.]+)$"
+                    }
+                }
+            ],
+            "modules": [
+                {
+                    "name": "sassc",
+                    "cleanup": [
+                        "*"
+                    ],
+                    "sources": [
+                        {
+                            "type": "git",
+                            "url": "https://github.com/sass/sassc.git",
+                            "tag": "3.6.2"
+                        },
+                        {
+                            "type": "script",
+                            "dest-filename": "autogen.sh",
+                            "commands": [
+                                "autoreconf -si"
+                            ]
+                        }
+                    ],
+                    "modules": [
+                        {
+                            "name": "libsass",
+                            "cleanup": [
+                                "*"
+                            ],
+                            "sources": [
+                                {
+                                    "type": "git",
+                                    "url": "https://github.com/sass/libsass.git",
+                                    "tag": "3.6.5"
+                                },
+                                {
+                                    "type": "script",
+                                    "dest-filename": "autogen.sh",
+                                    "commands": [
+                                        "autoreconf -si"
+                                    ]
+                                }
+                            ]
+                        }
+                    ]
+                }
+            ]
+        },
         {
             "name": "file-roller",
             "buildsystem": "meson",
```

</details>

### DIFF
--- a/lib/Styles/Gtk/Arrow.scss
+++ b/lib/Styles/Gtk/Arrow.scss
@@ -1,7 +1,5 @@
 arrow {
-    // Use non-arrow icon when no direction specified, assuming use case as a
-    // main menu button
-    -gtk-icon-source: -gtk-icontheme("open-menu-symbolic");
+    -gtk-icon-source: -gtk-icontheme("pan-down-symbolic");
     min-height: rem(16px);
     min-width: rem(16px);
 
--- a/lib/Styles/Gtk/Arrow.scss
+++ b/lib/Styles/Gtk/Arrow.scss
@@ -1,4 +1,6 @@
 arrow {
+    // Use non-arrow icon when no direction specified, assuming use case as a
+    // main menu button
     -gtk-icon-source: -gtk-icontheme("open-menu-symbolic");
     min-height: rem(16px);
     min-width: rem(16px);
--- a/lib/Styles/Gtk/Arrow.scss
+++ b/lib/Styles/Gtk/Arrow.scss
@@ -1,5 +1,5 @@
 arrow {
-    -gtk-icon-source: -gtk-icontheme("pan-down-symbolic");
+    -gtk-icon-source: -gtk-icontheme("open-menu-symbolic");
     min-height: rem(16px);
     min-width: rem(16px);
 
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ You'll need the following dependencies:
 * gettext
 * libgee-0.8-dev
 * libgirepository1.0-dev
-* libgtk-4-dev >= 4.12.0
+* libgtk-4-dev >= 4.22.0
 * sassc
 * valac
 
--- a/lib/Styles/Gtk/Button.scss
+++ b/lib/Styles/Gtk/Button.scss
@@ -51,3 +51,11 @@ button {
         border-spacing: $button-spacing / 2;
     }
 }
+
+menubutton {
+    arrow {
+        // Use non-arrow icon when no direction specified, assuming use case as a
+        // main menu button
+        -gtk-icon-source: -gtk-icontheme("open-menu-symbolic");
+    }
+}
--- a/lib/Styles/Gtk/Button.scss
+++ b/lib/Styles/Gtk/Button.scss
@@ -51,11 +51,3 @@ button {
         border-spacing: $button-spacing / 2;
     }
 }
-
-menubutton {
-    arrow {
-        // Use non-arrow icon when no direction specified, assuming use case as a
-        // main menu button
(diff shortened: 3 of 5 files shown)


lib/Styles/Gtk/Arrow.scss
```diff
@@ -1,5 +1,5 @@
 arrow {
-    -gtk-icon-source: -gtk-icontheme("pan-down-symbolic");
+    -gtk-icon-source: -gtk-icontheme("open-menu-symbolic");
```
danirabbit: Can you add a comment? I feel like this is not obvious why we're using something that isn't an arrow

ryonakano: Thanks for your comment. Does 245a593f62b4f4f5119fab415426527d971d09ad look fine for you?

danirabbit (approved):

danirabbit:
Thanks! We'll want to cherry pick this fix back to the granite-7 branch I imagine since main is granite 9?

ryonakano:
> Thanks! We'll want to cherry pick this fix back to the granite-7 branch I imagine since main is granite 9?

Am I understand correctly that what we should work on is the stylesheet repository not here because the new stylesheet has not been released yet?

danirabbit:
@ryonakano yes exactly. Sorry I misspoke about committing it to granite-7. It would need to be in Stylesheet

ryonakano:
Ah, I noticed we should limit this style to `menubutton` [as done in libadwaita](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/42dc1102d07b75638abb283ac7ae7765baf23638/src/stylesheet/widgets/_buttons.scss#L388), otherwise it will widely applied like this:

image

ryonakano:
@danirabbit I modified this fix so that it won't be affected to other places like dropdowns. Also backported it to stylesheet: https://github.com/elementary/stylesheet/pull/1354

danirabbit (changes requested):
Can you add a MenuButton file so we can stick to one class per file?

Sorry to keep asking for changes 🙇🏻‍♀️

ryonakano:
Should be good now :+1:

danirabbit (approved):

danirabbit:
Perfect thank you!

PR #1017, opened 2026-08-30
Settings: add new Portal appearance properties

danirabbit:
Created a delegate func here to DRY the DBus calls

### DIFF
--- a/lib/StyleManager.vala
+++ b/lib/StyleManager.vala
@@ -73,9 +73,6 @@ public class Granite.StyleManager : Object {
         update_color_scheme ();
         update_accent_color ();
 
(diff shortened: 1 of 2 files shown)


leolost2605 (changes requested):
Hmm reduced motion should work automatically since on wayland gtk checks the portal.
I just tried it and it does work for me on OS 9...

leolost2605:
I just checked gtk code and it should work with contrast too once when provide it in settings daemon

leolost2605:
https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/wayland/gdksettings-wayland.c#L269

I think it should even work for color scheme :shrug:

danirabbit:
Oh lmao yeah. Interesting!

PR #1018, opened 2026-08-31
Fix style class deprecations

danirabbit:
In preparation for #1010

doesn't touch classes that will be completely removed like settingssidebar

### DIFF
--- a/lib/Widgets/Dialog.vala
+++ b/lib/Widgets/Dialog.vala
@@ -74,7 +74,7 @@ public class Granite.Dialog : Gtk.Dialog {
 
         var content_area = get_content_area ();
         content_area.vexpand = true;
-        content_area.add_css_class (Granite.STYLE_CLASS_DIALOG_CONTENT_AREA);
+        content_area.add_css_class ("dialog-content-area");
     }
 
     public override void constructed () {
--- a/lib/Widgets/Dialog.vala
+++ b/lib/Widgets/Dialog.vala
@@ -43,7 +43,7 @@
 *   dialog.add_button ("Cancel", Gtk.ResponseType.CANCEL);
 *
 *   var suggested_button = dialog.add_button ("Suggested Action", Gtk.ResponseType.ACCEPT);
-*   suggested_button.add_css_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
+*   suggested_button.add_css_class (Gtk.CssClass.SUGGESTED);
 *
 *   dialog.show_all ();
 *   dialog.response.connect ((response_id) => {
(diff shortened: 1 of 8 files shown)


lib/Widgets/Dialog.vala
```diff
@@ -74,7 +74,7 @@ public class Granite.Dialog : Gtk.Dialog {
 
         var content_area = get_content_area ();
         content_area.vexpand = true;
-        content_area.add_css_class (Granite.STYLE_CLASS_DIALOG_CONTENT_AREA);
+        content_area.add_css_class ("dialog-content-area");
```
zeebok, line 77: According the deprecation note the replacement for this is "Granite.Dialog"

lib/Widgets/MessageDialog.vala
```diff
@@ -266,7 +266,7 @@ public class Granite.MessageDialog : Granite.Dialog {
             wrap = true,
             xalign = 0
         };
-        primary_label.add_css_class (Granite.STYLE_CLASS_TITLE_LABEL);
+        primary_label.add_css_class ("title");
```
zeebok, line 269: Note says the replacement is "Granite.MessageDialog"

lib/Widgets/MessageDialog.vala
```diff
@@ -290,7 +290,7 @@ public class Granite.MessageDialog : Granite.Dialog {
 
         get_content_area ().append (message_grid);
 
-        add_css_class (Granite.STYLE_CLASS_MESSAGE_DIALOG);
+        add_css_class ("message");
```
zeebok, line 293: This also noted to become "Granite.MessageDialog"

lib/Widgets/Placeholder.vala
```diff
@@ -144,9 +144,9 @@ public class Granite.Placeholder : Gtk.Widget {
         grid.attach (description_widget, 1, 1);
 
         var button = new Gtk.Button () {
-            child = grid
+            child = grid,
+            has_frame = false
         };
-        button.add_css_class (Granite.STYLE_CLASS_FLAT);
```
zeebok, line 149: This does not look be removed in the other PR. Is there a different reason to remove it here?

lenemter (approved):

zeebok (commented):
Only a few questions/notes since I am not fully up-to-date on where Granite 9 is

lib/Widgets/Dialog.vala
danirabbit: This is Granite Dialog haha. Basically this is an internal class name and doesn't have a public constant

lib/Widgets/Placeholder.vala
danirabbit: Using has_frame is just better than using a style class constant anyways so might as well clean it up

zeebok (approved):

PR #1022, opened 2026-09-01
Add notification enums and flags

danirabbit:
[Glib.Notification](https://valadoc.org/gio-2.0/GLib.Notification.html) is kind of outdated and doesn't support what's in the new portal spec like button purpose or sounds, but  [Xdp.Portal.add_notification](https://valadoc.org/libportal/Xdp.Portal.add_notification.html) is super open ended and not nice to use. I think we need to provide our own notification class, so I'm making enums and flags with functions that can be used with Xdg.Portal instead of just constants. This will be clunky to use right now but will be much easier to use with a future Granite.Notification

Unless otherwise specified, I'm going off the [XDG Portal spec](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Notification.html) here

Categories
---
There are two specs (wow suprise). One is from [XDG Portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Notification.html), but there's also an [Fd.o Spec](https://specifications.freedesktop.org/notification/latest/categories.html).

A lot of the ones mentioned in the portal are not app-facing but OS-facing so I'm inclined not to include them. We can touch the portal directly (or use GLib.Notification) in host applications to provider these notification categories if necessary.

The fd.o spec includes "presence" notifications which may have been a thing in the time of AIM, but seems really outdated. It also includes things like "network" which seems like the domain of the host.

Priority
---

The [fd.o spec](https://specifications.freedesktop.org/notification/latest/urgency-levels.html) includes 3 levels, whereas the portal and [GLib ](https://valadoc.org/gio-2.0/GLib.NotificationPriority.html) include 4 priority levels. I'm inclined to go with portal and GLib here

I'd like to name these in a way that can help give guidance for their usage. "High" vs "Urgent" is too difficult to make good decisions about.

I'd like to also codify some expectations about notification expiration here to help clean up old notifications etc.

Display Flags
---

This includes some ways to specific where your notification should show up. Some of the names they use I don't like for our purposes so trying to give those better names

Button Purpose
---

So we can do nice things like colored circular icon buttons for calls etc

The portal spec also includes:
* "im.reply-with-text" which we currently don't support and seems difficult to support, so I didn't include it for now
* "call.enable-speakerphone" and disable, which I guess we're not mobile yet, so  I didn't include it for now

### DIFF
--- a/lib/Notification.vala
+++ b/lib/Notification.vala
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2026 elementary, Inc. (https://elementary.io)
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ */
+
+public namespace NotificationCategory {
+    /**
+     * A received instant message notification
+     */
+     public const string IM_RECEIVED = "im.received";
+}
+
+public enum NotificationCategory {
+    /**
+     * A received instant message notification
+     */
+    IM_RECEIVED,
+
+    public string to_string () {
+        switch (this) {
+            case IM_RECEIVED:
+                return "im.received"
+        }
+    }
+}
+
+
+// // For alarm clock apps
+// ALARM_RINGING = "alarm.ringing";
+// // A audio or video call is incoming
+// CALL_INCOMING = "call.incoming";
+// // A audio or video call is ongoing.
+// CALL_ONGOING = "call.ongoing";
+// // An incoming audio or video call was not answered
+// CALL_UNANSWERED = "call.unanswered";
+// // Used to display an extreme weather warning.
+// WEATHER_WARNING_EXTREME = "weather.warning.extreme";
+// // A nationwide or presidential warnings broadcasted by the cell network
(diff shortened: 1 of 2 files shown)


iamghonim:
Names are clear and  NORMAL is better than DEFAULT to avoid confusion with system settings
Suggestion: rename TIME_SENSITIVE to  HIGH and EMERGENCY  CRITICAL for a clearer gradual scale (LOW - NORMAL - HIGH - CRITICAL)

danirabbit:
@iamghonim I'm pretty sure that I don't want to use "high" and "critical" and would rather use more semantic labels that make it clearer to developers what the appropriate level is for them to choose

tintou:
It is not clear to me what Glib.Notification does not support?

tintou:
Oh sorry, [Xdp.Portal.add_notification](https://valadoc.org/libportal/Xdp.Portal.add_notification.html) wasn't mentioning the additions of XDP Notification version 2

PR #1027, opened 2026-09-08
Styles/MenuButton: fix default icon, add an example menu

danirabbit:
Fixes making the default icon open-menu since in modern gtk it does get a direction style class.

While we're here, add a basic example menu that can build on in future PRs to expand PopoverMenu support


demo/Views/ControlsView.vala
```diff
@@ -23,6 +23,69 @@ public class ControlsView : DemoPage {
...
+        button_section.set_attribute_value ("display-hint", "circular-buttons");
+
+        var menuitem_section = new GLib.Menu ();
+        menuitem_section.append ("Move", null);
+        menuitem_section.append ("Delete", null);
+
+        var menu_model = new GLib.Menu ();
+        menu_model.append_item (button_section);
+        menu_model.append_section ("SwitchModelButton", switch_section);
+        menu_model.append_section ( null, menuitem_section);
```
leolost2605:
```suggestion
        menu_model.append_section (null, menuitem_section);
```

leolost2605 (approved):
One nit, apart from that LGTM

lib/Styles/Gtk/MenuButton.scss
```diff
@@ -1,7 +1,6 @@
 menubutton {
-    arrow {
-        // Use non-arrow icon when no direction specified, assuming use case as a
-        // main menu button
+    // Assuming use case as a main menu button
+    > button.image-button arrow.down {
         -gtk-icon-source: -gtk-icontheme("open-menu-symbolic");
     }
```
ryonakano, line 5: I think this change introduces https://github.com/elementary/granite/issues/1012 again, while I admit my first code applies to all arrow types. I'll make a followup PR to do the equivalent thing with [what libadwaita does](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/42dc1102d07b75638abb283ac7ae7765baf23638/src/stylesheet/widgets/_buttons.scss#L454-468).

ryonakano: Or simply make sure this is applied to `arrow.none` too.

ryonakano: Opened:

- https://github.com/elementary/granite/pull/1055
- https://github.com/elementary/stylesheet/pull/1356

PR #1028, opened 2026-09-08
Styles: make background levels into variables

danirabbit:
I don't see why this needs to be a function tbh. Lets us use media queries


danirabbit:
Getting some errors about "expected a valid color" but not sure why? It just points to Strawberry 200?

lib/Styles/_exported.scss
```diff
@@ -5,9 +5,6 @@
 :root {
     // Text, images, and other foreground elements
     --fg-color: var(--BLACK_500);
-    --bg-color: #{"" + bg-color(2)};
```
RMcNeely, line 8: These are in another `:root` from the earlier place where you defined background colors. Do they conflict or does one of them get compiled out?

RMcNeely (commented):
Im afk so I can't grep the places where the function was but i do love to see it. Even if I don't know GTK well 😅

lib/Styles/_exported.scss
danirabbit: I moved them from here to there so that it would be in one place instead of spread out since all variables are exported now

RMcNeely (approved):
I'm not gonna pretend to know how CSS works in GTK but from a CSS perspective this looks real nice 😎👍

PR #1030, opened 2026-09-08
Styles/Button: add suggested and destructive styles for image buttons

danirabbit:
And fix padding while we're here

Screenshot from 2026-09-08 16 23 58


lenemter (commented):
These buttons look identical with the Strawberry accent color:

Image

danirabbit:
Hm yeah that's no good. Will have to think about how to solve this

wpkelso:
@danirabbit Perhaps we could reuse the "white-on-red" motif from the text buttons?

danirabbit:
Screenshot from 2026-09-14 10 27 44

Fixed!

lenemter (approved):
I'm not a huge fan of the pressed animation, but than that it looks good

PR #1034, opened 2026-09-09
Styles/lighting: toplevel shadows

danirabbit:
First steps towards eventually removing the shadow function. Use class extends for toplevels


lib/Styles/Common/_lighting.scss
```diff
@@ -115,3 +105,29 @@
...
+    --shadow-spread: calc(var(--shadow-radius) * -0.5);
+    --shadow-color: color-mix(in srgb, black calc(var(--shadow-level) * 4% + 10% + var(--shadow-multiplier)), transparent);
+
+    --bloom-shadow-offset: calc(var(--shadow-offset) * 1.1);
+    --bloom-shadow-radius: calc(var(--shadow-radius) * 1.05 + 5px);
+    --bloom-shadow-spread: calc(var(--shadow-radius) * 0.6);
+    --bloom-shadow-color: color-mix(in srgb, var(--shadow-color) 50%, transparent);
+
+    box-shadow:
+        inset 0 1px 0 0 color-mix(in srgb, var(--highlight-color) 30%, transparent),
```
alice-mkh: Note: these will be under  the window content, not over. If you want it to be over the content too, it should be outline instead, or you need to have another element within the window

danirabbit: Oh you're so smart

danirabbit: Nvm seem like outline won't work here. Long term we probably want to provide a Granite.Window with an overlay

PR #1038, opened 2026-09-10
Styles: add basic spinbutton support

danirabbit:
The bare minimum:

Screenshot from 2026-09-09 19 28 44


wpkelso (approved):
we speen now

PR #1040, opened 2026-09-10
Styles: support levelbar

danirabbit:
Screenshot from 2026-09-10 11 27 18

My reasoning behind levelbar sizing is that currently we use it for things like disk fullness, battery charge, steps in a paged view etc.

For vertical discrete levelbars, the only use of this I've seen is for audio levels, so that's why I did this design

By default GTK assigns some of its own internal style classes, but I don't like having developers put an arbitrary string here, so I added support for named constants

I didn't include accent color here because your accent can be red or green or yellow

### DIFF
--- a/demo/Views/ControlsView.vala
+++ b/demo/Views/ControlsView.vala
@@ -169,10 +169,9 @@ public class ControlsView : DemoPage {
         var hdiscrete_levelbar = new Gtk.LevelBar.for_interval (0, 4) {
             mode = DISCRETE
         };
-        hdiscrete_levelbar.add_offset_value ("low", 1);
-        hdiscrete_levelbar.add_offset_value ("middle", 2);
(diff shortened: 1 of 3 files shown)


lenemter:
How to test this? With Granite stylesheet I get invisible levelbars:

image

And with system stylesheet I get old levelbars:

image

lenemter (commented):
It looks like Granite is forcing a minimum width/height here. Is this intentional?

Image


Also with some values, your accent color is visible, which, according to the description, shouldn't be

Image

danirabbit:
@lenemter color fixed!

Can you explain what you mean about the minimum width/height? We have to set some minimums or it would be 0 width/height, but otherwise what you see in the demo is filling the box size:

Screenshot from 2026-09-20 13 16 55

wpkelso (approved):

PR #1042, opened 2026-09-10
Install Granite icons

### LINKED ISSUE #1041: Install Granite icons to disk
### Problem

* Host icon theme icons override our resourced icons and
* symbolic icons are out of spec for an fd.o icon theme, but
* we still have apps that rely on them that can't be ported to Granite 9 soon, so
* we can't remove them from the host icon theme

### Proposal

Install Granite icons as a theme to the host and then we can make elementary icons inherit it

### Prior Art (Optional)

_No response_

danirabbit:
Fixes #1041

### DIFF
--- a/lib/Icons/index.theme
+++ b/lib/Icons/index.theme
@@ -1,0 +1,13 @@
+[Icon Theme]
+Name=Granite
+Hidden=true
+
+Example=folder
+Directories=scalable
+
+[scalable]
+Context=UI
+Size=16
+MinSize=8
+MaxSize=512
+Type=Scalable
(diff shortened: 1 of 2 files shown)


leolost2605 (approved):
Just to be sure: the long term goal here is to move all non spec icons to granite so that apps only have to depend on granite and be sure they will have all icons.
But because that would require them to have granite 9 in order to be sure that the new icons will be added to their search path we for now install them **also** as an icon theme that old apps use via the elementary icon theme so that we can already remove them from the elementary icon theme repo?

If so LGTM 😄

danirabbit:
Exactly!

PR #1044, opened 2026-09-10
Icons: sort into folders

danirabbit:
Prepares us to absorb the entire symbolic icon theme


lib/Icons/icons.gresource.xml
```diff
@@ -1,95 +1,107 @@
...
-    <file alias="16x16/status/emblem-error.svg" compressed="true" preprocess="xml-stripblanks">16/emblem-error.svg</file>
-    <file alias="16x16/status/emblem-mixed.svg" compressed="true" preprocess="xml-stripblanks">16/emblem-mixed.svg</file>
-    <file alias="16x16/status/emblem-warning.svg" compressed="true" preprocess="xml-stripblanks">16/emblem-warning.svg</file>
-
-    <file alias="16x16@2/status/emblem-disabled.svg" compressed="true" preprocess="xml-stripblanks">16/emblem-disabled.svg</file>
-    <file alias="16x16@2/status/emblem-enabled.svg" compressed="true" preprocess="xml-stripblanks">16/emblem-enabled.svg</file>
-    <file alias="16x16@2/status/emblem-error.svg" compressed="true" preprocess="xml-stripblanks">16/emblem-error.svg</file>
-    <file alias="16x16@2/status/emblem-mixed.svg" compressed="true" preprocess="xml-stripblanks">16/emblem-mixed.svg</file>
-    <file alias="16x16@2/status/emblem-warning.svg" compressed="true" preprocess="xml-stripblanks">16/emblem-warning.svg</file>
+  <gresource prefix="/io/elementary/granite16x16/status">
```
leolost2605:
```suggestion
  <gresource prefix="/io/elementary/granite/16x16/status">
```

Shouldn't there be a `/`?

leolost2605 (changes requested):
One question apart from that LGTM

lib/Icons/icons.gresource.xml
danirabbit: Oops thanks for catching that

lib/Icons/icons.gresource.xml
```diff
@@ -1,95 +1,107 @@
...
+
+    <file compressed="true" preprocess="xml-stripblanks">eye-not-looking-symbolic.svg</file>
+    <file compressed="true" preprocess="xml-stripblanks">eye-open-negative-filled-symbolic.svg</file>
+
+    <file compressed="true" preprocess="xml-stripblanks">list-drag-handle-symbolic.svg</file>
+
+    <file compressed="true" preprocess="xml-stripblanks">pan-down-symbolic.svg</file>
+    <file compressed="true" preprocess="xml-stripblanks">pan-end-symbolic.svg</file>
+    <file compressed="true" preprocess="xml-stripblanks">pan-start-symbolic.svg</file>
+    <file compressed="true" preprocess="xml-stripblanks">pan-up-symbolic.svg</file>
```
wpkelso, line 99: `pan-*` implies an action more than a UI element to me. Should these be under `actions/` instead?

wpkelso (commented):
I think it would be good to namespace icons that don't have names coming from the fdo spec under `granite-`. This way it's straightforward to keep track of what is fdo and what is an extension.

danirabbit:
@wpkelso nothing here is fd.o. -symbolic is all outside of the spec. But also we can't do that for a lot of the names, for example the `poi` names are all coming from a library

lib/Icons/icons.gresource.xml
danirabbit: I don't think so. These are all used in dropdowns, submenus, and expanders etc

danirabbit:
@wpkelso I also think it's outside the scope of this branch to do any renames. I've just moved existing icons into folders

lib/Icons/icons.gresource.xml
wpkelso: Hmm, ok. That works then

wpkelso (approved):
No further comments or questions from me then; lgtm

PR #1045, opened 2026-09-11
Demo: move datetime to forms view

danirabbit:
Screenshot from 2026-09-11 10 49 08

* Fix deprecation warning with `calendar.year`/`month`/`day`
* DatePicker: Set date after we connect signals so it isn't blank when opening
* Add menu style class to popovers for padding


leolost2605 (approved):
Makes sense

PR #1049, opened 2026-09-17
meson: Always assume style sheet targets are stale

leolost2605:
Add `build_always_stale` to the style sheet custom targets. This is because we only pass the top level file to the custom target and sassc imports the rest. However because of that meson doesn't know about the rest and doesn't rebuild the targets if the top level hasn't changed. Therefore tell meson the targets are always stale as compiling them doesn't take that long.


leolost2605:
With this no `ninja clean` is needed when a sublevel .scss file was changed. Currently `ninja clean` is needed right or am I missing something?

stsdc:
I have the same issue with Monitor. Always have to use `ninja clean` if I'm dealing with a style changes. This change worth to apply in other repos.

stsdc (approved):

PR #1053, opened 2026-09-19
Styles: make window.devel a purpleprint

danirabbit:
Closes #997

There were too many changes I wanted, so I just opened a PR:

* Move to Window.scss. I don't love this `classes` file tbh, but especially when we're targeting a specific widget class I think it's better to use that class file
* Use standard CSS variables and functions
* Don't scale the grid line width, only the spacing
* DRY sidebar styles by using a variable for bg color
* Support Adw.ToolbarView while we're here, since a lot of our apps use that still
* Fix conflicts, some trailing spaces, some broken indentation, etc


wpkelso (approved):
I'm happy with this 👍

PR #1054, opened 2026-09-19
Bin/Box: add has_frame property

### LINKED ISSUE #1033: Add has_frame properties to Box and Bin
### Problem

Currently we have a Css class for Cards, but we've been moving away from manually adding style classes towards widgets having properties

### Proposal

Maybe it would be nicer if we just added `has_frame` properties to box and bin?

### Prior Art (Optional)

_No response_

danirabbit:
Fixes #1033


leolost2605 (commented):
What about just having a granite.card widget? Would scale a lot better

danirabbit:
@leolost2605 im not sure because it's really just a visual effect and other widgets have a frame property so it seems in keeping with what is expected of GTK widgets?

leolost2605 (approved):
Hmm but GTK uses it for kind of standalone widgets where it always *could* make sense to have a frame (button, entry, etc.). They don't have it on every container widget, that's what Gtk.Frame is for. 
Idk I'm not super opposed to this it just feels kinda weird. 
Code LGTM

PR #1055, opened 2026-09-20
MenuButton: Make sure open-menu icon used for Gtk.ArrowType.NONE

ryonakano:
## Before
File Roller built with fa3de95c966df62f02d1455f7b9d9bd867ab6d5c (note that https://github.com/elementary/granite/issues/1012 is introduced again):

Screenshot From 2026-09-20 16-25-15

### After
File Roller built with this branch (0ba2cc734752211076175d0cf99371e8b2a3370d):

Screenshot From 2026-09-20 16-27-00

It still shows the arrow icons if you specify `Gtk.ArrowType.UP` for example:

image

<details>
<summary>A patch to build this branch inside the sanbox</summary>

```diff
diff --git a/org.gnome.FileRoller.json b/org.gnome.FileRoller.json
index 8450685..837c924 100644
--- a/org.gnome.FileRoller.json
+++ b/org.gnome.FileRoller.json
@@ -99,6 +99,68 @@
                 }
             ]
         },
+        {
+            "name": "granite-7",
+            "buildsystem": "meson",
+            "config-opts": [
+                "-Ddemo=false"
+            ],
+            "sources": [
+                {
+                    "type": "git",
+                    "url": "https://github.com/elementary/granite.git",
+                    "commit": "0ba2cc734752211076175d0cf99371e8b2a3370d",
+                    "x-checker-data": {
+                        "type": "git",
+                        "tag-pattern": "^([\\d.]+)$"
+                    }
+                }
+            ],
+            "modules": [
+                {
+                    "name": "sassc",
+                    "cleanup": [
+                        "*"
+                    ],
+                    "sources": [
+                        {
+                            "type": "git",
+                            "url": "https://github.com/sass/sassc.git",
+                            "tag": "3.6.2"
+                        },
+                        {
+                            "type": "script",
+                            "dest-filename": "autogen.sh",
+                            "commands": [
+                                "autoreconf -si"
+                            ]
+                        }
+                    ],
+                    "modules": [
+                        {
+                            "name": "libsass",
+                            "cleanup": [
+                                "*"
+                            ],
+                            "sources": [
+                                {
+                                    "type": "git",
+                                    "url": "https://github.com/sass/libsass.git",
+                                    "tag": "3.6.5"
+                                },
+                                {
+                                    "type": "script",
+                                    "dest-filename": "autogen.sh",
+                                    "commands": [
+                                        "autoreconf -si"
+                                    ]
+                                }
+                            ]
+                        }
+                    ]
+                }
+            ]
+        },
         {
             "name": "file-roller",
             "buildsystem": "meson",
```

</details>


danirabbit:
Unless FileRoller is building with Granite 9, the stylesheet here shouldn't have an effect

danirabbit (approved):

danirabbit:
@ryonakano is this perhaps because FileRoller here is building with an older GTK4?

This works as intended in the Demo here with GTK 4.22.4:

Screenshot from 2026-09-20 12 40 42

It also works fine with your branch so I'm not opposed to merging, but it would be nice to know how to reproduce this

ryonakano:
> @ryonakano is this perhaps because FileRoller here is building with an older GTK4?

@danirabbit No, this is because File Roller sets `direction` to `GTK_ARROW_NONE` (`Gtk.ArrowType.NONE` in Vala), whereas granite-9-demo sets no `direction`:

https://gitlab.gnome.org/GNOME/file-roller/-/blob/44.7/src/fr-window.c?ref_type=tags#L4948

Probably you should be able to reproduce the issue in granite-9-demo if you add `direction = NONE`.

In #1027 we lost override against `arrow.none` (which has been too widely applied against whole `arrow` before that fix), resulting now `arrow.none` falls back to `pan-down-symbolic`. This PR makes sure we override `arrow.none`.

danirabbit:
Oh that's interesting they're explicitly setting it. Thanks for looking into that! I think this is good to merge

PR #1057, opened 2026-09-20
Add AccentColor enum and style radios

danirabbit:
Screenshot from 2026-09-20 16 03 54

We use AccentColor:
* In System Settings and Onboarding for selecting system accent color.
* In Files, Tasks and Calendar for setting colors for files and folders, lists ,and calendars respectively
* In a bunch of different places to style icons in lists, labels, etc

I've added a couple of functions here:
* `to_name` since we use these in tooltips for accent color selectors
* `to_css_class` since we use that to style radios, strings, labels, etc
* `to_hex` since in the case of Evolution and Files we need to store the actual hex representation.


lib/Constants.vala
```diff
@@ -160,6 +161,146 @@ namespace Granite {
...
+                case ORANGE: return "#f37329";
+                case YELLOW: return "#e6a92a";
+                case GREEN: return "#68b723";
+                case TEAL: return "#0e9a83";
+                case BLUE: return "#3689e6";
+                case PURPLE: return "#a56de2";
+                case PINK: return "#de3e80";
+                case BROWN: return "#8a715e";
+                case GRAY: return "#667885";
+                case LATTE: return "#";
```
danirabbit, line 298: Do we have a hex color for latte somewhere?

danirabbit: @micahilbery Should this just be LATTE 500?

zeebok (approved):
Besides the latte hex, the code looks good

lib/Constants.vala
wpkelso: I personally would vote for 300 or 400; they feel the most "latte" to me.

